PR #13, opened 2016-03-02
Importing PaymentRequest specifications from the WICG

adrianba:
This was [discussed at the F2F](https://www.w3.org/2016/02/24-wpwg-minutes#item12) and the proposal was made to move documents from https://github.com/WICG/paymentrequest/tree/gh-pages/specs to the WG repo.


adrianba:
Merged as 12cda4283ccf66380c814b6d17a4f664f5c098d0

Issue #9, opened 2016-03-02
Different card schemes have different mandatory field requirements


adrianba:
This issue comes from WICG/paymentrequest#20 and is now called out in the `basic-card-payment` document.

mattsaxon:
@adrianba 
I've submitted a pull request with a proposed update to address this

mattsaxon:
The PR has a flat structure for the data with some optional fields.

I've done some further research and it appears that the only card types that use this data structure today are Meastro cards. There were other brands before such as Switch and Solo, but these are no longer issued.

We should discuss if Maestro should be a specific type (e.g. sub-class of BasicCardResponse) of PaymentResponse object.

What are your views?

adrianba:
> I've done some further research and it appears that the only card types that use this data structure today are Meastro cards. There were other brands before such as Switch and Solo, but these are no longer issued.

This is what I found too but it wasn't clear to me from a quick look where Maestro still had this requirement. For this reason, I didn't include the fields in the initial draft. 

> We should discuss if Maestro should be a specific type (e.g. sub-class of BasicCardResponse) of PaymentResponse object.

I think this is a good question. First I wonder whether there are other fields used in other geos that would also need to extend the response. If we end up with a myriad of options fields that are rarely used then that might not make the spec very easy to consume. On the other hand, if most implementations will need to consider start date or issue number then including it in the basic data structure might be considered beneficial.

It would be good to get more data before deciding but equally we should pick something soon and change it if new information arises later. We need to address this to determine what to do with #25.

mattsaxon:
Adrian, I've dug deeper and found the following: Whilst Maestro is still a brand, Maestro cards that need issue number or start date to be submitted were actually using the Switch network scheme (even though the brand was Maestro), this network was closed in 2011, this effectively means that these card went out of circulation sometime in 2014.

I think this closes the issue and we can stick to the simpler format. I'll edit my PR to remove these, but will leave the revisions to the field names.

adrianba:
> Adrian, I've dug deeper and found the following: Whilst Maestro is still a brand, Maestro cards that need issue number or start date to be submitted were actually using the Switch network scheme (even though the brand was Maestro), this network was closed in 2011, this effectively means that these card went out of circulation sometime in 2014.

That matches the results of the research we've been doing too.

nickjshearer:
[...]

PR #25, opened 2016-03-04
Redefinition of BasicCardResponse to address Issue 9

mattsaxon:
Also revised property names for nameOnCard, pan and securityCode to be
more descriptive and consistent.

### DIFF
--- a/specs/basic-card-payment.html
+++ b/specs/basic-card-payment.html
@@ -169,7 +169,7 @@
       <h2>BasicCardResponse</h2>
       <pre class="idl">
         dictionary BasicCardResponse {
-          required DOMString cardHolderName;
+          required DOMString cardholderName;
           required DOMString cardNumber;
           required DOMString expiryMonth;
           required DOMString expiryYear;
--- a/specs/basic-card-payment.html
+++ b/specs/basic-card-payment.html
@@ -142,9 +142,14 @@
 
       <table>
         <tr><th>Identifier String</th><th>Description</th></tr>
-        <tr><td>visa</td><td>VISA cards</td></tr>
-        <tr><td>...</td><td>...</td></tr>
-      </table>
+        <tr><td>visa</td><td>Visa (Credit, Debit, Delta or Electron)</td></tr>
+        <tr><td>mastercard</td><td>MasterCard and EuroCard</td></tr>
+        <tr><td>amex</td><td>American Express</td></tr>
+        <tr><td>discover</td><td>Discover</td></tr>
+        <tr><td>maestro</td><td>Maestro</td></tr>
+	    <tr><td>diner</td><td>Diners Club</td></tr>
+        <tr><td>jcb</td><td>JCB</td></tr>
+	  </table>
     </section>
 
     <section id="init-data">
@@ -182,8 +187,8 @@
       </p>
 
       <dl>
-        <dt><dfn><code>cardHolderName</code></dfn></dt>
-        <dd>The <code>cardHolderName</code> field contains the card holder's name as it appears on the card.</dd>
+        <dt><dfn><code>cardholderName</code></dfn></dt>
+        <dd>The <code>cardholderName</code> field contains the cardholder's name as it appears on the card.</dd>
         <dt><dfn><code>cardNumber</code></dfn></dt>
         <dd>The <code>cardNumber</code> field contains the primary account number (PAN) for the payment card.</dd>
         <dt><dfn><code>expiryMonth</code></dfn></dt>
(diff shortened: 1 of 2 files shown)


zkoch:
Thanks, Matt. I just noticed that in the response we still don't have anything that indicates the scheme. Do we want to have this returned explicitly? Or should we rely on merchants using the BIN to figure it out?

mattsaxon:
@zkoch isn't the scheme represented as the methodName is the PaymentResponse object? I presume this would come from the Payment Method Identifer and I've added to the list of these in the basic card payment spec in this pull request

mattsaxon:
@zach, though we do need to discuss what happens for extensions to basic card details, e.g. tokenisation. I need to review Adrian's proposals and will get back to you post reading those

specs/basic-card-payment.html
```diff
@@ -142,9 +142,14 @@
...
-      </table>
+        <tr><td>visa</td><td>Visa (Credit, Debit, Delta or Electron)</td></tr>
+        <tr><td>mastercard</td><td>MasterCard and EuroCard</td></tr>
+        <tr><td>amex</td><td>American Express</td></tr>
+        <tr><td>discover</td><td>Discover</td></tr>
+        <tr><td>maestro</td><td>Maestro</td></tr>
+	    <tr><td>diner</td><td>Diners Club</td></tr>
+        <tr><td>jcb</td><td>JCB</td></tr>
+	  </table>
     </section>
```
adrianhopebailie, line 153: I suggest we namespace these for this payment method because it's quite possible (likely) that we'll have a "tokenized card payments" or "encrypted card payments" spec pretty soon.

``` html
<tr><td>basic-card-visa</td><td>Visa (Credit, Debit, Delta or Electron)</td></tr>
<tr><td>basic-card-mastercard</td><td>MasterCard and EuroCard</td></tr>
<tr><td>basic-card-amex</td><td>American Express</td></tr>
<tr><td>basic-card-discover</td><td>Discover</td></tr>
<tr><td>basic-card-maestro</td><td>Maestro</td></tr>
<tr><td>basic-card-diners</td><td>Diners Club</td></tr>
<tr><td>basic-card-jcb</td><td>JCB</td></tr>
```

mattsaxon:
Let me have a think on that, there was talk of a hierarchical naming too rather than the flat structure you're suggesting, perhaps "card.plain.visa", "card.encrypted.visa", "card.token.visa". I'll read Adrian B's  proposals and comment further

adrianhopebailie:
That could work but I think we can't enforce something like that I don't think. What would Bitcoin be for example? "cryptocurrency.bitcoin"?

mattsaxon:
I read Adrian B's proposal now, so appreciate we need the domain name in there to make the minting feasible outside of w3c, my examples were effectively shortnames, but I appreciate my use of the dotted naming made it unclear.

Using Adrian's Option 1 as a template, here are some examples

visa.com/card/plain
visa.com/card/tokenised
mastercard.com/card/plain
worldpay.com/tokenised
w3.org/card/visa/plain
bitcoin.org/V1

visa/plain would be the shortname for http://w3.org/card/visa/plain

adrianhopebailie:
I'm less concerned about the format and more that we don't spread the idea that "Visa" is a good example of payment method identifier. That's an important point to come to consensus on (we started drilling into it late in the afternoon at the F2F).

Payment Method Identifiers (PMI) serve 2 purposes so their design must consider both:
1. To indicate request and response message format for the interactions between the website, browser and payment app. A single specification of these message formats could apply to a set of many payment method identifiers (as is evidenced in this spec).
2. To indicate specifically which underlying payment instruments (brands) are supported by the website. i.e. A merchant can process card payments but might not accept all card brands.

From an architectural perspective one could think of PMIs as foreign keys that provide a many-to-many link between payment instruments and the messaging protocols that can be used to pass payment data back and forth for those instruments.

I can imagine people minting new PMIs for two reasons:
1. They have developed a new messaging protocol and have documented it in a specification like this one and need to provide PMIs that link payment instruments that can be used with this protocol to the protocol.
2. They have a payment instrument that can be used with an existing messaging protocol and so they mint a PMI that they get added to the list of supported PMIs in the existing protocol's spec.

burdges:
I suppose "Verified By Visa" is an actual PMI though, no?

adrianhopebailie:
> I suppose "Verified By Visa" is an actual PMI though, no?

Could be, if there is a message protocol spec for how it would work using the browser API

mattsaxon:
@adrianhopebailie, I agree that visa is not a good PMI without some namespacing, it is a good example of a "brand" though and the naming does to some extent have the brand as a separately identifiable part of the PMI.

Are you not happy that my examples are reasonable PMIs?

mattsaxon:
Verified By Visa is the 3DS implementation of Visa's, depending on the implementation, it could be done in various ways, but if we want the payment application to support it, we would need either new PMI defined, or an extension to the BasicCardResponse that allowed for additional authentication data to be added.

adrianba:
Merged as 0469d928a80bc28f7a9e9f295add65b866c1fa97 for further discussion in #9.

mattsaxon:
Added UnionPay as PR 90

PR #36, opened 2016-03-10
Gh pages

ianbjacobs:
- removed empty registry section
- swapped two sections since one is essentially empty and it should go
  last.


adrianba:
Merged as 7e3ad021ca0084bfc04e9a06ffed8bbbed3edd06

PR #37, opened 2016-03-11
Editorial and some clarifications

adrianhopebailie:
VISA, Mastercard etc are not good examples of payment method
identifiers as there will likely be many payment methods that support
these payment instruments


adrianba:
Included by @ianbjacobs in #36 and merged as 7e3ad021ca0084bfc04e9a06ffed8bbbed3edd06.

PR #64, opened 2016-03-15
Some friendly editorial changes

halindrome:
A couple of typos and suggested wording changes.  Use at your discretion.

### DIFF
--- a/specs/architecture.html
+++ b/specs/architecture.html
@@ -348,7 +348,7 @@
       	  <tr>
       	    <td>Payment App</td>
       	    <td>Digital Wallet, Payment Instrument </td>
-      	    <td>A piece of software to manage payment methods</td>
+      	    <td>A piece of software to manage payment methods and coordinate the issuance of payments.</td>
       	  </tr>
       	  <tr>
       	    <td>Payment Method</td>
--- a/specs/architecture.html
+++ b/specs/architecture.html
@@ -348,7 +348,8 @@
       	  <tr>
       	    <td>Payment App</td>
       	    <td>Digital Wallet, Payment Instrument </td>
-      	    <td>A piece of software to manage payment methods and coordinate the issuance of payments.</td>
+      	    <td>An application, registered by the payer, that supports processing payment 
+      	    requests and returning payment responses for one or more payment methods.</td>
       	  </tr>
       	  <tr>
       	    <td>Payment Method</td>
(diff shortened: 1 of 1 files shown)


specs/architecture.html
```diff
@@ -348,7 +348,7 @@
       	  <tr>
       	    <td>Payment App</td>
       	    <td>Digital Wallet, Payment Instrument </td>
-      	    <td>A piece of software to pay</td>
+      	    <td>A piece of software to manage payment methods</td>
```
zkoch: It's more than that really, no? It's either: a piece of software that does pay or returns back a payment credential to enable payment.

zkoch:
This mostly LGTM. I'm not sure about the change to the payment app definition (though in general, I agree it should probably be modified).

specs/architecture.html
```diff
@@ -358,7 +358,7 @@
       	  <tr>
       	    <td>Payment Mediator</td>
       	    <td>Digital Wallet, Payment Instrument</td>
-      	    <td>A piece of software that coordinates messages between Web apps and Payment apps</td>
+      	    <td>A piece of software that coordinates messages between sellers and Payment apps</td>
       	  </tr>
```
adrianhopebailie, line 363: Don't agree with this change. The messages are exchanged between the website and the payment app via the mediator.

specs/architecture.html
```diff
@@ -348,7 +348,7 @@
       	  <tr>
       	    <td>Payment App</td>
       	    <td>Digital Wallet, Payment Instrument </td>
-      	    <td>A piece of software to pay</td>
+      	    <td>A piece of software to manage payment methods and coordinate the issuance of payments.</td>
       	  </tr>
```
adrianhopebailie, line 353: A payment app does not manage payment methods. It supports a specific set of payment methods of which only a subset may be enabled because the user may have only configured the app with the necessary information (like payment instrument credentials) for certain methods.

I also don't like the wording "issuance of payments".

Proposed change:

```
<td>An application, registered by the payer, that supports 
processing payment requests and returning payment responses 
for one or more payment methods.</td>
```

halindrome:
I don't hate that.  It is closer to what I think should happen, and is loose enough that many different implementation strategies could be used.

msporny:
+1 to @adrianhopebailie's suggested change above.

adrianhopebailie:
@halindrome Can you update your PR to reflect the discussion here?

halindrome:
I have updated @adrianhopebailie - however, GitHub indicates that this is an outdated diff, so perhaps it was overcome by events anyway.

adrianhopebailie:
@halindrome LGTM - looks like it will merge fine unless I am missing something

adrianba:
Merged as e72f051cf9c7460b115ee72f97d90a6747632cd4

PR #67, opened 2016-03-16
Remove duplicate conformance language

halindrome:
This text is provided automatically, so is duplicated in the generated output.


adrianba:
Merged as eb7f42ac7204c8032f759842bff4a481f6919ebb

PR #70, opened 2016-03-16
Make activation language more permissive

halindrome:
I know this is just an example anyway, but I would prefer that we not even have the perception that this is limited to activating a button.  It could, for example, be a voice command in an interactive game ("Level Up!") that starts the process.


adrianhopebailie:
LGTM

adrianba:
Merged as 8d813f4cda5016900ef0d1bb16cb3c4af781485c

Issue #40, opened 2016-03-14
How do organizations layer additional information in the core payment messages?


msporny:
Migrated from https://github.com/w3c/webpayments/issues/18:

Are the core set of payment messages extensible?

If they are extensible, what is the extensibility mechanism?

Is the browser required to understand this extensibility mechanism?

msporny:
This was partially resolved (via WG resolution) in the Web Payments CG spec via this language:

http://wicg.github.io/web-payments-browser-api/#extensibility

ianbjacobs:
Also, the API allows for payment-method specific data.
https://w3c.github.io/browser-payment-api/specs/paymentrequest.html#paymentrequest-constructor

ianbjacobs:
See also this issue about feature detection:
 https://github.com/w3c/browser-payment-api/issues/33

I suggest we close this issue.

Ian

msporny:
-1 to closing the issue because of the following reasons:
1. The spec doesn't have a section that talks about extensibility of messages.
2. There is no way to get to the 'data' object that is passed into the PaymentRequest constructor.

It's currently left unspecified how any of this payment-method specific data finds its way to a payment app.

ianbjacobs:
I believe point one is covered by 
https://github.com/w3c/browser-payment-api/issues/44

msporny:
@ianbjacobs agreed that point 1 is covered by #44. I was responding to the "close this issue", which we shouldn't do until there is a PR that's been accepted that covers 1 and 2 above.

adrianba:
Extensibility is covered by #44. The developer accesses `data` using their own code.

```
var data = { ... };
var request = new PaymentRequest(a,b,c,data);
//TODO: use data here
```

msporny:
> developer accesses data using their own code.

Where is it specified how the payment app gets access to that information? A native payment app? What about a web-based payment app?

msporny:
> Extensibility is covered by #44.

For some very limited definition of extensibility. I'm asserting that not everything about extensibility is covered by #44 and there is nuance here that we need to discuss as a group.

We still haven't had the discussion around extensibility that the Web Payment CG specs raised via the Messaging spec (see PaymentRequest) here:

http://web-payments.github.io/web-payments-messaging/#payment-request

ianbjacobs:
@msporny wrote:

"Where is it specified how the payment app gets access to that information? A native payment app? What about a web-based payment app?"

Is that question registered as issue 50?
[...]

PR #72, opened 2016-03-16
Add a reference to issue #40.

msporny:
Add a reference to issue #40.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -89,7 +89,7 @@
         <a href="https://www.w3.org/community/wicg/">Web Platform Incubator Community Group</a>.
       </p>
     </section>
-    
+
     <section class='informative'>
       <h2>Introduction</h2>
         <p>Buying things on the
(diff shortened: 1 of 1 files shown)


specs/paymentrequest.html
```diff
@@ -275,6 +275,15 @@
...
         </p>
+
+        <p class="issue" data-number="40" title="How do developers layer additional information in the core payment messages?">
+          How does a developer access the <code>data</code> object?
+          How is <code>data</code> sent to the payment app?
+          Are the core set of payment messages (request/response) extensible?
+          If they are extensible, what is the extensibility mechanism?
+          Is the browser required to understand this extensibility mechanism?
+        </p>
+
```
adrianhopebailie, line 286: I think this wording is ambiguous and many of these questions are red herrings.

Proposed issue marker text:

```
<p class="issue" data-number="40" title="How does a website pass additional 
(not payment method specific) data to the payment app?">
It is proposed that a conformance criteria for implementations of this API be 
that any data passed into the request is passed on to the payment app unaltered. 
This would allow extensions of the data schema such as the addition of 
properties that are not documented in this specification or known to implementors 
such as JSON-LD @context or similar to be passed between the website and 
payment app.
</p>
```

adrianhopebailie, line 286: Also note the resolution of the group which included wording for this issue marker:
https://github.com/w3c/webpayments/issues/83

This is not appropriate wording while we have no extensibility guidance document but should be included as soon as we do.

msporny, line 286: Works for me, happy to go w/ the issue text the group agreed to. I trust that the editors will use the issue text you pointed to.

mountainhippo, line 286: +1 for me. This feels like an important PR for the FPWD.

adrianba:
Merged as be49a84a51f1ec8b2e15d2577de2ef66d191f4b2 with text from @adrianhopebailie.

Issue #38, opened 2016-03-14
How are payment apps shared between different browser brands?


msporny:
Migrating issue from https://github.com/w3c/webpayments/issues/15:

When a user registers a new payment instrument, how is that payment instrument shared between different browser brands? For example, if I register a Visa debit card issued by my bank in Google Chrome on my laptop, when I go to purchase something using my mobile phone, is that same debit card available to me via my Firefox web browser (assuming I've authenticated in some way with both browsers)?

There are a variety of choices here, in order of easiest to implement to hardest to implement:
1. There is no sharing mechanism
2. There is a unified payment instrument export mechanism
3. There is a storage and discovery mechanism that all browsers use to import data from other browsers

I assert that the last bullet point is what we should strive for, but we may need to take baby steps to get there. In any case, we should implement the API as if the third item existed and figure out a way to polyfill our way there.

ianbjacobs:
Centralized registration data is convenient but IMO not mandatory for v1. 

Could be done with a centralized payment app manager that invokes the API? The user tells the payment app manager: "Please keep the following browsers I have installed aware of my payment apps" and the manager invokes the API for each browser. 

Ian

mattsaxon:
@adrianhopebailie, re: comment on PR #73  I think this depends on the view of the WG with regard to the following statement in the Architecture document

> We expect that user agents will primarily act as payment mediators in this architecture.

This really is saying that browser will act as payment mediators, but I believe we should be allowing ,in the architecture, for a way for a user-agent to delegate to a 3rd party mediator. Furthermore the allowance for a user to choose to register a 3rd party mediator could be part of the conformance to the specification. This would create a more open ecosystem.

If we interpret the statement literally, whilst browser may be the "primary mediator", we must, to fulfil the architecture allow for an extensibility mechanism as I describe above, i.e. to enable non-primary mediators.

ianbjacobs:
It is already the case that any piece of software could implement the API. You could be in the middle of your favorite photo editing tool and pay for some filters using the API if the photo editing tool supports the API. 

Delegation of the mediator role does not seem to me to be a v1 feature in any case.

Ian

mattsaxon:
I'm suggesting that user-agent and mediator in the architecture are different actors and we should define an interface between them.

mattsaxon:
I'd be happy with this not being V1 though, but we really do need to update our (the IG) use cases documents to track these V2+ items. Or create an alternative approach to tracking requirements.

[...]

PR #73, opened 2016-03-16
Add a reference to issue #38

msporny:
Add a reference to issue #38.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -459,6 +459,14 @@
             later be resolved by the <a>user accepts the payment request algorithm</a> through interaction with the user interface.
           </li>
         </ol>
+        <p class="issue" data-number="38" title="How are payment apps shared between different browser brands?">
+          When a user registers a new payment instrument, how is that payment
+          instrument shared between different browser brands? For example, if I
+          register a Visa debit card issued by my bank in Google Chrome on my
+          laptop, when I go to purchase something using my mobile phone, is
+          that same debit card available to me via my Firefox web browser
+          (assuming I've authenticated in some way with both browsers)?
+        </p>
       </section>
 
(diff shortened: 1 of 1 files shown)


zkoch:
I think this is an issue for the registration spec when we have one. I don't think it belongs in PaymentRequest.

msporny:
Agreed that it eventually belongs in a registration spec, but we won't have one for FPWD, so we do want to raise awareness that we're thinking about this issue. We could add to the spec issue text that we'll eventually have a registration spec (or integrate registration into this spec).

adrianhopebailie:
:+1: for adding a note on this here until we have a registration spec to put it in

mountainhippo:
@zkoch if you're unhappy with this being in the API spec, we could instead include a note or issue marker in the architecture spec in the payment app section (section 2.2). But I agree with @msporny and @adrianhopebailie that we should flag this issue in the FPWD.

adrianhopebailie:
@mountainhippo and @msporny - having re-read the issue I think we should leave it out of the FPWD.

I don't think we should place any mechanism for browsers to synchronise payment apps or payment instruments between installations as explicitly in scope.

This seems like a feature browsers may offer to assist their users but I'm not sure if we can mandate that browsers support sharing of user data with other browsers?

That said, I think that when we address registration we should ensure that it is possible for payment apps to keep themselves synchronized across all of their installed/registered instances. i.e. If I have the same payment app installed on my phone and desktop and add new payment method data to one the other should get updated and this should not be something user agents can control/prevent unless there emerges a very strong security or privacy argument against it

mattsaxon:
@adrianhopebailie, response made within issue #38

msporny:
Withdrawn in favor of pull request #96.

PR #74, opened 2016-03-16
Remove note and add description for diagram

halindrome:
The diagram as it stood was not accessible.  Also, it was in a "note", which was not really needed and made it format poorly in a printed version.  Since notes are non-normative by definition, also marked the section as informative so that it is tagged accordingly.

This should really have a more accessible description if it is important.  The right way to do that is to mark up the SVG.  I have not done that at this time.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -478,11 +478,10 @@
       </section>
 
       <section>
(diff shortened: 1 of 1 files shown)


adrianba:
Merge as f107f5741bc391b7ef0c4d2c9cadbf89c829798b

Issue #39, opened 2016-03-14
How are web-based payment apps supported?


msporny:
Migrating issue from https://github.com/w3c/webpayments/issues/16

How do we support Web-based payment apps? These instruments require a user to:
1. Start on the merchant website
2. Click buy
3. Select the web-based instrument
4. Navigate to a 3rd party website to approve the transaction
5. Return to the merchant website with a token, approval code, or other proof that the transaction has been approved.

I suggest that we support web-based payment instruments and local payment instruments using the same mechanism. There is a URL that is used to invoke the payment instrument. Local Android payment instruments would use an Intent URL (or something similar). Cloud-based instruments would use an HTTP-based URL. Both would have a callback URL (either another Intent URL or an HTTP-based URL).

I admit that this proposal is a bit hand wavy as I do not have a great deal of experience routing messages from a browser to an intent-based URL and back to the browser.

adrianhopebailie:
@msporny Can you edit this to use the terminology "payment app" as opposed to "payment instrument"? It seems more appropriate now based on the proposed architecture.

adrianhopebailie:
While I can understand that the integration between user agents and native payment apps is a complex issue (see #50) I believe that we should have some minimum conformance criteria for user agents and web based payment apps.

This could be as simple as mandating that the payment request is submitted to the app URL via a POST. 

We could define two standard response types:
1. HTML - in which case the user agent renders the HTML (the app UI)
2. Some other type that indicates to the user agent that the app is not web based but provides some signal that enables the user agent to invoke the native app.

I need to flesh this idea out a little more :smile:

burdges:
I think web based payment apps might necessary be rather similar to the way they work today:

After the pay button is hit, the user gets forwarded to the payment app selector, which forwards them to the web based payment app.  After the user approves the transaction, the web based payment app forwards them back to the payees fulfillment page.  

We must watch anything fancier closely for same origin violations.  In this space, the spec's real value is probably standardizing the data formats to enable payees to support more web payment apps more easily, not enabling some additional interaction.

wayneca:
Since Web pages don't have access to tamper proof execution on the user device, a Web page payment app would likely execute secure functions on a server.  One means of doing that could be:
[...]

Issue #50, opened 2016-03-14
How are payment requests and responses passed between the browser and third-party native wallets?


msporny:
Migrating from https://github.com/w3c/webpayments/issues/42:

@rsolomakhin sort of described it here:

I think that native apps will connect through OS-specific means, like intents on Android. Here's how user agent code would look like on Android, I imagine:

``` java
Intent intent = new Intent("org.w3.intent.action.PAY", Uri.parse("https://bobpay.xyz"));
intent.putExtra("Details", "{\"price\": 5500, \"currency\": \"USD\", \"merchant\": \"superstore1\"}");
intent.putExtra("SchemeData", "{\"bobPaySpecificField\": \"foo\"}");
startActivityForResult(intent, 0);
```

The payment app should send back the result as a string (or HashMap) plus the result code. Like so:

``` java
Intent result = new Intent("org.w3.intent.action.PAY");
result.putExtra("InstrumentDetails", "{\n"
        + "  \"cardNumber\": \"4111111111111111\",\n"
        + "  \"nameOnCard\": \"Bob J. Paymentman\",\n"
        + "  \"expMonth\":   \"12\",\n"
        + "  \"expYear\":    \"2016\",\n"
        + "  \"cvv2\":       \"123\"\n"
        +"}");
setResult(Activity.RESULT_OK, result);
finish();
```

However, the spec should give folks some idea of how the integration might happen via a NOTE or similar mechanism to ensure that people know it won't be just the OS vendors providing payment apps

adrianba:
At this point, this is an implementation detail. In the absence of an interoperable approach to this (and nobody has yet proposed one) we should not favour particular browsers or operating systems by including their proprietary solutions in cross-platform specs.

msporny:
> In the absence of an interoperable approach to this (and nobody has yet proposed one)

There have been various proposals.

Rouslan has suggested something that is at least payment-app agnostic above. 

Anders and I suggested Intent URLs: https://developer.chrome.com/multidevice/android/intents

You could use the Web Payments CG spec's special domain / callback URL approach for web-based payment apps and register a handler for special payment URLs.

We could settle on intents and ask the other vendors to implement intent URLs, that would be one well spec'd interoperable approach.

> we should not favour particular browsers or operating systems by including their proprietary solutions in cross-platform specs.

By not specifying anything we are favoring the proprietary implementations of the largest browser vendors and operating system vendors.

I don't think our approach should be "do nothing" (aka leave it unspecified).

[...]

PR #77, opened 2016-03-16
Add a reference to issue #39

msporny:
Add a reference to issue #39.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -459,6 +459,18 @@
             later be resolved by the <a>user accepts the payment request algorithm</a> through interaction with the user interface.
           </li>
         </ol>
+
+
+        <p class="issue" data-number="39" title="How are web-based payment apps supported?">
+          How do we support Web-based payment apps? These instruments require
+          a user to: <br/>
+          1. Start on the merchant website<br/>
+          2. Click buy<br/>
+          3. Select the web-based instrument<br/>
+          4. Navigate to a 3rd party website to approve the transaction<br/>
+          5. Return to the merchant website with a token, approval code, or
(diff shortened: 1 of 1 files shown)


zkoch:
I don't think issues around registration belong in the paymentrequest spec. @ianbjacobs addresses this in his draft blog post, so I think that might be good enough for now.

adrianhopebailie:
@zkoch I don't think this is about registration it's about integration between the user agent and a web based payment app.

@msporny For consistency, I think the issue markers for issues #50, #39 should follow a similar format as they both relate to payment app interface (one for native apps and the other for web based apps).

I would suggest the following text and would include this at an appropriate point in the section [User Agent delegates payment request algorithm](https://w3c.github.io/browser-payment-api/specs/paymentrequest.html#user-agent-delegates-payment-request-algorithm) and possibly another reference in the [complete()](https://w3c.github.io/browser-payment-api/specs/paymentrequest.html#complete) section as this refers to messages being passed back to the payment app.

Suggested text:

``` html
<p class="issue" data-number="39" title="It is unclear how the user agent will interface with 
a web-based payment app.">The architecture document suggests that payment apps may 
take numerous forms, including as web-based apps. This specification should describe how 
the user-agent will pass the payment request data and the complete signal to a web-based 
payment app and also how it will receive the payment response from the payment app.</p>
```

msporny:
> I would suggest the following text and would include this at an appropriate point in the section User Agent delegates payment request algorithm

Done, used your text and preferred location.

adrianba:
Merged as 188ae18bff7d6fe592c866f09d9c2900d4beb2b0

Issue #45, opened 2016-03-14
Should Payment Method Identifiers and Messages be expressed using a Linked Data Vocabulary?


msporny:
Migrating from https://github.com/w3c/webpayments/issues/29:

The [Payment Method Identifiers](http://wicg.github.io/paymentrequest/specs/method-identifiers.html) spec states:

> The PaymentRequest API requires that merchants supply a list identifiers for supported payment methods. This document defines those identifier strings and how they are created.

The [Payment Request Architecture](http://wicg.github.io/paymentrequest/specs/architecture.html) spec states:

> we expect some message definitions to be shared amongst different payment apps.

The two sets of spec proposals agree that:
- There should be a list of identifiers for payment methods.
- There should be shared attributes across payment apps and methods.
- There should be short names for payment methods.
- We should support distributed extensibility.

This sounds an awful lot like a Linked Data vocabulary. Is it?

This is related to #11.

Spec refs:
http://wicg.github.io/paymentrequest/specs/method-identifiers.html#introduction
http://web-payments.github.io/web-payments-messaging/#payment-instrument-registration

adrianba:
This is a duplicate of issues #11 and #10. Today the answer to the question "Is it Linked Data vocabulary" is no. If you prefer the answer to be yes, we need a proposal for the Method Identifiers specification.

msporny:
> This is a duplicate of issues #11 and #10.

Issue #11 is just about Payment Method Identifiers. 

Issue #10 is just about short codes for payment method identifiers.

This issue is about PMIs and the content of payment messages. It takes a more holistic view around how we're going to achieve a solution for #10 and #100. Specifically, this issue asks whether or not some/all of the 'key's and 'value's of the objects map to URLs.

I do agree that using a Linked Data Vocabulary expressed as JSON-LD would resolve #10, #11, and #45 in one fell swoop, but I don't think the rest of the WG is there yet, so would rather keep these issues separate for now (and assert that while there is overlap, the issues don't duplicate one another).

ianbjacobs:
@msporny,

"Specifically, this issue asks whether or not some/all of the 'key's and 'value's of the objects map to URLs."

Could you say more about what requirement(s) you have in mind?

Ian

msporny:
> Could you say more about what requirement(s) you have in mind?

The requirements are at the top of this issue. I'll re-state them again and link the 'keys' and 'values' text above to the requirements stated at the beginning of the issue:
[...]

PR #79, opened 2016-03-16
Add reference to issue #45.

msporny:
Add reference to issue #45.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -89,7 +89,7 @@
         <a href="https://www.w3.org/community/wicg/">Web Platform Incubator Community Group</a>.
       </p>
     </section>
-    
+
     <section class='informative'>
       <h2>Introduction</h2>
         <p>Buying things on the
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
@msporny I think this would be better placed in the Payment Method Identifiers spec and even better as a concrete proposal alongside the two current proposals.

https://w3c.github.io/browser-payment-api/specs/method-identifiers.html

msporny:
Issue #45 isn't just about Payment Method Identifiers, but Payment Requests and responses as well. Happy to put a concrete proposal together alongside the current proposals in the Payment Method Identifier spec, but that's not the only place this issue marker belongs. 

I'm not going to be able to put together a concrete proposal for the Payment Method Identifier spec in the next 48 hours (and process all the other requests for PR changes). Happy to re-word this issue, but I need some guidance on what would be acceptable to the group. Would this work?

> There is a discussion on whether or not payment messages and the data they contain should have a more formal and extensible description format than WebIDL, such as JSON-LD or JSON Schema. This is being tracked in issue #45.

Would something like that work?

adrianhopebailie:
@msporny that suggested text makes no sense to me because it doesn't reflect what is being discussed in #45.

At a minimum I think you need to re-title the issue thread. 

I think what you are suggesting is that all payment method data keys and payment method identifiers themselves should be rooted in a linked data vocab and (maybe) that the payment data object passed to the payment request should be expressed in something more extensible than WebIDL.

adrianhopebailie:
@msporny 

I'd like to close this PR unless you still wish to address my comments above?

Issue #46, opened 2016-03-14
Should a payment method identifier (URL) resolve to a machine readable resource that describes it?


msporny:
Migrating from https://github.com/w3c/webpayments/issues/30:

> If the payment app identifier or payment method identifier is a URL, should there be a resource that can be fetched from the URL? 

Yes. We should suggest that the resource should be machine-readable (expressed as JSON-LD, at a minimum) for the reasons specified here:

http://www.w3.org/TR/json-ld/#introduction

> Is there a need to describe the payment method at the URL or provide some other information?

We should not require it, but encourage it. 

At a bare minimum, you could publish a human readable string in various languages for the payment method / app. You could also publish the payment method and/or apps branding image.

You could also make it so that a payment method URL may list the payment apps that are capable of processing the payment method. This would make it easy for user agents to retrieve a payment app if one isn't available in the user agent already. It could address the "no available payment apps" use case. I'm not suggesting that's how it would work, just that it seems like there are enough use cases to think more deeply about it.

kirkalx:
I believe that any resource pointed to should only be explanatory (e.g. pointing to an appropriate URL about the web payments API). There should be no sense that the URL owner has any additional powers over that payment method than ownership of that particular URL gives them. My reasoning is that for decentralized currencies like bitcoin, there is no one central domain name that would be seen as representing bitcoin (e.g. the traditional choice bitcoin.org is now owned by figures who are quite controversial in the bitcoin community). It might be tempting for these to be given a URL directly owned by the working group, which again indicates that there should be no additional power associated with ownership of the URL.

adrianhopebailie:
@kirkalx consider that there may be multiple payment methods for paying via Bitcoin proposed by various people. If the WG takes a position on the "best" way for that to work it could publish a payment method spec at a w3.org owned URL and hope that the Web community favors that one but we can't enforce that any more than we can enforce the card networks to adopt our basic card payment spec.

Put another way, I disagree with this:

> There should be no sense that the URL owner has any additional powers over that payment method than ownership of that particular URL gives them.

If the Bitcoin community doesn't trust the owners of bitcoin.org then I'd expect them to not use payment methods where the identifier is a bitcoin.org URL but to self-select a better payment method managed by an entity they trust (which could be w3.org).

kirkalx:
[...]

PR #80, opened 2016-03-16
Add reference to issue #46.

msporny:
Add reference to issue #46.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -282,6 +282,15 @@
             ["visa", "bitcoin", "bobpay.com"]
           </pre>
 
+          <p class="issue" data-number="46" title="Should a payment app identifier (URL) or a payment method identifier (URL) resolve to a machine readable resource that describes it?">
+            If the payment app identifier or payment method identifier is a
+            URL, should there be a resource that can be fetched from the URL?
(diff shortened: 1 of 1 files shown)


adrianba:
This issue should be flagged in the method-identifiers.html document like issues #10 and #11.

adrianba:
See 3c528926467fe574db98926e55037840608f83e0

msporny:
-1 for closing this issue. I raised this PR to help the reader understand that we're still discussing payment method identifiers while they're reading the Payment Request API spec. They wouldn't understand that payment method identifiers are under active discussion unless they went and read the Payment Method spec and I don't think we can hope that folks that read the specs read all of them to understand the lay of the land. We can't get the WG to read all the specs as it is.

ianbjacobs:
I agree that we need readers to understand that there are active discussions on a variety of topics. But I think we should make life simpler for readers by conveying that a small number of times, and not overloading them with redundant issue flags.

So for example, in the Payment Request API spec section "3. Dependencies" we would add a statement like: "Note: Each of the specifications published by this Working Group includes issue flags where the group is seeking input."

I also expect to write a blog post framing all the publications and letting people know that there are issues and questions for feedback in each of the specs. (I'll discuss that at today's call.)

Ian

Issue #47, opened 2016-03-14
Should a payment request be just data, or a programmable object?


msporny:
Migrating from https://github.com/w3c/webpayments/issues/36:

The paymentRequest API asserts that a payment request is a programmable object:

http://wicg.github.io/paymentrequest/specs/paymentrequest.html#paymentrequest-interface

that is instantiated like so

``` javascript
var payment = new PaymentRequest(supportedInstruments, details, options, schemeData);
```

and can have methods called on it like so:

``` javascript
payment.addEventListener("shippingAddressChange", function (changeEvent) {
    // Process shipping address change
});
```

The Web Payments CG's Browser API asserts that a payment request is just data:

http://wicg.github.io/web-payments-browser-api/#processing-a-payment-request

and that data is processed by functions:

``` javascript
var request = {
  type: 'PaymentRequest',
  description: 'Payment to ExampleMerch for widgets',
  acceptablePayment: {
    paymentMethod: 'https://w3id.org/payment-methods#Visa',
    paymentAmount: {
      amount: '4.35',
      currency: 'USD'
    }
  }
};

// request payment and get a promise that will resolve to
// a payment acknowledgement
var payment = navigator.payment.request(request);

// called when the payment request has been processed
payment.then(function(acknowledgement) {
  // payment request was at least partially processed, do something with the
  // acknowledgement, such as check its approval code or forward an approval
  // token to actually execute the payment
}).catch(function(err) {
  // payment request failed
});
```

These are two different general design approaches for the browser API:
1. There is data, and functions act on that data (Web Payments CG approach).
2. There are objects that consist of data, and verbs that hang off of those objects (paymentRequest approach).

So, what is our design approach for the browser API?

adrianba:
The design of the PaymentRequest API allows events to be fired in any order. A promise can only be fulfilled or rejected. I recommend that this issue be closed.

adrianhopebailie:
@adrianba it's unclear to me how your comment relates to the question being asked. I assume you mean that the current design is required because it uses events to notify the website of user choices with regard to things like the shipping address and therefore an event emitting object is required.

I don't dispute that, I agree that we do need a valid event emitter to support the current design.

However, my understanding of the crux of this question is whether:
1. The website should construct the payment request data in the exact form that it will be passed to the payment app OR
[...]

Issue #16, opened 2016-03-03
Use navigator.payments singleton, factory method, or PaymentRequest constructor


adrianba:
This issue comes from WICG/paymentrequest#42.

From the spec:

> There is an open issue about how a payment request is initiated. The options proposed include using a singleton at `navigator.payments`, using a factory method to create an instance, or using a constructor as currently described in the specification.

adrianhopebailie:
In TAG review @triblondon said:

> I prefer a constructor. I'd definitely say no to a singleton unless it was a solid certainty that we'll never want a UA to be processing multiple payments at a time, since the instance provides a means of storing state which seems like it would become more useful as the API is extended in future.

adrianhopebailie:
@msporny said:

> To clarify, the singleton would instantiate a Promise and return it. You'd be able to run multiple payment requests in a page. I'm not strongly arguing for/against - just clarifying.

adrianhopebailie:
Part of the problem with using the constructor is that we have no namespaced place to hang other API methods that we are still working on for things like registration, getting the PaymentRequest and submitting the PaymentResponse as a web based payment app.

A factory pattern might be a good compromise?

## Proposal

Put the `PaymentRequest` constructor behind a factory method.

Sample javascript:

``` javascript
var payment = navigator.payments.newPaymentRequest(supportedMethods, details, options, data);
payment.addEventListener("shippingaddresschange", function (changeEvent) {
    // Process shipping address change
});

payment.show().then(function(paymentResponse) {
  // Process paymentResponse
  // paymentResponse.methodName contains the selected payment method
  // paymentResponse.details contains a payment method specific response
  paymentResponse.complete(true);
}).catch(function(err) {
  console.error("Uh oh, something bad happened", err.message);
});
```

triblondon:
Couldn't you use static methods of the `PaymentRequest` constructor if you wanted to create new non-instance methods in future?  But regardless the factory seems fine.  The web platform has pretty much no consensus or consistency in this regard so there isn't a strong convention, imho.

sideshowbarker:
> The web platform has pretty much no consensus or consistency in this regard so there isn't a strong convention, imho

Finding out what the recommended best practices are for details like this is a big part of the motivation of asking for close review from the TAG. But if the answer really is that there really is no consensus or consistency in this particular case, then that actually is also useful to know.

triblondon:
It would be worth raising this as a specific question on the TAG mailing
[...]

Issue #15, opened 2016-03-03
Combine API parameters into a single request object + options


adrianba:
This issue comes from WICG/paymentrequest#41.

There is an open issue about whether `supportedMethods`, `details`, and `data` should be combined into a single object.

adrianhopebailie:
In the TAG review @triblondon said:

>  I'm strongly in favour of these being merged into a single object, which provides greater extensibility, and is more in line with the direction of travel of JavaScript now that destructuring makes dealing with large object structures easy.

sideshowbarker:
Here as with #121 the comments are actually from @triblondon

msporny:
I agree strongly with @triblondon wrt. merging arguments into a single object. I was going to submit a PR to do just that after FPWD was out, especially since it would be nice if the messages looked the same in the Web Payments HTTP API and the Browser API. You can't easily split arguments out in an HTTP call unless you put some stuff in as URL parameters and the rest in the body.

The Web Payments Community Group specs used the single object pattern so that Web Payments messages could be shared between the Browser API and the HTTP API.

halindrome:
+1 to having a single object.

adrianhopebailie:
@sideshowbarker said:

> Thanks @triblondon. This seems like an important issue, so I would like to hear from more of the TAG (e.g., @mnot, @travisleithead) about whether this is how the TAG recommends that groups design JavaScript APIs. Do either https://www.w3.org/TR/api-design/ or https://w3ctag.github.io/design-principles/ provide some relevant guidance on this?

adrianhopebailie:
I'd like the inputs to be split between 2 parameters because some are part of the payment request that goes to the app and others are explicitly intended to only be consumed by the browser (such as `requestShipping`).

Since the line item stuff in `PaymentDetails` is purely for display purposes by the browser, especially if we pull the total out into it's own item, I'd not include it in the payment request that goes to the payment app.

## PROPOSAL
- Move `items` and `shippingOptions` from `PaymentDetails` into `PaymentOptions`
- Drop the `supportedMethods` parameter.
- Define a `PaymentMethod` object which has:
  - A required member called `identifiers` which is a sequence of strings, each a payment method identifier. 
  - A `data` member which is a generic object and contains the payment method specific data for the payment method.  
- Add a `supportedMethods` member to `PaymentDetails` which is a sequence of `PaymentMethod` objects.

adrianhopebailie:
[...]

PR #81, opened 2016-03-16
Add reference to issue #47.

msporny:
Add reference to issue #47.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -249,6 +249,30 @@
         <a>user agent</a> while the user is providing input before approving or denying a payment request.
       </p>
 
+      <div class="issue" data-number="47" title="Should a payment request be just data, or a programmable object?">
+        Instead of a programmable object, it has been suggested that a payment
+        request should be just pure data that can be operated on by methods
+        like so:
+        <pre class="example highlight">
+// data-only payment request object example
+var request = {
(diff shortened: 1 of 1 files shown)


specs/paymentrequest.html
```diff
@@ -249,6 +249,30 @@
...
+  }
+};
+
+// request payment and get a promise that will resolve to
+// a payment acknowledgement
+var payment = navigator.payment.request(request);
+        </pre>
+      </div>
+
       <p>The following example shows how to construct a <a><code>PaymentRequest</code></a> and begin the
```
adrianhopebailie, line 252: @msporny Can you reword this to be a lot simpler and leave the deeper explanation for users to read in the issue itself? I would also suggest placing this issue marker alongside those for #15 and #16 which also relate to the shape of the API.

Suggestion:

``` html
<div class="issue" data-number="47" title="Should a payment request be just data, or a 
programmable object?"> There is an open issue about whether the payment request should 
be a programmable object or should be just pure data that can be operated on by methods.</div>
```

msporny:
@adrianhopebailie wrote:

> Can you reword this 

Done and moved to where you suggested.

adrianba:
Merged as 814702c5f937ebf66f075071ae13f29b20a13535

Issue #48, opened 2016-03-14
Should list of accepted payment methods be strings or objects?


msporny:
Migrated from https://github.com/w3c/webpayments/issues/37:

The paymentRequest API asserts that the list of accepted payment methods that a payee exposes to the payer should be a list of strings:

http://wicg.github.io/paymentrequest/specs/paymentrequest.html#paymentrequest-interface

For example:

``` javascript
["visa", "bitcoin", "bobpay.com"]
```

The Web Payments CG Browser API asserts that the list should be composed of objects:

http://wicg.github.io/web-payments-browser-api/#processing-a-payment-request

For example:

``` javascript
acceptedScheme: [{
    scheme: [
      'https://w3id.org/payment-schemes#Visa',
      'https://w3id.org/payment-schemes#Mastercard',
      'https://w3id.org/payment-schemes#Discover'
    ],
    transfer: {
      amount: '4.35',
      currency: 'USD'
    },
    paymentRequestService: 'https://merchant-psp.example.com/services/getPaymentInfo'
  }, {
    scheme: 'https://w3id.org/payment-schemes#Bitcoin',
    transfer: {
      amount: '0.0177',
      currency: 'BTC'
    },
    destination: '3QJmV3qfvL9SuYo34YihAf3sRCW3qSinyC'
  },
```

Should list of accepted payment methods be strings or objects?

adrianhopebailie:
PR #133 proposes a solution to the issue.

PR #82, opened 2016-03-16
Add reference to issue #48.

msporny:
Add reference to issue #48.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -282,6 +282,39 @@
             ["visa", "bitcoin", "bobpay.com"]
           </pre>
 
+
+          <div class="issue" data-number="48" title="Should list of accepted payment methods be strings or objects?">
+            <p>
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
@msporny Can you reword this to be a lot simpler and leave the deeper explanation for users to read in the issue itself? I would also suggest placing this issue marker alongside those for #15 and #16 which also relate to the shape of the API.

Suggestion:

``` html
<div class="issue" data-number="48" title="Should list of accepted payment methods be 
strings or objects?"> There is an open issue about whether the list of supported payment 
methods should be passed to the user agent as a simple sequence of strings or as a more 
complex and flexible object structure.</div>
```

msporny:
@adrianhopebailie wrote:

> Can you reword this to be a lot simpler 

Done. Used your text and requested location.

adrianba:
Merged as fcfe1d13bc8955a8ad8e38781d60481e17863f27

Issue #49, opened 2016-03-14
Should the payment request contain line item details?


msporny:
Migrating from https://github.com/w3c/webpayments/issues/38:

The paymentRequest API currently includes a Payment Details object in a payment request:

http://wicg.github.io/paymentrequest/specs/paymentrequest.html#paymentdetails-dictionary
http://wicg.github.io/paymentrequest/specs/paymentrequest.html#paymentrequest-interface

The Web Payments CG's Browser API specifically does not provide such an interface because small and large retailers have stated that they do not want that information leaked outside of their customer-merchant environment.

Does introducing this line-item list create an issue w/ the "private" customer-merchant environment? Will most merchants opt to not include line item data?

adrianba:
Line items are entirely optional so merchants can provide what they want. At most you need to provide one item with the total amount.

halindrome:
Note that I continue to object to having one of the line items be the total.  The total should be a separate entry in the data structure.(see issue #18)

adrianba:
> Note that I continue to object to having one of the line items be the total. The total should be a separate entry in the data structure.(see issue #18)

I wasn't taking a view on that here. The current proposal is modeled on an existing (non-web) API. Issue #18 is called out in the spec in section 6.

msporny:
Walmart and Target, for example, have specifically stated that they do not want to include line item data. Will most merchants opt to not include line item data once they find out that the browser vendors could mine that data? If so, that would be a strong signal to not include line item data in the API.

adrianhopebailie:
I think this is a red herring.

The list items are not intended for showing a basket of goods (although I guess there is nothing stopping a merchant doing that) but rather the set of final "sub-totals" that constitute the final total. i.e. Things like shipping, discounts, etc

halindrome:
I disagree with your conclusion @adrianhopebailie - that list of items is 1) called items, and 2) the examples include the information about what is the basket.

adrianhopebailie:
Which examples are you referring to @halindrome ?

This one from the spec has no details of the basket of goods:

``` javascript
{
  "items": [
    {
      "id": "basket",
      "label": "Sub-total",
      "amount": { "currencyCode": "USD", "value" : "55.00" }, // US$55.00
    },
    {
      "id": "tax",
      "label": "Sales Tax",
      "amount": { "currencyCode": "USD", "value" : "5.00" }, // US$5.00
    },
    {
      "id": "total",
      "label": "Total due",
[...]

PR #83, opened 2016-03-16
Add reference to issue #49.

msporny:
Add reference to issue #49.


adrianhopebailie:
Line item details are optional and merchants are able to simply provide a total if they don't want to leak shopping cart data. Not sure this is actually an issue?

adrianba:
I recommend closing this PR (and I've also proposed closing the underlying issue [#49](https://github.com/w3c/browser-payment-api/issues/49#issuecomment-211511678)).

PR #84, opened 2016-03-16
Add a reference to issue #50.

msporny:
Add a reference to issue #50.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -779,13 +779,6 @@ dictionary PaymentOptions {
         };
       </pre>
 
-      <p class="issue" data-number="50" title="How are payment requests and responses passed between the browser and third-party native wallets?">
-        How are payment messages routed between the browser and native payment
-        apps? The spec should give developers some idea of how the integration
-        might happen via a NOTE or similar mechanism to ensure that people know
-        it won't be just the OS vendors providing payment apps.
-      </p>
-
       <p>
         A <code><dfn>PaymentResponse</dfn></code> is returned when a user has selected a payment method and
         approved a payment request. It contains the following fields:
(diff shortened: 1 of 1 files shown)


specs/paymentrequest.html
```diff
@@ -779,6 +779,13 @@ <h2 id="state-transitions">State transitions</h2>
...
       </pre>
 
+      <p class="issue" data-number="50" title="How are payment requests and responses passed between the browser and third-party native wallets?">
+        How are payment messages routed between the browser and native payment
+        apps? The spec should give developers some idea of how the integration
+        might happen via a NOTE or similar mechanism to ensure that people know
+        it won't be just the OS vendors providing payment apps.
+      </p>
+
       <p>
```
adrianhopebailie, line 782: @msporny For consistency, I think the issue markers for issues #50, #39 should follow a similar format as they both relate to payment app interface (one for native apps and the other for web based apps).

I would suggest the following text and would include this at an appropriate point in the section [User Agent delegates payment request algorithm](https://w3c.github.io/browser-payment-api/specs/paymentrequest.html#user-agent-delegates-payment-request-algorithm) and possibly another reference in the [complete()](https://w3c.github.io/browser-payment-api/specs/paymentrequest.html#complete) section as this refers to messages being passed back to the payment app.

Suggested text:

``` html
<p class="issue" data-number="50" title="It is unclear how the user agent will interface with 
a native payment app.">This specification should describe how the user-agent will pass the payment request data and the complete signal to a native payment app and also how 
it will receive the payment response from the payment app.</p>
```

msporny:
@adrianhopebailie wrote:

> For consistency, I think the issue markers for issues #50, #39 should follow a similar format as they both relate to payment app interface

Done. Used your language and preference to issue marker location.

adrianba:
Merged as 1988c408f5aff6e63ccbb032133e9a310905466c

PR #85, opened 2016-03-16
Add a reference to issue #51.

msporny:
Add a reference to issue #51.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -242,6 +242,23 @@
         };
       </pre>
 
+      <div class="issue" data-number="51" title="Should the payment API be more conversational or less conversational?">
+        <p>
+          A more conversational flow for the API has been suggested by
+          Dave Longley. Instead of the above, developers could do something
+          like this instead:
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
@msporny Can you reword this to be a lot simpler and leave the deeper explanation for users to read in the issue itself? I would also suggest placing this issue marker alongside those for #15 and #16 which also relate to the shape and behavour of the API.

Suggestion:

``` html
<div class="issue" data-number="51" title="How should the API interact with the website
when exchanging data such as shipping address and options"> There is an open issue about 
whether the current pattern of using events for exchange of data between the user agent and
the website is the best design and an alternative has been proposed in this issue thread.</div>
```

msporny:
@adrianhopebailie wrote:

> Can you reword this to be a lot simpler 

Done with minor changes to your language. The issue has been moved to where you wanted it to go.

adrianba:
Merged as 7ef4a69eac8a652ebcdc19672a0e541e393aa2e1

Issue #56, opened 2016-03-14
Should a website be able to provide a label for the "Buy" or "Checkout" button displayed in the payment app?


msporny:
Migrated from https://github.com/w3c/webpayments/issues/66 and  https://github.com/WICG/paymentrequest/issues/46:

@adrianba:

> It may help users understand what they are accepting if the web site is able to label the "accept" button. For example, if a user is about to "Buy" something, "Reserve" something, "Subscribe" to something, etc.

@mattsaxon:

> This is linked to the comment I just made on issue #65
> +1 to distinctions as suggested by @adrianba 

@ianbjacobs:

> It seems there are several topics here:
> 1. The API may be used in a variety of user interactions, and the entity that displays the relevant payment apps may need a label to communicate the interaction to the user. Note that these labels may not be in English, and that the specification probably cannot prevent misuse of labels.
> 2. The API may be used in a variety of flows, such as "pay", "register", or "reserve". There may be hacky ways to implement the different scenarios, like using amount=0 to imply "register". However, we should consider whether we want to define a small number of verbs that become part of the data exchanged with the payment application. We will not be able to address every possible scenario, but I have now heard three that sound like it could be useful for the merchant to be able to pass on to the payment application.
> 
> Furthermore, it seems there is a relationship between this topic and a topic we discussed previously that translated in the following charter language: "The Working Group will consider support for deferred payment execution to enable use cases where the actual execution of a payment requires steps that are out-of-band with respect to the message flows and APIs defined by the Working Group." That was a use case where the user chose a push payment instrument, but the merchant did not want the payment to be completed before the merchant regained control of the flow.
> 
> Please let me know if this makes sense:
> - The distinction between "what the user sees" (label) and "what the merchant wants to accomplish"
> - We may wish to define a set of verbs so the merchant can communicate what they want to accomplish.

adrianhopebailie:
My original response to this issue was the following:

:-1:  unless we have a pre-defined set of verbs the payee can choose from or these are automatically inferred by the payment app based on the terms of the request.

The ability for the entity requesting payment to manipulate the user interface presented to the payer should be considered VERY carefully. This is heavily locked down in many existing payment systems today for good reason.

**EXAMPLE:** A developer writing custom firmware for physical card acceptance devices is unable to use custom prompts when the device is requesting input from the user. The reasoning is that the developer could publish a malicious application that prompts a user to input their PIN when the data is not being captured securely and the developer is therefor able to steal the user's PIN.

[...]

PR #101, opened 2016-03-29
Currency amount

adrianhopebailie:
Updated definition to allow for flexible currency identifiers and not
allow negative amounts

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -295,17 +295,17 @@
                 {
                   "id": "basket",
                   "label": "Sub-total",
-                  "amount": { "currencyCode": "USD", "value" : "55.00" }, // US$55.00
+                  "amount": { "currency": "USD", "amount" : "55.00" }, // US$55.00
                 },
                 {
                   "id": "tax",
                   "label": "Sales Tax",
-                  "amount": { "currencyCode": "USD", "value" : "5.00" }, // US$5.00
+                  "amount": { "currency": "USD", "amount" : "5.00" }, // US$5.00
                 },
                 {
                   "id": "total",
                   "label": "Total due",
-                  "amount": { "currencyCode": "USD", "value" : "60.00" }, // US$60.00
+                  "amount": { "currency": "USD", "amount" : "60.00" }, // US$60.00
                 }
               ]
             }
@@ -578,8 +578,8 @@
       <h2>CurrencyAmount</h2>
       <pre class="idl">
 dictionary CurrencyAmount {
-  required DOMString currencyCode;
-  required DOMString value;
+  required DOMString currency;
+  required DOMString amount;
 };
       </pre>
       <p>
@@ -587,19 +587,20 @@ dictionary CurrencyAmount {
         The following fields MUST be supplied for a <a><code>CurrencyAmount</code></a> to be valid:
       </p>
       <dl>
-        <dt><code><dfn>currencyCode</dfn></code></dt>
+        <dt><code><dfn>currency</dfn></code></dt>
         <dd>
-          <code>currencyCode</code> is a string containing a three-letter alphabetic code for the
-          currency as defined by [[!ISO4217]]. For example, <code>"USD"</code> for US Dollars.
+          <code>currency</code> is a string containing a currency identifier. The most common
+          identifiers are three-letter alphabetic codes as defined by [[!ISO4217]] (for example,
+          <code>"USD"</code> for US Dollars) however any string is considered valid and
+          <dfn data-lt="user agents">user agents</dfn> MUST not attempt to validate this string.
         </dd>
-        <dt><code><dfn>value</dfn></code></dt>
+        <dt><code><dfn>amount</dfn></code></dt>
         <dd>
           A string containing the decimal monetary value. If a decimal separator is needed then the string
-          MUST use a single U+002E FULL STOP character as the decimal separator. The string MUST begin
-          with a single U+002D HYPHEN-MINUS character if the value is negative. All other characters must
+          MUST use a single U+002E FULL STOP character as the decimal separator. All other characters must
           be characters in the range U+0030 DIGIT ZERO (0) to U+0039 DIGIT NINE (9).
           <div class="note">
-            The string should match the regular expression <code>^-?[0-9]+(\.[0-9]+)?$</code>.
+            The string should match the regular expression <code>^[0-9]+(\.[0-9]+)?$</code>.
           </div>
         </dd>
       </dl>
@@ -607,8 +608,8 @@ dictionary CurrencyAmount {
       <p>The following example shows how to represent US$55.00.</p>
       <pre class="example highlight">
 {
-  "currencyCode": "USD",
-  "value" : "55.00"
+  "currency": "USD",
+  "amount" : "55.00"
 }
       </pre>
     </section>
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -295,17 +295,17 @@
                 {
                   "id": "basket",
                   "label": "Sub-total",
-                  "amount": { "currency": "USD", "amount" : "55.00" }, // US$55.00
+                  "amount": { "currency": "USD", "value" : "55.00" }, // US$55.00
                 },
                 {
                   "id": "tax",
                   "label": "Sales Tax",
-                  "amount": { "currency": "USD", "amount" : "5.00" }, // US$5.00
+                  "amount": { "currency": "USD", "value" : "5.00" }, // US$5.00
                 },
                 {
                   "id": "total",
                   "label": "Total due",
-                  "amount": { "currency": "USD", "amount" : "60.00" }, // US$60.00
+                  "amount": { "currency": "USD", "value" : "60.00" }, // US$60.00
                 }
               ]
             }
@@ -579,7 +579,7 @@
       <pre class="idl">
 dictionary CurrencyAmount {
   required DOMString currency;
-  required DOMString amount;
+  required DOMString value;
 };
       </pre>
       <p>
@@ -594,7 +594,7 @@ dictionary CurrencyAmount {
           <code>"USD"</code> for US Dollars) however any string is considered valid and
           <dfn data-lt="user agents">user agents</dfn> MUST not attempt to validate this string.
         </dd>
-        <dt><code><dfn>amount</dfn></code></dt>
+        <dt><code><dfn>value</dfn></code></dt>
         <dd>
           A string containing the decimal monetary value. If a decimal separator is needed then the string
           MUST use a single U+002E FULL STOP character as the decimal separator. All other characters must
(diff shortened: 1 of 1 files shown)


halindrome:
Wait.... if the item list is truly just a holder for various items, then it should permit negative amounts to allow for discounts/credits/exchanges as line items, shouldn't it?

adrianhopebailie:
@halindrome the group's consensus was to follow the lead of ISO20022 an not have negative amounts. The only allowed characters for the amount field are digits and a decimal marker - https://github.com/w3c/webpayments/issues/57

adrianhopebailie:
The way this is done in most payment protocol is that the "settlement effect" of the amount is determined by the context of where the amount is encountered.

One way to do this is to have separate places to put debits and credits another way is to use transaction type to imply if this is a debit or credit.

I'm inclined to support the latter as I think this would also solve https://github.com/w3c/browser-payment-api/issues/56

halindrome:
I understand.   I think the whole item list structure is sort of broken, in that it is basically a list of the things that would appear on a receipt as opposed to information that defines a transaction.  If we need a receipt itemization, then we should provide a holder for that.  If we need a total for an offer, then we should have that.  They should not be intermingled.  But I digress.

mattsaxon:
@adrianhopebailie whilst we want to follow ISO20022, I think where we find exceptions, we agreed to feed this into ISO20022, I agree that for line items, we should have the ability to show -ve amounts.

adrianhopebailie:
@mattsaxon, @halindrome this is not directly related to line items. It's the definition of how we represent currencies and amounts (effectively the CurrencyAmount datatype that we may use in various places in our data). 

These are currently used in the definition of line item BUT if there is a requirement for line items to have negative amounts a flag or similar should be added to the line item interface to accommodate this.

I want to reiterate, that we should have a way for the browser to understand some semantic about the line item and therefor know if it should be displayed as negative or positive. We already have this to some extent with the specialized line items for shipping costs.

This would help us address #56

adrianhopebailie:
Submitted a PR to deal with negative numbers (amongst other things):
https://github.com/w3c/browser-payment-api/pull/111

zkoch:
I'm not sure what value these changes to the names provide, but I'm not opposed, though I don't like doing: amount.amount. Why not leave value?

kirkalx:
Forgive me if this issue has been addressed elsewhere, but if the currency code is to be flexible enough viz. 'any string is considered valid and user agents MUST not attempt to validate this string' then what good is the currency code field? e.g. I could insert the code BTC, intending it to mean bitcoin but someone could interpret it as some currency from the country of Bhutan.

kirkalx:
Further to my comment above, perhaps the URL approach could be used for the currency code as well, e.g. 3 letter iso currency code or URL identifying currency. Apologies if this has already been considered...

adrianhopebailie:
@zkoch the PR addresses the inconsistency between the spec and the resolution made by the group not just in the naming but also the fact that we resolved to use unsigned numbers and allow any string as a currency identifier. See w3c/webpayments#57

On that basis the PR is non-controversial and should be merged. However, if there is a desire by the group to revisit that resolution then let's do that once the spec is at least consistent with the WG consensus as previously recorded.

@kirkalx The browser API is not the place to enforce validation against a) a list we don't control and b) a list that does not include all of the values we know we want to support. I'd be very concerned if we expected user agents to do data validation on this beyond very basic format conformance.

The set of supported currencies SHOULD be defined per payment method so we should leave it for the payment method specifications to restrict the currencies they support if they wish to do so.

This will then be validated by the payment app that processes the payment request.

adrianhopebailie:
PROPOSAL: Use the label "value" instead of "amount" for the compound CurrencyAmount type.

This would imply the following rewording of the resolution previously made by the WG (w3c/webpayments#57):

`The amount object MUST have at least two named properties: "currency" and "amount".`
becomes
`The amount object MUST have at least two named properties: "currency" and "value".`

Please indicate your support with a +1 or -1 and I'll update the PR based on feedback.
(Note that GitHub now supports a +1 or -1 directly on this comment using the +:smile: icon at the top-right)

kirkalx:
> The set of supported currencies SHOULD be defined per payment method so we should leave it for the payment method specifications to restrict the currencies they support if they wish to do so.

Thanks again Adrian, I had thought that should probably be the case after posting my suggestion. Cheers!

dlongley:
> PROPOSAL: Use the label "value" instead of "amount" for the compound CurrencyAmount type.

I think "value" is too generic and also seems potentially misleading to me (does a simple reading cause someone to think it represents the value of the currency, which fluctuates daily?). If we need to change something to avoid `amount.amount`, I'd prefer that we change the former, not the latter.

I'm a general -1 to any modeling of data that uses extremely generic properties that change their meaning based upon the position of a particular object -- when that data is intended to be serialized to a JSON representation and transmitted as such. I believe that's the case here.

Meaning should be better carried in the data itself. Instead this meaning is only present when the data has been parsed into a strongly-typed system that is running code that required reading human-only-consumable documentation. Using more specific terms makes it easier to apply machine-readable context after the fact. As "value" is probably the most generic term that could possibly be chosen here, I'm a big -1. "Amount" is still generic, but at least considerably less so. I'd also be ok with "monetaryValue" (or some other prefix) as a better alternative.

Let's remember the [Rule of Least Power](https://www.w3.org/2001/tag/doc/leastPower.html).  I think we often forget how much context or extra power is required to understand the data when we're making modeling decisions. We may not even consider it all -- which is not a good approach for the Web.

dlongley:
A good article on meaningful data APIs:

http://blog.codeship.com/json-ld-building-meaningful-data-apis/

adrianba:
- I don't support having the spec say `amount.amount` - I don't believe that was what the group intended when this discussion happened in January.
- I believe the current API shape, context, and @dlongley's comments are new information.
- I'm not sure that I support the idea that currency amounts can't be negative numbers. Every finance application I've worked with supports the notion of negative amounts.

adrianhopebailie:
@dlongley 

> I'm a general -1 to any modeling of data that uses extremely generic properties that change their meaning based upon the position of a particular object -- when that data is intended to be serialized to a JSON representation and transmitted as such.

Can you provide a proposal for this interface that meets the criteria you have defined? Yours was the only objection to reverting amount back to value but I suspect you would like to propose an entirely different set of less generic names.

@adrianba 

> I don't support having the spec say amount.amount - I don't believe that was what the group intended when this discussion happened in January

I agree, but this could also be fixed by changing the PaymentItem interface to use a different attribute name. My proposal above to revert back to _amount.value_ has not got consensus.

> I believe the current API shape, context, and @dlongley's comments are new information.

Again I agree, and I think this is sufficient grounds to stick with the original resolution of the group until the new information has been discussed and we have a new proposal that is accepted by the WG.

> I'm not sure that I support the idea that currency amounts can't be negative numbers. Every finance application I've worked with supports the notion of negative amounts.

This is the norm in financial messaging (see ISO20022 and ISO8583). Using a signed amount infers "settlement impact" which gets very confusing when combined with a transaction type. Example, what does a -$50 credit mean? Is it actually a debit. The UI is expected to abstract this complexity away and can use a whole host of heuristics to make this easier for the user to understand such as context, transaction type, locale etc.

There are three issues addressed in this PR:
1. naming the amount/value attribute of this compound type
2. allowing negative numbers
3. allowing more free form currency identifiers

I believe we only have strong objections to the way the first is handled (although no consensus and one supporter). I hear reservations (but mostly questions) about the second and no further issues with the last.

The first feels a little like bike-shedding to me and while I hear @dlongley 's concerns it seems to me he is proposing a far more sweeping change to all attribute names which could be addressed in it's own subsequent PR.

On that basis I have edited the PR to revert to "value" instead of "amount" and request that the editors merge the PR as is (which reflects the previous consensus of the group and a minor terminology change which I expect @dlongley to propose major changes to anyway).

adrianba:
Merged as 563ecd594b5757f5175ee6ede3b41cdd8810d932 per instructions from chairs.

dlongley:
@adrianhopebailie,

> On that basis I have edited the PR to revert to "value" instead of "amount" and request that the editors merge the PR as is (which reflects the previous consensus of the group and a minor terminology change...

I don't consider it a minor terminology change (for the reasons stated in this thread); the consensus of the group and the resolution was to use "amount". If it had said "value", it wouldn't have gotten my vote. See the use of "amount" in: https://github.com/w3c/webpayments/issues/57

In any case, I'm not going to raise an objection that gets in the way of FPWD here, but we should accurately reflect the consensus of the group. We can then debate changing it further. As I said, I'm "ok" with "amount" (it got my vote), because I think it's "sufficiently specific" in this case. But if "amount" is problematic here, I would propose "monetaryValue" rather than the ultra-generic "value".

It's true that I have a general concern that we're being too cavalier in how we go about naming terms in data that is intended to be transported as JSON and shared between various actors (user agents, Payment Apps, etc). Often we end up picking pretty good terms incidentally anyway -- but we should be more deliberate in respecting best practices like the Rule of Least Power.

I don't have any other specific terms that I believe need further bikeshedding right now, but I do hope people will be more mindful of this and if there are other terms that need addressing I (and hopefully others) will bring them up in separate issues in the future.

adrianba:
Removing the ability to have negative amounts from the spec is a significant change and not one that I support. I don't think we will support the FPWD with this change. Applying discount codes is an important aspect that we think should be in the FPWD scope.

dlongley:
@adrianba,

> Applying discount codes is an important aspect that we think should be in the FPWD scope.

Can't these just be modeled using a "type" that indicates "Credit" or "Debit"? This seems like it's a "how should we model it" question, not whether or not it should be possible.

adrianba:
> Can't these just be modeled using a "type" that indicates "Credit" or "Debit"? This seems like it's a "how should we model it" question, not whether or not it should be possible.

Perhaps, but until this change the only way to support it was with a negative amount. Now there is no way to do this. I don't object to proposing a different way but I do object to removing the capability completely.

dlongley:
@adrianba,

> Now there is no way to do this. I don't object to proposing a different way but I do object to removing the capability completely.

That's reasonable. I think everyone wants to support the functionality and some perhaps didn't really consider it supported previously because they weren't thinking you would/should ever use a negative number. So from their perspective there was a gap and continues to be a gap for this feature -- that will eventually be solved with a different solution.

dlongley:
@adrianba,

Perhaps the best thing to do this late in the game would be to add an issue marker that indicates we want to support discounts, etc., and that would be sufficient for FPWD. I doubt there would be an objection to that.

adrianba:
I think the design issues you raised are important. I think we should be thoughtful about how we deal with negative values (removing the possibility without a replacement isn't that). I think the working group couldn't make an informed choice in the absence of context in January. I think #101 is a bad change that we didn't need to make before FPWD and I think we should revert it.

kirkalx:
If this PR is reverted, I believe the currency code liberalization part, or something similar, is necessary.

PR #106, opened 2016-03-30
Resubmitting because the document moved

halindrome:
As per Nick's request, reapplying the agreed upon change from #64.


halindrome:
Technically the definitions moved - my bad.  But you get the idea.

adrianba:
The same change was in #64 merged as e72f051cf9c7460b115ee72f97d90a6747632cd4.

PR #108, opened 2016-03-30
Add option 1b to payment method identifier spec.

msporny:
This PR adds another option 1b to the Payment Method Identifier specification per @adrianhopebailie's request.

The new option:
1. Entirely avoids the need to do Relative URL and Base URL processing while ensuring that short identifiers can be used.
2. Ensures that the Web Payments Working Group only has to maintain a short identifier mapping file, not a complete payment method identifier registry.
3. Encourages hard coding and/or aggressive caching of the short identifier mappings to ensure that we don't fall into the XML namespaces trap.
4. Encourages machine-readable information to be published at payment method identifier URLs but does not require its use.

### DIFF
--- a/specs/method-identifiers.html
+++ b/specs/method-identifiers.html
@@ -74,7 +74,7 @@
         <a href="https://www.w3.org/community/wicg/">Web Platform Incubator Community Group</a>.
       </p>
     </section>
-    
+
     <section class='informative'>
       <h2>Introduction</h2>
         <p>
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
Thanks LGTM

adrianba:
Merged with mark-up changes to correct Respec errors as cf2d0ec13418a3bd475ae776a4550dcd71108b26.

kirkalx:
Would the short identifiers be allocated on a first come, first served basis? Or something more complicated? Some of the same issues as the DNS produces arise but hopefully with particular short identifiers not being that attractive since only developers would see them...

ianbjacobs:
@kirkalx,

> Would the short identifiers be allocated on a first come, first served basis? Or something more complicated?

I believe the proposals on the table all require short names to be approved by the Web Payments WG.
The WG would allocate those short names through a consensus process.

Ian

PR #114, opened 2016-04-02
Specification of required fields

mattsaxon:
I'm suggesting a sequence string field that allows mandatory fields to
be specified and have removed some of the 'required' specifications in
the BasicCardResponse, related to discussion on
#9. I've only left PAN as

required as I think that is a safe assumption, then the merchant can
specify for example that CSC must be supplied for a particular
transaction using the requiredFields sequence in the BaseRequestData
dictionary.

I'm suggesting that if we like this model, we would move this
BaseRequestData dictionary object into the PaymentRequest API spec to
allow other payment methods to extend it.

I'm sure there are other ways to do this and am a newbie to
WebIDL/JavaScript APIs, so please forgive  if I've modelled this
completely wrong for the language. I'll learn!

### DIFF
--- a/specs/basic-card-payment.html
+++ b/specs/basic-card-payment.html
@@ -142,14 +142,21 @@
 
       <table>
         <tr><th>Identifier String</th><th>Description</th></tr>
-        <tr><td>visa</td><td>Visa (Credit, Debit, Delta or Electron)</td></tr>
-        <tr><td>mastercard</td><td>MasterCard and EuroCard</td></tr>
+        <tr><td>visa</td><td>Visa (Credit, Debit and Electron)</td></tr>
+        <tr><td>visa/credit</td><td>Visa Credit</td></tr>
+        <tr><td>visa/debit</td><td>Visa Debit</td></tr>
+        <tr><td>visa/electron</td><td>Visa Electron</td></tr>
+        <tr><td>mastercard</td><td>MasterCard (and EuroCard)</td></tr>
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
I'm +1 on the basic ideas here but unsure if we should have WebIDL in these specs at all.
See #132 for a discussion of that topic in general.

@mattsaxon are you happy to hold on this until we get a resolution on #132 or would you like the editors to merge and then change it later if required?

burdges:
Apologies if this sounds kinda obvious :  We should make it clear the payment app's preferences and defaults dominate here over the merchants, unless the payment app knows it cares about specific preferences from the merchant.  If we do not, then we'll might wind up with sill stuff like a combined BTC and CC wallets that hand out your CC billing address even when buying with BTC.

mattsaxon:
I think we should merge now, the ideas are agnostic to the language we use to define them. If we change documebtation approach, we will do it then.

mattsaxon:
Closed awaiting resolution to PR 133

rsolomakhin:
> I'm suggesting a sequence string field that allows mandatory fields to
> be specified

+1 from me here. This would look sensible to me:

``` javascript
var request = new PaymentRequest([{
    supportedMethods: ["visa", "amex", "mastercard", "discover"],
    data: {
      requiredFields: ["cardNumber", "cardSecurityCode", "billingAddress.postalCode"]
    }
  }]);
```

This should be specified in [Basic Card Payment](https://w3c.github.io/browser-payment-api/specs/basic-card-payment.html), right?

PR #116, opened 2016-04-03
De-duplicate markup for “user agents” def

sideshowbarker:
The current markup defines _user agents_ twice, and causes Respec to emit errors about that.


adrianba:
This was fixed by dd66e74646a58aa4554096768501fc65040cbffe.

PR #117, opened 2016-04-03
Replace non-RFC-2119 “may”s in normative text

sideshowbarker:
Conform to the best practice of never using _may_ in normative text (even uncapitalized) unless when it has its RFC 2119 meaning.


adrianba:
Merged as 05520e93293a6abf97e1da4aeee3830d8e57205d.

PR #120, opened 2016-04-04
Proposal to add negative value support

adrianhopebailie:
An addition to the format proposed by the WG to support negative
amounts. This is an alternative proposal to #111


adrianhopebailie:
This PR separates the proposal by the editors to support negative values from the format originally agreed upon by the WG.

adrianba:
Merged as 118d561c7b554417fcaeb3036c49f90c22bdcd5e with changes from @ianbjacobs.

PR #124, opened 2016-04-07
grammatical updates to the payment method identifier spec

rvm4:


specs/method-identifiers.html
```diff
@@ -305,7 +305,7 @@
         <p>This section relies on several other underlying specifications.</p>
         <dl>
           <dt>URL</dt>
-          <dd>The terms <em>URL</em>, <em>absolute URL</em>, <em>URL parser</em>, and <em>URL equivalence</em> are defined by [[!url-1-20141209]] <small><em>(or the editor's draft)</em></small>.</dd>
+          <dd>The terms <a>URL</a>, <a>absolute URL</a>, <a>URL parser</a>, and <a>URL equivalence</a> are defined by [[!url-1-20141209]] <small><em>(or the editor's draft)</em></small>.</dd>
         </dl>
```
adrianhopebailie, line 309: Should these `<a>` tags not be `<dfn>` tags?

lordzen: > Should these <a> tags not be <dfn> tags?

Not if they are terms define in other specs, right? As such, I think Respec automatically resolves the `\<a>`-marked references to generate hyperlinks that lead to the external definitions in the correct specs.

sideshowbarker, line 309: > Should these `<a>` tags not be `<dfn>` tags?

Not if they are terms define in other specs, right? As such, I think Respec automatically resolves the `<a>`-marked references to generate hyperlinks that lead to the external definitions in the correct specs.

halindrome, line 309: Actually no - respec does not resolve references to definitions in other documents.  We have some plug-ins that can do that if it is something we are interested in.  Normally the way people handle this is to have a common file of terms that is included in each document dynamically.  There is a plugin for this as well.  It removes terms that are unused from each document to reduce clutter.

adrianba, line 309: The issue here is that we have two different options that both should be `<dfn>`. You can't define the same term twice so I made the second instances anchors until we pick an option. We should keep these as `<a>` until we decide (otherwise there are errors).

adrianhopebailie, line 309: @adrianba - I think the markup you intended to put in your response was lost.

adrianba, line 309: Oops. Updated my comment.

adrianba:
In general, this LGTM (aside from the `<dfn>` vs `<a>` comments above).

rvm4:
abrianba, can you elaborate, the dfn/a issue is no longer present in the PR...

adrianba:
I think you updated the PR after my last comment.

Merged as 1c303e0b0266b44b5bc61ac9b3b781c2bd4b2157.

PR #125, opened 2016-04-07
Reference names should have hyphens

### LINKED ISSUE #63: [architecture] Reference names should have hyphens
The reference names used (e.g., PAYMENTREQUESTAPI) may be hard for non-native English speakers to parse.  It would be better to use hyphens; PAYMENT-REQUEST-API, BASIC-CARD-PAYMENTS, METHOD-IDENTIFIERS.  Also, for future discoverability, it might make sense to prefix all of these with something.  Not critical.

adrianba:
Fixes #63.

### DIFF
--- a/specs/architecture.html
+++ b/specs/architecture.html
@@ -48,7 +48,7 @@
           }],
 
           localBiblio:  {
-              "PAYMENTREQUESTAPI": {
+              "PAYMENT-REQUEST-API": {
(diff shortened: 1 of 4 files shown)


adrianhopebailie:
LGTM

zkoch:
Merged as a9669e568fa56332241aaed94711dd0a53704e19

PR #131, opened 2016-04-08
Flows issue 32

mattsaxon:
The flow diagram is referenced from a remote rep, but I will address that with the next PR if we agree this is a good way to proceed.

### DIFF
--- a/specs/basic-card-payment.html
+++ b/specs/basic-card-payment.html
@@ -162,15 +162,6 @@
       </table>
     </section>
 
-	<section id="flow">
-      <h2>Payment Method Flow</h2>
-      <p>The following represent the flow for all the supported <a>payment method identifier</a> strings as they could be used by a website</p>
-      <p>The blue call-outs show where and how the API is invoked.</p>
-
-	  <img width="100%" src="http://www.plantuml.com/plantuml/proxy?fmt=svg&src=https://raw.githubusercontent.com/w3c/webpayments-flows/gh-pages/TargetFlows/Card/MerchantHosted-CardPayment-TargetWithSimpleApp.pml"/>
-	
-    </section>
-	
     <section id="init-data">
       <h2>Payment Method Specific Data for the PaymentRequest constructor</h2>
       <p>This section describes payment method specific data that is supplied as part of the <code>data</code>
--- a/specs/basic-card-payment.html
+++ b/specs/basic-card-payment.html
@@ -142,14 +142,21 @@
 
       <table>
         <tr><th>Identifier String</th><th>Description</th></tr>
-        <tr><td>visa</td><td>Visa (Credit, Debit, Delta or Electron)</td></tr>
-        <tr><td>mastercard</td><td>MasterCard and EuroCard</td></tr>
+        <tr><td>visa</td><td>Visa (Credit, Debit and Electron)</td></tr>
+        <tr><td>visa/credit</td><td>Visa Credit</td></tr>
+        <tr><td>visa/debit</td><td>Visa Debit</td></tr>
+        <tr><td>visa/electron</td><td>Visa Electron</td></tr>
+        <tr><td>mastercard</td><td>MasterCard (and EuroCard)</td></tr>
+        <tr><td>mastercard/credit</td><td>MasterCard Credit</td></tr>
+        <tr><td>mastercard/debit</td><td>MasterCard Debit</td></tr>
         <tr><td>amex</td><td>American Express</td></tr>
         <tr><td>discover</td><td>Discover</td></tr>
         <tr><td>maestro</td><td>Maestro</td></tr>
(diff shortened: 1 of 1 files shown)


AxelNennker:
Thanks for the sequence diagram. That is really helpful.

Where is the "Payment Mediator" defined?
When looking at the [PaymentRequestUpdateEvent](https://cdn.rawgit.com/w3c/browser-payment-api/0d1d5d7ff0f1bb7b37970994f1eb719101aaccbc/fpwd/paymentrequest.html#paymentrequestupdateevent) I was wondering whether the Payment App gets the updated Payment Details. From the sequence diagram it seems that the Payment App never gets the updated Payment Details like Shipping Address. Why is this?

ianbjacobs:
@AxelNennker 

Mediators are defined int he architecture document:
 https://w3c.github.io/browser-payment-api/specs/architecture.html#payment-mediators

See also:
https://github.com/w3c/browser-payment-api/issues/75

Ian

AxelNennker:
@ianbjacobs Are there plans to add sequence diagrams to the Payments API spec or only to the Basic Card Payment document? None of the documents refers to Payment Mediator - which is strange.

ianbjacobs:
@AxelNennker,

Thank you for the suggestions. We will continue to update the documents.

In fact, Matt Saxon has an open pull request to include a diagram in the Card Payment Document, so stay tuned:
 https://github.com/w3c/browser-payment-api/pull/131/commits/89591e61c7dd6eb20b6a71d328642143e3fc6b21

Regarding usage of Payment Mediator: That will be described in detail in the upcoming specification on payment app registration and communication:
 https://lists.w3.org/Archives/Public/public-payments-wg/2016Apr/0016.html

In short: we are catching up to you! :)

Ian

mattsaxon:
> When looking at the PaymentRequestUpdateEvent I was wondering whether the Payment App gets the updated Payment Details. From the sequence diagram it seems that the Payment App never gets the updated Payment Details like Shipping Address. Why is this?

The sequence diagram doesn't model the event you mention at the moment, I will add this when I get opportunity.

In terms of the shipping address specifically, at the moment, the interface between the Payment Mediator and the Payment App is not clearly defined as far as I am concerned. I believe this to be an emerging area of the specification. I modelled it this way as I believe it is the current intentions of the Editors of the specification, but I think it may be worthy of further discussion.

mattsaxon:
@AxelNennker 

> Are there plans to add sequence diagrams to the Payments API spec or only to the Basic Card Payment document?

Yes I plan to submit a PR for the Payments API spec also and any others that would benefit from them.

adrianhopebailie:
@AxelNennker said:

> When looking at the PaymentRequestUpdateEvent I was wondering whether the Payment App gets the updated Payment Details. From the sequence diagram it seems that the Payment App never gets the updated Payment Details like Shipping Address. Why is this?

You are correct. There is a distinction between the interaction between the website <-> browser and the website <-> payment app (where the browser is a proxy/mediator).

The collection of shipping data is entirely optional and would only be done by the website if required. This data is collected by the browser and provided back to the website in real0time so that the website can update the details of the payment request if necessary before it is passed to the payment app.

If the merchant wishes to pass the shipping info to the payment app they can do so by including it in the request details but this is only likely if the payment app requires it.

AxelNennker:
@adrianhopebailie Instead of "This data is collected by the browser" you mean "This data is collected by the Payment Mediator", right?

adrianhopebailie:
@axelnennker

Actually no. The browser API is specific to user agents and has some functions that are not a requirement of all mediators.

These include the capturing of shipping data.

adrianhopebailie:
I'd suggest we put this on hold until we have resolved #132 and decided on the best way to express payment method data in these specs.

mattsaxon:
-1, I don't see how the inclusion of the flow in the spec is related to the choice to use webidl or not.

adrianhopebailie:
@mattsaxon - I think this PR (perhaps mistakenly) includes changes that are also supposed to be part of PR #114 which is why I made the comment. i.e. It includes more than just the flow image.

mattsaxon:
There are two PRs out there, but both should be accepted and whilst one does change webidl, I don't believe that leaving it pending is the right approach. The concept is what the PR should be accepted on (or not) . If we change from webidl we would change one that were decided.

adrianhopebailie:
@mattsaxon, my point is that this PR contains all the changes that are also in PR #144.
i.e. If the editors merge this change then they can close PR #144.

I'm happy to merge the change and update from WebIDL to another schema/IDL later was just checking if this PR contained commits that were unintended.

+1 to merge and also close #114.

adrianba:
Cherry-picked d1e70f97baeb8741cef7f3c0e87e155cf70af453 after discussion with @mattsaxon. Added link to image since it isn't all visible by default.

Merged as 8df40d52c78c683d94f324658faab17666b55fa7.

Issue #132, opened 2016-04-10
Should payment method specifications contain WebIDL?


adrianhopebailie:
I wonder if we need to use WebIDL in the payment method specifications.

This data is supposed to be passed through the user agent with no processing so I think we could simply use a JSON schema definition of some form or just examples and explanations?

WebIDL is over-kill in my opinion and may be a barrier to payments domain experts defining these specifications (which is what we desire).

sideshowbarker:
> I wonder if we need to use WebIDL in the payment method specifications.

I think we do not. When I first saw it I wondered why a choice has been made to use WebIDL for this.

> This data is supposed to be passed through the user agent with no processing so I think we could simply use a JSON schema definition of some form or just examples and explanations?

I personally have no strong preference as far as how to express it without WebIDL. But as you point out, given that the UAs need to implement no parsing or handling of the parsed data, I can’t see any value to expressing it using WebIDL.

> WebIDL is over-kill in my opinion and may be a barrier to payments domain experts defining these specifications (which is what we desire).

Yes. WebIDL is a necessary evil in cases where it can’t be avoided. In any cases where it can be avoided, it should be avoided. At least as far as web authors-developers go, WebIDL is an expression language completely foreign/unfamiliar to the experiences of anyone except browser-engine developers.

adrianhopebailie:
The [Web App Manifest](https://w3c.github.io/manifest/) specification describes pure JSON objects and their members.

Would anyone oppose using a similar format?

dezell:
> WebIDL is over-kill in my opinion and may be a barrier to payments domain experts defining these specifications (which is what we desire).

I probably have to take some of the blame the WebIDL shows up in various places.  The reason for that has to do with the following:
1) the W3C NFC specification (written in IDL) and the fact that there was - at least early on - some hope of a more fully featured NFC edition that would include payment.  
2) the general idea that software stacks that have components defined in IDLs (not necessarily WebIDL) tend to be stronger than those that are simply ad-hoc.
3) an IDL gives some idea - in a more formal way, of how responses align with requests, and some idea of how various features like "Promises" fit.

> WebIDL is an expression language completely foreign/unfamiliar to the experiences of anyone except browser-engine developers.

That WebIDL was unsuitable in this area was not clear earlier.  Michael is an expert, so no argument here.
[...]

PR #133, opened 2016-04-10
Merged PaymentRequest params and tweaked to address some issues

adrianhopebailie:
Merged PaymentRequest constructor params to allow for options (only
processed by the UA) and details (passed to the payment app).
This commit proposes solutions to:
#3, #4, #15 and #18


specs/paymentrequest.html
```diff
@@ -284,16 +284,77 @@
...
         <div class="note">
-          <p>The <code>supportedMethods</code> sequence contains the <a>payment method identifiers</a>
-          for the <a>payment methods</a> that the merchant web site accepts.</p>
+
+          <p>The <code>details</code> contains the core payment request data that will be passed on to the user selected
+            <a>payment app</a>.</p>
+
+          <p>The <code>amount</code> sequence contains the amount (possibly in different currencies) that the website
+            is requesting to be paid.</p>
+
```
adrianhopebailie, line 293: Addresses #3

adrianhopebailie:
Also resolves #48

specs/paymentrequest.html
```diff
@@ -646,64 +709,64 @@ <h2 id="state-transitions" class="informative">State transitions</h2>
...
-      <h2>PaymentDetails dictionary</h2>
+      <h2>PaymentMethod</h2>
       <pre class="idl">
-dictionary PaymentDetails {
-  sequence&lt;PaymentItem&gt; items;
-  sequence&lt;ShippingOption&gt; shippingOptions;
+dictionary PaymentMethod {
+  required sequence&lt;DOMString&gt; identifiers;
+  object data;
 };
```
adrianhopebailie, line 718: Could add a `sequence<CurrencyAmount> amount` to solve issue #4

adrianhopebailie:
The separation of concerns that is introduced by this change also makes solving #1 more elegant.

adrianhopebailie:
I believe that merging this would also solve for #49 and make PR #83 redundant

specs/paymentrequest.html
msporny, line 718: I'm very wary of catch-all containers such as "data". I'd rather we figure out how we do extensibility properly before punting to "generic data buckets". So, -1 to this line. The rest of the PR is looking good - still reviewing it.

specs/paymentrequest.html
```diff
@@ -224,7 +225,7 @@
     <section>
       <h2>PaymentRequest interface</h2>
       <pre class="idl">
-        [Constructor(sequence&lt;DOMString&gt; supportedMethods, PaymentDetails details, optional PaymentOptions options, optional object data)]
+        [Constructor(PaymentDetails details, optional PaymentOptions options)]
```
msporny, line 228: Yes, this is a very good step in the right direction.

specs/paymentrequest.html
```diff
@@ -284,16 +284,77 @@
           payment app.
         </div>
         <div class="note">
-          <p>The <code>supportedMethods</code> sequence contains the <a>payment method identifiers</a>
-          for the <a>payment methods</a> that the merchant web site accepts.</p>
+
+          <p>The <code>details</code> contains the core payment request data that will be passed on to the user selected
+            <a>payment app</a>.</p>
+
+          <p>The <code>amount</code> sequence contains the amount (possibly in different currencies) that the website
```
msporny, line 291: Is this "amount" in the top-level details object, or is it in every paymentMethod object? I'm a -1 for the former (top-level) and a +1 for the latter (in every payment method object). The -1 is because it feels like doing so would duplicate data (because we'd also have that same data in each payment method object?)... sorry, can't see the complete spec to see where the change is (and don't have the time to checkout this branch and view it locally). Hopefully this comment makes sense to you @adrianhopebailie.

specs/paymentrequest.html
```diff
@@ -284,16 +284,77 @@
...
+            is requesting to be paid.</p>
+
+          <p>The <code>supportedMethods</code> sequence contains the <a>payment methods</a> that the merchant web
+            site accepts. Each <a>payment method</a> is linked to one or more <a>payment method identifiers</a>.</p>
+
+          <p>The <code>identifers</code> sequence contains the <a>payment method identifiers</a> that identify the
+          <a>payment methods</a> the web site supports and for which the associated <code>data</code> is part of the
```
msporny, line 298: -1 for generic "data" container elements.

specs/paymentrequest.html
```diff
@@ -284,16 +284,77 @@
...
+
+          <p><code>data</code> is a <a>JSON-serializable object</a> that provides optional information that might
+          be needed by the supported <a>payment methods</a>.</p>
+
           <pre class="example highlight">
-            ["visa", "bitcoin", "bobpay.com"]
+            {
+              "amount" : [{
+                  "currency" : "USD",
+                  "value" : "55.00"
```
msporny, line 308: Looks like a syntax error here? This is what's causing me to be confused about whether "amount" applies to all "supportedMethods" or some of them, etc.

specs/paymentrequest.html
msporny, line 291: To be clear, I'm a +1 to merge either way as this is better than what's in the current spec, imho.

adrianhopebailie, line 291: There is a top-level amount but as I say in the issue marker we could also define an optional per method amounts field which (if present) overrides the top-level amount. See https://github.com/w3c/browser-payment-api/issues/4#issuecomment-207913477

specs/paymentrequest.html
adrianhopebailie, line 298: We need a mechanism to pass payment method specific data in the payment request. The format of this data is not specified here so this is just a hook. I don't think it should be defined as WebIDL at all (See https://github.com/w3c/browser-payment-api/issues/132). What do you suggest?

specs/paymentrequest.html
adrianhopebailie, line 308: Fixed

specs/paymentrequest.html
msporny, line 298: > I don't think it should be defined as WebIDL at all (See #132).

+1 for NOT using WebIDL. It's overkill for what we're trying to do and not very helpful for non-browser environments.

> What do you suggest?

I'm working on a VERY ROUGH proposal for a payment message data model and syntax-specific validation mechanism here:

http://msporny.github.io/webpayments/proposals/core-messages/#the-data-model

That might be ready to talk about in a day or two...

Instead of "data", just allow data to be placed into each "supportedMethod" w/o an encapsulating container. The downside there is the potential for a developer to use a common term like "account" in the wrong way (term conflicts). The upside being that we get rid of all of those "data" containers.

I'm not convinced that the risk of term conflicts is high enough for us to worry about it. Keep in mind that developers will know what goes in the payment method specific messages because there will be specs with well defined schemas (with limits expressed in English and perhaps even JSON Schema). If they use JSON-LD, this is a non-issue as you're almost guaranteed not to conflict.

In any case, I hope to have a more solid proposal/alternative to "data" in the next day or two. Again, please merge the PR even w/ "data" in there as well as the "amount" concerns as I like this approach more than what's in the current spec.

We can make a separate PR to take "data" out if that's what the group wants to do later on.

specs/paymentrequest.html
msporny, line 291: Hmm, I see. I'd prefer the ability to override amount on a per method basis, but am concerned about the complexity for the developer (they will need to know that amounts in a payment method override the top-level amount - just one more thing for them to know/understand). 

The other alternative is to only have an amount per supported payment method set (nothing in the top-level object); the downside here being that the same amount might be repeated. The downside, though, might not be that terrible as maybe most prices would in fact be different per payment method set?

The primary problem is in duplication of content with a secondary concern around cognitive load placed on the developer. Do we get more duplication with a top-level "amount" w/ overrides or a per supportedMethod amount w/ no top-level "amount"? My gut is telling me we get less duplication w/ the latter, but I admit to not having put a tremendous amount of thought into it.

specs/paymentrequest.html
adrianhopebailie, line 298: > We can make a separate PR to take "data" out if that's what the group wants to do later on.

Rather than hold up this PR I'd request that we do this.

specs/paymentrequest.html
adrianhopebailie, line 291: I'm inclined to leave it as a top element (initially I think it will be seldom overridden) and make the per method member optional (which I think makes it clearer that it's an override).

dlongley:
:+1: To this PR, but note that I'm in agreement with Manu's comments.

specs/paymentrequest.html
```diff
@@ -284,16 +284,78 @@
...
+          <a>payment methods</a> the web site supports and for which the associated <code>data</code> is part of the
+          request.</p>
+
+          <p><code>data</code> is a <a>JSON-serializable object</a> that provides optional information that might
+          be needed by the supported <a>payment methods</a>.</p>
+
           <pre class="example highlight">
-            ["visa", "bitcoin", "bobpay.com"]
+            {
+              "amount" : [{
```
zkoch, line 306: I am not a fan of providing multi currency support, especially from a UX perspective. If a merchant really wants this, they could always ask their users in what currency they want to pay before calling paymentrequest.

specs/paymentrequest.html
```diff
@@ -284,16 +284,78 @@
...
+                  "value" : "55.00"
+              }],
+              "supportedMethods" : [
+                {
+                  "identifers" : ["visa", "mastercard"],
+                  "data" : {}
+                },
+                {
+                  "identifers" : ["bobpay.com"],
+                  "data" : {
```
zkoch, line 317: 'data' seems too generic. Perhaps... identifierData?

Also, do we imagine being able to support identifier-specific pricing in here?

specs/paymentrequest.html
adrianhopebailie, line 306: @zkoch We have had strong support for including this.

I understand that it makes the UX more complex but that's an implementer challenge that I assumed would be embraced as a way to differentiate.

Off the top of my head I can imagine a browser having a small selector that allows the user to select their payment currency and when they do the rest of the UI adjusts to display the pricing in that currency. The selector would only be visible if the request actually contains more than one currency.

On the other hand, I don't see us having any conformance language that forces the implementer to display all currencies to the user in the app selection UI. If the browser picks one currency and displays that currency and price that is still valid as the user's option to select another currency will then be made in the app.

specs/paymentrequest.html
adrianhopebailie, line 317: > 'data' seems too generic. Perhaps... identifierData?

Happy to bikeshed this, it's a very minor detail. @msporny also has reservations about the specific member name. As this PR has some impactful changes it is hard to merge anything else while this is still pending. I recommend that if there are no objections to the PR we merge and then deal with these minor issues in new PRs.

> Also, do we imagine being able to support identifier-specific pricing in here?

Yes, I have an issue note to that effect but didn't want to try and do too much in one PR. This is already a necessarily wide ranging change so anything that could be left to another PR is still pending.

adrianba:
I'm not in favour of this change. The motivation for the four arguments in the current API was as follows:
- `supportedMethods` - this sequence changes infrequently (and might even be hard-coded into a site). It's possible that a payment service provider might provide this data to a site. This data depends on the commercial agreements in place and what the site can handle. It is possible that payment responses passed to a service provider might not require any other code changes to the site.
- `details` - this is the data that varies per transactions. Most commonly this data reflects the contents of a shopping cart or basket. The site will construct this data before making a payment request.
- `options` - this is data that comes from the site's business needs. Do they need a shipping address? Do they want to request an email address? Etc. This data changes very infrequently and most likely requires a code change by the site to address.
- `data` - optional additional data that a payment method might need but that is entirely optional.

As I've mentioned before, the final `data` argument is the most experimental. I don't think we have much evidence yet as to whether this data is almost always required, rarely required, or sometimes required. The goal here was to keep the API simple in the cases where this data is not needed. If we have evidence to show that it is almost always needed then changing the shape here might be useful.

In general in the web platform, we have APIs that consume domain specific data rather than one large undefined blob of data. I think it is better to keep the arguments separate and consider who will change them and when.

adrianhopebailie:
@adrianba I understand the motivations and agree they are valid but I don't believe they justify the inflexibility that results from this format.

Consider the following simple example. A new input parameter is defined for basic card payments. In general merchants will use the same value for all card payments. Today they would need to list out all of the card payment method identifiers they support in the `supportedMethods` array and then again list each individually in the `data` object with the same value provided under each.

So you have this:

``` javascript
var request = new PaymentRequest(
  ["visa", "mastercard", "amex"], 
  details, 
  options, 
  {
    "visa" : {
      "customDataThing" : true,
    },
    "mastercard" : {
      "customDataThing" : true,
    },
    "amex" : {
      "customDataThing" : true,
    },
  }
);
```

Instead of this:

``` javascript
var request = new PaymentRequest(
  options, 
  [
    "identifiers" : ["visa", "mastercard", "amex"],
    "data" : {
      "customDataThing" : true,
    }
  ]
);
```

If you multiply that out for a merchant that supports more than 3 card types and you have a horrible mess.

While I see the value in isolating data like `supportedMethods` and `options` that seldom change I would argue that a website can just as easily keep a static ("hard coded") template that contains all of this data and they will simply merge in the transaction specific data for each request. This may be as simple as adding a total and some line items.

What this proposal also addresses is the idea of "messaging" and the need to define a single JSON-serialized payment request message that is 1) not clouded with data that is only useful to the browser (like shipping options or line items) and 2) can be passed over process or network boundaries to a payment app.

I assert that it is a requirement of the API that the website has complete control over the content of the payment request that is ultimately received by the payment app. i.e.This should not be modified in any way by the user agent. This suggests that this payment request message should be passed as a single object to the API and that any other data that is only required by the user agent should be put into one or more other parameters.

halindrome:
@adrianhopebailie said:

> I assert that it is a requirement of the API that the website has complete control over the content of the payment request that is ultimately received by the payment app. i.e.This should not be modified in any way by the user agent. This suggests that this payment request message should be passed as a single object to the API and that any other data that is only required by the user agent should be put into one or more other parameters.

+1 to this.  This data really needs to be compartmentalized in something that can be a "message".  Personally I would like to see the message signed in some way so that the payment app knows it was not altered on the way, but I know that is unlikely so I am just going to sit on that one until I get to say "told ya' so!"

adrianba:
As discussed on the [telcon](http://www.w3.org/2016/04/28-wpwg-minutes.html#item03), closing this issue. Some parts of this proposal were addressed in other PRs (#158, #162). We should address multiple currency (#3) and amounts per payment method (#4) in separate proposals.

Issue #75, opened 2016-03-16
[api] Section 11 PaymentResponse seems to be missing shippingAddress


halindrome:
If the web site requests sets requestShipping to true, shouldn't the user's shipping address be returned as part of the PaymentResponse?

rsolomakhin:
It's a part of the of the PaymentRequest object, I believe.

``` JavaScript
var request = new PaymentRequest(methods, details, {'requestShipping': true});
request.addEventListener('onShippingAddressChange', function(event) {
  console.log(request.shippingAddress);
});
request.show().then(handleResponse).catch(handleError);
```

zkoch:
That's true, @rsolomakhin, but it still seems like it should be part of the PaymentResponse. PaymentResponse should contain the entirety of information you requested.

halindrome:
I concur @zkoch.  It feels weird that we would be modifying the input object.  Actually I think the PaymentRequest object should NOT expose the shippingaddress.  That feels like a way less reuptable merchants could somehow mine customer address info even when a transaction did not complete (because event handlers get fired when address data is updated and thus could be harvested from the input object).

dlongley:
The reason why the address is available in the event handler is because merchants may want to specify different shipping options based on the chosen address. But, yes, this could be used to mine customer addresses.

dlongley:
I also wonder if users would be surprised to find out they had given their address info to the merchant site even if they didn't authorize payment.

halindrome:
Yes.  And I get that.  It just seems like a weakness.  Note also that a merchant may want to offer different pricing based upon shipping address.  Wine.com, for example, charges different rates depending on the state you live in.

dlongley:
@halindrome,

Yes, both are arguments for collecting shipping address earlier in the shopping process.

rvm4:
Hmm, this is tricky. We cannot (in general) determine final pricing without knowing where the product is going which necessitates passing addresses to the client before authorization occurs, but we cannot get to the authorization screen without determining final pricing.

If I suppose that payment mediators are likely to implement this by presenting address forms prior to payment methods/apps, then I would argue that this is no less secure that standard practices today and consumers are used to the risk of providing an address. 

For privacy sake though, I would suggest we only pass in vague address information until payment has been authorized. For instance, giving the client country, state, and zip code is generally enough for any shipping purposes and does not leak very sensitive information (these are things you may already be able to deduce).

adrianhopebailie:
@zkoch said:

[...]

PR #135, opened 2016-04-10
provide partial shipping address in PaymentRequest and full shipping …

rvm4:
Addressing issue #75 by introducing PartialShippingAddress and moving ShippingAddress to PaymentResponse.


rvm4:
oh geez, looks like my branch got all messed up :( I'll have to try again...

PR #136, opened 2016-04-10
provide partial shipping address in PaymentRequest and full shipping …

rvm4:
Addressing issue #75 by introducing PartialShippingAddress and moving ShippingAddress to PaymentResponse.


kirkalx:
Perhaps we could find a better name than PartialShippingAddress to suggest the nature of the part involved?

rvm4:
I cannot figure out how to update this pull request because my git is all wonky, starting over...

Issue #103, opened 2016-03-29
What component does the payment method intersection?


mattsaxon:
The architecture document reads...

> The user agent determines the intersection...

I believe this should read "The Payment Mediator...."

In addition we should be clear that the payment mediator may optimise the experience, applying knowledge such as user preferences (e.g. preferred payment method) and environmental factors (location).

Note, I will submit a PR post FPWD

ianbjacobs:
+1 to using payment mediator consistently as the primary component.

adrianhopebailie:
The browser API spec refers to user agents (not mediators) which is correct as this is a specification for user agents. #137 incorrectly suggests changes to the browser API spec. 

The browser API spec MAY refer to the role of payment mediator as a role that the user agent assumes in implementing the API but we should emphasize that some of the functions defined in the spec are not functions of a payment mediator they are functions that are only defined for user agents.

The architecture document should describe the role of a payment mediator which the browser API spec could reference

ianbjacobs:
While I believe the mediator is the primary entity that will match identifiers, I propose that we define the matching in the PMI spec. The complexity of the matching algorithm will depend on the semantics we build in (e.g., grouping and subclassing will increase the complexity).

adrianhopebailie:
Matching algo will be defined in the PMI spec

PR #137, opened 2016-04-10
codemod user agent to payment mediator in payment request

rvm4:
addresses issue #103

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -73,7 +73,7 @@
       <p>
         This specification describes a web API to allow merchants (i.e. web sites selling
         physical or digital goods) to easily accept payments from different payment methods with
-        minimal integration. User agents (e.g. browsers) will facilitate the payment flow between
+        minimal integration. Payment Mediators (e.g. browsers) will facilitate the payment flow between
         merchant and user.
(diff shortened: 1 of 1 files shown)


adrianba:
Note that #103 refers to the architecture document. Also, I'm not sure a global replacement of user agent into payment mediator is what is needed here. For example, it definitely is a user agent that implements the IDL interfaces.

ianbjacobs:
@adrianba,

> Also, I'm not sure a global replacement of user agent into payment mediator is what is needed here.

That sounds likely. The intent (in my mind) is to clarify roles (so "global replace" may not be the right way to do that.)

Ian

adrianhopebailie:
:+1: to @ianbjacobs and @adrianba 

I think we need to clearly separate the concerns. The user agent is performing two roles, as payment mediator and also as data collection mediator (facilitating the optimized checkout flow).

I have proposed splitting the API document into two to acheive this in #138

adrianhopebailie:
@adrianba said (and @ianbjacobs concurred): 

> I'm not sure a global replacement of user agent into payment mediator is what is needed here

I agree. This document is probably the wrong place to make these changes.

PR #139, opened 2016-04-18
provide shipping address in PaymentResponse

rvm4:


rvm4:
addresses issue #75

adrianba:
LGTM.

adrianhopebailie:
I am opposed to this change as it mixes up the data provided by the user agent with the data provided by the payment app.

The website is able to get the selected shipping address by listening for the relevant events or from the PaymentRequest object.

zkoch:
This change LGTM. I think the response should include all information that was requested by the merchant. This will also eventually hold true for email and phone number if the group decides to include those.

halindrome:
@adrianhopebailie I appreciate your concern.  Question: architecturally, is there a reason the _payment app_ could not supply shipping address etc?  My current payment app does this.  It's pretty handy.  I appreciate that some user agents have the built-in ability to do this as well, but are we precluding the possibility that a payment app could provide this service? 

I can answer my own question: Yes, we are precluding that.  But should we be?  If a payment app told the mediator, as part of its registration, that it had the ability to address merchant requests for address, phone, email, etc, then those requests could be passed along.  Right?

adrianhopebailie:
@halindrome said:

> we are precluding [collecting data from the payment app]. But should we be?

I agree we should absolutely not preclude this BUT I think that mixing the response from the user agent into the response from the payment app makes this worse not better.

How does the website know what the source of that data was? If, in future, the payment app is able to provide this data where does it put it and how does the website know it was from the app and not the user agent.

I am opposed to this change because we are conflating the role of the payment mediator (which returns responses from payment apps) with the role we have given the user agent (which gathers user data to make the checkout process faster).

adrianhopebailie:
@zkoch said:

> I think the response should include all information that was requested by the merchant. This will also eventually hold true for email and phone number if the group decides to include those.

If that is the requirement then we should do this in a way that makes it very clear which data in the response is from the payment app and which data is from the user agent. The current proposal does not do this.

adrianba:
> If that is the requirement then we should do this in a way that makes it very clear which data in the response is from the payment app and which data is from the user agent.

I don't necessarily believe this is true but this is what `details` is intended to contain so perhaps an amendment to the following text might alleviate your concern?

> `details`
> 
> A JSON-serializable object that provides a payment method specific message used by the merchant to process the transaction and determine successful fund transfer.

burdges:
I think the user agent should never provide shipping address without explicit user interaction, like the user being provided with defaults in a web form.  It potentially violates security properties of payment apps meant to protect user privacy.

adrianba:
I merged this as ea1cbc3d58b729ee8fd8a1d8911847d6ad4c027f. I believe all the issues discussed here related to issue #75 are addressed in this commit. @burdges comment is addressed in #170. Discussions about payment apps are best discussed as part of the other issues about interfacing to apps.

Fixes #75.

PR #140, opened 2016-04-18
Proposed text for issue 23.

### LINKED ISSUE #23: Can the merchant influence order of presentation of payment apps to the user
This issue was [discussed at the F2F](https://www.w3.org/2016/02/24-wpwg-minutes#item06).

The discussion suggested that we should support merchants specifying a preference and allow users to express a preference that overrides merchant preferences.

adrianba:
Fixes #23.


adrianhopebailie:
+1 - LGTM

PR #160, opened 2016-04-27
Remove id attribute from PaymentItem

adrianba:
PaymentItems are only used for display and we don't currently have a
reason to provide an id. Since this is a dictionary, developers can add
whatever fields they want but _requiring_ an ID is unnecessary.


rvm4:
Sounds good to me

adrianhopebailie:
LGTM

PR #162, opened 2016-04-27
Refactor the supportedMethods and payment specific data

adrianba:
This change updates the first argument to the PaymentRequest constructor
to combine the supportedMethods and payment method specific data. By
making this change it is possible to share an instance of data amongst
multiple payment methods that wasn't possible before.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -226,7 +226,7 @@
     <section>
       <h2>PaymentRequest interface</h2>
       <pre class="idl">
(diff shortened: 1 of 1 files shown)


msporny:
This is an improvement over what is in the spec now. +1 to merge, even though I disagree w/ generic data containers like "data". I still think the first argument should be a single PaymentRequest message ... something like what's defined in the Core Messages proposal.

All that said, I'd rather have this than what we have in the spec now, so +1 to merge.

adrianba:
Merged as c3a3ff7d2491b921efa4e22820255650e180cc8b per [telcon](http://www.w3.org/2016/04/28-wpwg-minutes.html#item03).

PR #158, opened 2016-04-27
Separate total from line items.

adrianba:
Fixes #18.

PR #166, opened 2016-04-28
Updated examples to reflect changes in recent PRs

### LINKED ISSUE #164: Examples need updating with recent PRs
I didn't update all the examples with recent PRs so as to avoid merge conflicts with separate proposals. Will do that now.

adrianba:
PRs #158, #160, and #161 made changes that affect the examples. These
weren't reflected in original PRs to keep those changes simple and to
avoid merge conflicts from changing the same lines in the spec. This
changes updates the examples.

Also updating the syntax to be consistent and use JavaScript (rather
than JSON).

Fixes #164.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -264,7 +264,7 @@
           // Process paymentResponse
           // paymentResponse.methodName contains the selected payment method
           // paymentResponse.details contains a payment method specific response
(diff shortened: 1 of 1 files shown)


adrianba:
Merged as acf1150b9a2ca068ee7adf6c771510a99d4ead45.

PR #168, opened 2016-04-28
Remove issue #119 note and assert that total should be non-negative

### LINKED ISSUE #119: Support for negative amounts
https://github.com/w3c/browser-payment-api/pull/101 changed the formatting rules for amounts so that the value is not signed (per the resolution of the group at https://github.com/w3c/webpayments/issues/57).

Part of this PR is being reverted (See #120) to reflect the proposal by the @adrianba to use signed numbers as the mechanism for indicating negative amounts and/or refunds.

PR #111 proposes an alternative mechanism for annotating line items such that the flow of funds (debit or credit) is defined as an attribute of the PaymentItem. However, some commentors have suggested that the TramsactionType of a payment request should be defined at a global level for the request and not per line item so this PR needs to be tweaked and further discussed before it can be merged.

The motivation for using such a mechanism (as opposed to signed numbers) is that it conforms to financial industry norms and existing payment messaging standards like ISO20022 and ISO8583.

adrianba:
`CurrentAmount` defines a monetary value. It is currently used in two places in the PaymentRequest API:
- In the total field indicating the total amount of the payment request; and
- In the `displayItems` collection that contain optional line items to be displayed by the user agent.

For the first version of the specification, I propose that we shouldn't try to address the meaning of a negative total. Changing the direction of the payment request and supporting things like refunds should be out of scope (at least for now). This means that we should assert that total must be a non-negative number.

A common use case for the `displayItems` will be to show discounts. These are commonly shown on web pages as negative amounts. For example, here is a recent order summary from a transaction I made:

image

There has been some discussion about whether the API should use "[financial industry norms and  existing payment messaging standards like ISO20022 and ISO8583](https://github.com/w3c/browser-payment-api/issues/119#issue-145613354)." This would presumably lead us to use terms like debit and credit for values. However, we're not designing an API for the financial industry - we're designing it for web developers who are trying to construct the UX shown above. In the PaymentRequest API, negative amounts are only ever used for display. In the checkout case, for example, it is clear what a negative number means here. Web developers will be more confused if we ask them to think in terms of debit and credit.

I believe we should keep the current spec text for negative amounts. It is simple determine if a value is negative (by testing whether the first character is U+002D HYPHEN-MINUS). It is more succinct than requiring an additional field in the dictionary to indicate a negative value. Web developers will readily understand what it is for.

This pull request removes the issue #119 note and asserts that total must be non-negative.

Fixes #119.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -390,8 +390,12 @@
             </div>
           </li>
           <li>
-            If <code>details</code> does not contain a sequence of <code>items</code> with length greater
-            than zero, then <a>throw</a> a <a><code>TypeError</code></a>.
+            If <code>details</code> does not contain a value for <code>total</code>, then throw a
+            <a><code>TypeError</code></a>.
+          </li>
+          <li>
+            If the first character of <code>details.total.amount.value</code> is U+002D HYPHEN-MINUS, then throw a
+            <a><code>TypeError</code></a>. <code>total</code> MUST be a non-negative amount.
           </li>
           <li>
             For each <a><code>PaymentMethodData</code></a> in <code>methodData</code>, if the <code>data</code> field
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -601,13 +601,6 @@ dictionary CurrencyAmount {
   required DOMString value;
 };
       </pre>
-      <div class="issue" data-number="119" title="Should negative amounts be reflected using signed values?">  
-      The resolution of the WG per <a href="https://github.com/w3c/webpayments/issues/57"> Issue #57</a> defined  
-      a format for currencies and amounts that lacked support for negative values. The format below adds this  
-      capability in a way that is not common for financial messaging standards (using signed numbers). The
-      rationale for negative numbers is to support discounts. The group is still discussing whether functionality
-      to support discounts might be implemented in a different manner (e.g., via a transaction type).
-      </div>  
       <p>
         A <a><code>CurrencyAmount</code></a> dictionary is used to supply monetary amounts.
         The following fields MUST be supplied for a <a><code>CurrencyAmount</code></a> to be valid:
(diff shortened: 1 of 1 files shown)


halindrome:
Now that I grok that displayItems (thanks for changing the name!) are actually for "display" and have nothing to do with the actual transaction, I really don't care.  Negative amounts make as much sense as anything in that context.  It is just optional information.

Note: As a merchant, I don't feel like I would ever use this feature because I have too little control over the user experience and I will have already shown the user that information before I call the API.  I would treat the API as "pay me", not "negotiate check out".

adrianba:
> Now that I grok that displayItems (thanks for changing the name!) are actually for "display" 

Your [suggestion](https://github.com/WICG/paymentrequest/issues/68) to separate `total` made this much easier to understand.

adrianba:
Added a commit to throw if `total` is negative. This causes a merge conflict, which I'll resolve if this PR is accepted.

nickjshearer:
> As a merchant, I don't feel like I would ever use this feature because I have too little control over the user experience and I will have already shown the user that information before I call the API. I would treat the API as "pay me", not "negotiate check out".

I think there's a difference between a check out negotiation and displaying a summary of, say, the sub-total alongside sales tax. There's definite value here for merchants and users.

adrianhopebailie:
+1 to merge

I think this is is the best approach to this issue. 

Passing a negative amount to payment systems is very dangerous because traditionally they only deal in unsigned numbers and use transaction types to decide if the settlement effect of the transaction is a debit or credit.

> For the first version of the specification, I propose that we shouldn't try to address the meaning of a negative total. Changing the direction of the payment request and supporting things like refunds should be out of scope (at least for now).

I would assert that it should never be in scope. This API shouldn't require any more knowledge about how a payment is processed than it does already. We have resolved to push the concept of transaction types to the payment method and I hope we stick to that.

I had previously asserted that it seemed like a bad idea to enforce these kinds of business rules (no negative amounts) at the API layer but after a conversation with @lyverovski it was pointed out that if we leave it to the payment method to define what a negative amount means, then merchants that wish to submit a request that can be processed by multiple payment methods are unable to decide if they should use a negative number or not (one payment method may require it and another may use a transaction type defined in the payment method data).

By not allowing negative numbers in the total we avoid this entirely and payment methods are free to assume they will always get a positive number and can define custom data like transaction types to facilitate refunds or other credits.

adrianba:
Merging as 02a6c15fe11ddbc0b5fb7329a3bfe513ca64bb0e.

PR #169, opened 2016-04-28
Fix old reference to items and use total/displayItems

adrianba:


specs/paymentrequest.html
```diff
@@ -390,10 +390,6 @@
             </div>
           </li>
           <li>
-            If <code>details</code> does not contain a sequence of <code>items</code> with length greater
-            than zero, then <a>throw</a> a <a><code>TypeError</code></a>.
-          </li>
-          <li>
             For each <a><code>PaymentMethodData</code></a> in <code>methodData</code>, if the <code>data</code> field
```
adrianhopebailie, line 393: Shouldn't this be replaced with a check for the presence of a `total` value or is that done elsewhere? i.e. 

``` html
<li>
   If <code>details</code> does not contain a value for <code>total</code> and that value is not 
  a <code>PaymentItem</code> then <a>throw</a> a <a><code>TypeError</code</a>.
</li>
```

or similar

adrianhopebailie, line 393: I see you have done that in #168

PR #174, opened 2016-04-29
Adding support for phone and email

adrianba:
This is a proposed solution for issue #1 and a counter proposal to PR
#65.

This change adds support for two new `PaymentOptions` for payer's email
and payer's phone number. These values are provided on the
`PaymentResponse` object. Since these values should not change the total
price, no `PaymentRequestUpdate` event is fired for them.

The change also adds a phone number to the `ShippingAddress` data
structure with the implication that this gets populated when
`requestShipping` is `true`.

Here are some variations to this change that should be considered:
- Do not include `requestPayerPhone` (issue #1 actually only mentions
  shipping phone number)
- Do include a `requestShippingPhone` value and only populate the
  `ShippingAddress.phone` value if both `requestShipping` and
  `requestShippingPhone` are true.

### DIFF
--- a/specs/paymentrequest.html
+++ b/specs/paymentrequest.html
@@ -688,6 +688,8 @@ dictionary PaymentDetails {
       <h2>PaymentOptions dictionary</h2>
       <pre class="idl">
 dictionary PaymentOptions {
+  boolean requestPayerEmail = false;
+  boolean requestPayerPhone = false;
   boolean requestShipping = false;
 };
       </pre>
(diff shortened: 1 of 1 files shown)


adrianba:
One advantage to using boolean types here is that we could change them in the future to be an enum that allowed for required/optional/notRequired values. I'm proposing that we start with boolean to keep it simple and get implementation experience.

ianbjacobs:
@adrianba,

> One advantage to using boolean types here is that we could change them in the future to be an enum that allowed for required/optional/notRequired values.

Is the implication that changing from boolean to some other values is a low cost change from an implementation standpoint?

Ian

adrianba:
@ianbjacobs 

> Is the implication that changing from boolean to some other values is a low cost change from an implementation standpoint?

I'm not implying anything about cost (there are definitely UI challenges for example) but I am implying that it is possible to do this later in a more-or-less backwards compatible way.

zkoch:
This PR LGTM. I'm also okay as a starting point with leaving out "requestPayerPhone" but I do think there are a variety of situations where it would be useful (e.g. hotel bookings, where a shipping address is not requested, but a phone # is).

rvm4:
This sounds fine to me, but I don't understand why you have added phone number to the address, but not email. Neither seems more relevant to the address and I don't think either one has a stronger use case than the one. Can we add both to the address, or create some other Contact structure?

mattsaxon:
Happy to merge. Will update comment with more detailed outcome later.

adrianba:
Merged as 31346cfa8d0588743e21bd183bcb85c523a00f59 after last telcon and further discussion with @mattsaxon.

Issue #134, opened 2016-04-10
BasicCardResponse: cardholderName should be optional.


Torlus:
I see at least two cases where this field cannot be returned:
- In France, it is forbidden to provide such information, due to the CNIL rules about personal data and rights.
- Visa, MasterCard, and others provide Anonymous Prepaid Cards, making this information irrelevant.

mattsaxon:
In my PR #114, I have indeed showed this as optional.

adrianhopebailie:
Addressed in PR #114

mattsaxon:
Adrian, that PR has not been accepted yet and I've in fact closed it awaiting resolution of PR 133 which changes the shape of how PR 114 will be done.

adrianhopebailie:
Sorry, my mistake. I saw that @adrianba had merged #131 which included the same change but now see he picked only the one specific commit.

mattsaxon:
PR #176 submitted to address

adrianba:
Fixed in 9c32ebc5e797c3a83828d86bc397cd534032ec14.

PR #176, opened 2016-04-29
Losen requirements around Credit Card required fields

mattsaxon:
Removed incorrect required fields as per issue 134

### DIFF
--- a/specs/basic-card-payment.html
+++ b/specs/basic-card-payment.html
@@ -191,10 +191,10 @@
       <h2>BasicCardResponse</h2>
       <pre class="idl">
         dictionary BasicCardResponse {
-          required DOMString cardholderName;
+          DOMString cardholderName;
           required DOMString cardNumber;
-          required DOMString expiryMonth;
-          required DOMString expiryYear;
+          DOMString expiryMonth;
+          DOMString expiryYear;
 		  DOMString cardSecurityCode;
 		  
           BillingAddress? billingAddress;


zkoch:
Hey Matt, cardholder name being optional makes sense because of rationale provided in #134, but can you explain expiration dates also being optional?

mattsaxon:
UnionPay debit cards don't have an expiry date.

adrianhopebailie:
@zkoch why would we enforce this at the API level anyway?

To me this IDL simply provides a standardized data model for the payment app to hang data off so that the website (or PSP) that gets the response knows where to find things. If we are too restrictive we will force people to write new payment method specs to simply change a field from mandatory to optional.

A question for @zkoch and @adrianba (or any other browser vendor) is whether you expect the browser to do any validation of this data when it comes back from a third-party payment app? I expected it to be entirely opaque.

halindrome:
Frankly, this is my problem with using WebIDL as a way of representing a data model.  IDL isn't _for_ data modelling (I speak as someone who was on that committee at OMG back in the day).  It was _for_ defining APIs that may be connected asynchronously over a network - especially when the different ends of that connection might be in differing programming languages.  It represents a contract.

In this case the contract is between the payee and the payment app for the important parts of the data.  And we are in no position to enforce nor even negotiate that contract.  At least, I don't think we are.  It is too fluid.

adrianhopebailie:
@halindrome that is why I suggested we don't use WebIDL in these specs in #132

@AxelNennker is opposed to changing from WebIDL so I suggest we put the issue up for discussion on a future call. I'll see what the priorities look like for this week and add it if we have time.

adrianba:
This change LGTM.

PR #181, opened 2016-05-04
Clarified value to use to resove [[acceptPromise]]

adamroach:
I think this is implied by the purpose and signature of the show() method, but it's better to be explicit.


adrianhopebailie:
LGTM

PR #183, opened 2016-05-05
Update to method identifiers spec to use absolute URLs

### LINKED ISSUE #11: What is the format for payment method identifiers for distributed extensibility

adrianba:
Proposed update to the method identifiers spec to reflect the discussion in today's [telcon](http://www.w3.org/2016/05/05-wpwg-minutes.html#item04).

Fixes #11.

### DIFF
--- a/specs/method-identifiers.html
+++ b/specs/method-identifiers.html
@@ -89,15 +89,13 @@
         </p>
     </section>
 
-    <section class="informative">
-      <h2>Requirements for identifiers</h2>
-      <p>There are a set of requirements that the payment method identifiers are designed to support:</p>
-      <ol>
-        <li id="req1">It must be possible for the Working Group to mint a payment method identifier for any payment method.</li>
-        <li id="req2">It must be possible for anyone to mint a payment method identifier for a payment method under their control.</li>
-        <li id="req3">It should be possible to use a standard short string to identify common payment methods.</li>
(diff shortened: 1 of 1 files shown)


ianbjacobs:
+1.

Ian

specs/method-identifiers.html
```diff
@@ -142,294 +115,34 @@
...
-        <ol>
-          <li>MUST be served over HTTPS,</li>
-          <li>SHOULD support HTTP Content Negotiation,</li>
-          <li>SHOULD provide a human-readable document that describes the
-          payment method when <code>text/html</code> is requested, and</li>
-          <li>SHOULD provide a machine-readable document that describes the
-          payment method when <code>application/ld+json</code> is requested.
-          </li>
-        </ol>
-      </section>
```
dlongley, line 280: This section on information for payment method identifiers that do happen to resolve to content shouldn't be completely removed just because we settled on using absolute URLs. This is still an additional option under consideration.

dlongley:
We decided to use absolute URLs, but there were other options that were still under consideration (like describing requirements if payment method identifiers happen to be dereferencable) that were not affected by that decision. I'm +1 for this PR except for the removal of those other options, which should remain.

adrianba:
Merged as d609cd7c0d6888c72fffdc54cc0152ba2b39f445 per telcon.

PR #191, opened 2016-05-11
Explain how to treat unrecognised complete values

### LINKED ISSUE #122: Consider revising the design of complete()
In TAG review @triblondon said:

> paymentRequest.complete feels slightly weird, and I'm wondering why the merchant page does not instead receive an unresolved promise that it can resolve or reject. That would seem more aligned with the use of promises elsewhere.

adrianba:
I don't think we should make any further changes related to #122 for now. It's hard to make a trinary value fit into the promises model in an obvious way.

We should be clear on what to do with unrecognised values to allow us to add new ones in future.

Closes #122.


adrianba:
Merged as 4da43cbc479fd6a598cded52fab37e4886232278 per [telcon](http://www.w3.org/2016/05/12-wpwg-minutes.html#item07).

PR #193, opened 2016-05-12
Rename the address type since it will likely be used for other purposes.

adrianba:
We will most likely want to use the same type in other places. For example, billing address with a card payment.


adrianba:
Merged as 576550c4de9f47feb889f8f7e587b2853e2da668.

Issue #186, opened 2016-05-10
shippingAddress and "c/o" field


adamroach:
Do we need a "c/o" field? In some locations, mail will be delivered only if the known resident's name appears on the address label. Sure, people can hack this by providing it as part of the street address or business name, but since we're defining this greenfield, I think we should accommodate this in a semantically sensible way.

ianbjacobs:
Relates to issue #1.

Ian

PR #197, opened 2016-05-16
Care of

adamroach:
See issue #186


dlongley:
+1 to doing this in a structured way, I'm fine with this PR we can bikeshed later if people dislike the name or something.

davidlehn:
Squash this into one commit before merging.

adrianba:
I believe this is fixed by #198.
Fixes #186.

Issue #189, opened 2016-05-11
ShippingAddress versus Address


adamroach:
The currently defined ShippingAddress interface contains several fields that semantically line up with many of the fields in the Address interface (https://www.w3.org/TR/geolocation-API-v2/#address_interface); however, most such names are unnecessarily different (with the term "region" meaning radically different things between the two interfaces).

This will lead to unnecessary developer confusion.

One approach would be to have ShippingAddress inherit from Address, and then add the other fields that we decide we need. This has the effect of splitting the "streetnumber" field from the "street" field and of adding "county" and "premises" fields, but is otherwise consistent with what is currently defined:

``` webidl
interface ShippingAddress : Address {
    readonly attribute FrozenArray<DOMString> additionalAddressLine;
    readonly attribute DOMString              dependentLocality;
    readonly attribute DOMString              sortingCode;
    readonly attribute DOMString              languageCode;
    readonly attribute DOMString              organization;
    readonly attribute DOMString              recipient;
};
```

Another approach would be to simply rename analogous fields to have identical names:

``` webidl
interface ShippingAddress {
    readonly attribute DOMString              country;       /* was regionCode */
    readonly attribute FrozenArray<DOMString> addressLine;
    readonly attribute DOMString              region;        /* was administrativeArea */
    readonly attribute DOMString              city;          /* was locality */
    readonly attribute DOMString              dependentLocality;
    readonly attribute DOMString              postalCode;
    readonly attribute DOMString              sortingCode;
    readonly attribute DOMString              languageCode;
    readonly attribute DOMString              organization;
    readonly attribute DOMString              recipient;
};
```

Arguably, we could consider "organization" and "premises" to have sufficiently close meanings to consider them the same field, but I'd be curious to hear from others about whether that would serve their purposes.

rsolomakhin:
According to https://www.chromestatus.com/features#geolocation, Chrome implements https://www.w3.org/TR/geolocation-API/, but not https://www.w3.org/TR/geolocation-API-v2/, which includes the `Address` interface. Do other browsers implement it?

tommythorsen:
I like the sound of approach number two -- renaming fields to have identical names. I don't think we should lock ourselves to having to use geolocation's Address interface directly, as some things, like the separation of street and streetNumber may turn out to be inconvenient. Being consistent with the naming of the Address interface, however, sounds like an excellent idea.

[...]

PR #198, opened 2016-05-16
Address alignment

adamroach:
Proposal for issue #189


rvm4:
LGTM. I think the names are largely immaterial, but I do prefer the proposed names as I think they are more commonly used. I suggest we merge this an PR 197.

jnormore:
:+1: makes sense

adrianba:
Merged as 55dfb49470026b9d6b1b3f17d73851e5c26fa212 per telcon.
Fixes #189.

PR #201, opened 2016-05-20
Removing erroneous <svg> clip boundaries

adamroach:
For some reason, Visio appears to put invalid clip boundaries on nested <svg> elements. The webkit/blink implementation of SVG seems to ignore them; however, Firefox's SVG implementation honors them and therefore hides the enclosed elements. See https://bugzilla.mozilla.org/show_bug.cgi?id=1270210 for details.

This change allows the diagram to render properly in conformant SVG implementations.


adrianba:
Merged as be8244c1a556624c93fd2869a052540a71f7abf4.

PR #208, opened 2016-06-01
Rename remaining types so that all begin Payment*

adrianba:
This is normative in that it changes the type system but is almost an editorial change since it doesn't affect the code that most web developers write. The goal is to partition the type system changes in the browser so that new types begin with the string `Payment`.


webpayments-specs:
Seems fine.

Ian

adrianhopebailie:
LGTM

Issue #187, opened 2016-05-10
Update promise rejection


adamroach:
The procedure in section 13.2 does not mention any processing steps in the case the update promise d is rejected. In this case, I think that the only sensible thing to do is to revert the change that the user made which triggered the event. This implies that payment apps must retain the state prior to the change that triggered the event so that it can be rolled back to.

adrianhopebailie:
@adamroach this processing is intended to happen before the user agent has passed the payment request to the payment app so payment apps are not involved in this processing at all.

However, I think this is still a valid issue for the editors to address:

> The procedure in section 13.2 does not mention any processing steps in the case the update promise d is rejected.

Also, the case where d is resolved but is not a valid PaymentDetails dictionary

adrianba:
@adrianhopebailie, I believe this is all handled in the algorithm. It makes a rejected promise a no-op. It also extracts appropriate data from PaymentDetails and ignores other details.

As I said in mail, I disagree that _the only sensible thing to do is to revert the change that the user made which triggered the event_.

adrianhopebailie:
@adrianba are you asserting that there is no need to describe how the UA should behave in the case of a no-op? i.e. The UI stays as is (with current user selections) and the website is unaware that the browser didn't like the new payment details?

adrianba:
The web site is in control of what happens here. The code running is the web site's code. It is the web site that chooses to reject a promise and deal with the consequences. It is the web site that chooses what data to pass. It is the web site's responsibility to pass in data that satisfies the algorithm's checks. The web site has all the opportunity it needs to be aware of what it is doing.

What I said in mail was, rather than adding some complex recovery logic, we should let the web site be in control and get some implementation experience.

adamroach:
Regardless of how we expect this to behave, it needs clear documentation.

I'm a little perplexed by what @adrianba is proposing here, though. Let's imagine that I'm running a website that can only ship to the US and Canada. I get a `PaymentRequestUpdateEvent` with a US address, and update my shipping methods and total appropriately. Then I get a `PaymentRequestUpdateEvent` with an address in Finland. In @adrianba's model, my options appear to be:
1. Ignore the event, and leave the US pricing and shipping methods in front of the user
2. Call `updateWith()` and do something odd like removing all shipping options and zeroing out the cost of the shipping line
3. Call `updateWith()`, and reject the promise, which has exactly the same effect as (1).
[...]

PR #209, opened 2016-06-08
Proposal to abort if updateWith promise is rejected

adrianba:
This is a proposal for how to handle #187. It causes the payment request to be aborted if the promise passed to `updateWith()` is then rejected.

If the promise passed to `updateWith()` is rejected then you end up in an inconsistent state. Either the page hasn't correctly processed the change event (which might mean it didn't correctly update the price, for example) or, if you revert the user's change, the user may end up in a state that they didn't expect. Neither option is good. Instead, after discussion with @zkoch, we're proposing that this aborts the payment request.

### DIFF
--- a/index.html
+++ b/index.html
@@ -400,8 +400,17 @@
         <a><code>TypeError</code></a>. <code>total</code> MUST be a non-negative amount.
       </li>
       <li>
+        If a <a>payment method identifier</a> appears more than once in the <code>methodData</code>
+        or <code>details.modififiers</code> sequences, then <a>throw</a> a <a><code>TypeError</code></a>.
+      </li>
+      <li>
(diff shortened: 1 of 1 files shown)


rsolomakhin:
Looks OK to me.

adrianhopebailie:
LGTM

@adamroach - does this address #187 as you expected? If so I suggest this is merged and #187 closed.

adamroach:
I'm not tickled with this as a resolution, but it does at least put defined behavior around the promise rejection, which was the high-order problem from my perspective. We can close the issue.

adrianba:
Merged as b6af22db68412525589833d13d2d9d206adb1f9b.

PR #210, opened 2016-06-08
Remove special case handling for single shipping option.

adrianba:
We originally added special support for a single shipping option to allow a more streamlined user experience for the cases where you want to get a shipping address but you only have one shipping option. This means the user wouldn't have to select anything.

Since then, we added the `selected` field which allows you to indicate any shipping option should be the selected one. This means the special handling for a single option isn't really necessary any longer. If we remove this then it makes the mental model for developers easier. Developers are responsible for maintaining the selected state for shipping options. If they don't do this then an update to the shipping options causes the current option to be deselected and requires the user reselect.

One of the primary goals here is to ensure that a user never accepts a payment request with details they didn't expect. Another is that we should make it harder for web developer mistakes from misunderstanding the model to result in accepted payment requests with incorrect details. This proposal improves this by removing some of the magic from the API and making web developers be explicit.

### DIFF
--- a/index.html
+++ b/index.html
@@ -715,9 +715,6 @@ dictionary PaymentDetails {
       A sequence containing the different shipping options for the user to choose from.
       <p>If the sequence is empty, then this indicates that the merchant
         cannot ship to the current <a><code>shippingAddress</code></a>.</p>
-      <p>If the sequence only contains one item, then this is the shipping option that
-        will be used and <a><code>shippingOption</code></a> will be set to the <code>id</code>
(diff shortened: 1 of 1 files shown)


rsolomakhin:
+1. I this change makes shipping option handling more straightforward for the web developer.

ianbjacobs:
+!

adrianhopebailie:
LGTM

adrianba:
Merged as 73231dc71e782645afe921115e01589a081cb925.

Issue #4, opened 2016-03-02
Should it be possible to vary amounts depending on payment method


adrianba:
This was [discussed at the F2F](https://www.w3.org/2016/02/24-wpwg-minutes#item07).

The question is whether it should be possible to request different amounts depending on payment method selected and if so how to supply the amounts. During the discussion there was some desire to support this use case but also a desire to keep the case where amounts do not vary simple.

mattsaxon:
This is a very common use case, certainly in the UK; for example it is common for debit cards to attract no additional fees from the merchant, whereas credit cards often do.

Also some payment methods may support specific currencies, e.g. BitCoins, so a different amount is required in those cases. I guess I could conceive of this being true for other more traditional payment methods also that might support specific currencies.

burdges:
I've no answer, but..

I think many merchants have traditionally used this to attract customers to their own Credit Card brand. 

There are definitely European merchants that give the user a choice of being billed in EUR or GBP when paying with a UK bank card, certainly ATMs and several budget airlines.  It's very hard for users to figure out the optimal decision with such systems, certainly I never figured out if it was good or bad. 

Conversely, there are now some legal restrictions on charging different fees based upon the payment method in some jurisdictions.  One could imagine some choices here leading to merchants violating laws, although that landscape appears complicated.

mattsaxon:
Yes, the landscape is complicated,,, I don't think we should be trying to code legislative restrictions into the API, the legislations moves with time and region.

mattsaxon:
As a further complication, the charging that certain merchants apply goes to a lower level than brand. For example a Visa Debit card can attract different changes from the Visa Credit. Currently the Payment Method identifiers don't specify a subbrand, but I've submitted PR #95 to show a reasonable example of how these split down. The PR shows a hierarchy and matching of PMIs could be done at multiple levels, so a merchant may say they accept "Visa" meaning any type of visa or be more specific if they only want to accept sub-brands, e.g. "Electron"

davidillsley:
Just to provide a concrete (though definitely unusual) example, different rates for different cards are laid out in legislation for card payment of taxes in the UK: http://www.legislation.gov.uk/uksi/2016/333/pdfs/uksi_20160333_en.pdf

mattsaxon:
[...]

PR #211, opened 2016-06-08
Support different amounts per payment method.

adrianba:
This is a proposal for issue #4: Should it be possible to vary amounts depending on payment method.

The goal of this change is to make it simple for web developers in the majority case
when price does not vary by payment method. If a developer wishes to have a different
price for specific payment methods then they include a PaymentDetailsModifier that
includes the new total amount and optionally additional display items that are appended
to the main payment details display items for display. This allows the developer to
explain why the price is different due to a discount or surcharge, for example.

### DIFF
--- a/index.html
+++ b/index.html
@@ -400,8 +400,17 @@
         <a><code>TypeError</code></a>. <code>total</code> MUST be a non-negative amount.
       </li>
       <li>
+        If a <a>payment method identifier</a> appears more than once in the <code>methodData</code>
+        or <code>details.modififiers</code> sequences, then <a>throw</a> a <a><code>TypeError</code></a>.
+      </li>
+      <li>
         For each <a><code>PaymentMethodData</code></a> in <code>methodData</code>, if the <code>data</code> field
         is supplied but is not a <a>JSON-serializable object</a>, then <a>throw</a> a <a><code>TypeError</code></a>.
+      </li>
+      <li>
+        For each <a><code>PaymentDetailsModifier</code></a> in <code>details.modifiers</code>, if the <code>total</code> field
+        is supplied and the first character of <code>total.amount.value</code> is U+002D HYPHEN-MINUS, then throw a
+        <a><code>TypeError</code></a>. <code>total</code> MUST be a non-negative amount.
       </li>
       <li>Let <em>request</em> be a new <a><code>PaymentRequest</code></a>.</li>
       <li>
@@ -686,6 +695,7 @@ dictionary PaymentDetails {
   PaymentItem total;
   sequence&lt;PaymentItem&gt; displayItems;
   sequence&lt;PaymentShippingOption&gt; shippingOptions;
+  sequence&lt;PaymentDetailsModifier&gt; modifiers;
 };
       </pre>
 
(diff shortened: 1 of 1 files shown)


rsolomakhin:
I am overall positive about this change because it gives us the ability to specify surcharges and discounts for things like debit cards and credit cards. It does not handle different currency for a payment method, but the currency manipulations can happen inside of the payment app if necessary, IMHO.

ianbjacobs:
Hi @adrianba,

Compared to the proposal that Jason, Andre, Rouslan, and I were working on [1]:
- Both proposals have per-payment-method total (which allows for both price and currency changes per payment method). In addition, your proposal alllows for "local" display items, which makes sense, so +1.
- Our proposal suggests a PaymentMethodsChange event when the user chooses (in the browser) a payment instrument from a given payment method. Your proposal does not include that. Is the expectation that the browser will update the total based on user selection? I see this sentence and want to be sure that it refers to the display of total-after-user-selection: "This field is commonly used to add a discount or surcharge line item indicating the  reason for the different <code>total</code> amount for the selected <a>payment method</a>  that the browser MAY display." Or do we need additional language recommending the user 
  agent SHOULD display the updated information based on user selection?

Finally, your change includes this:

 "If a <a>payment method identifier</a> appears more than once in the <code>methodData</code> or <code>details.modififiers</code> sequences, then <a>throw</a> a <a><code>TypeError</code></a>."

In a separate discussion we are talking about how to manage subclassing of
payment methods (with constraints) and we might propose an approach where
PMIs are repeated but with a different set of constraints. If we do this, and if
we express the constraints within the same URL, then the above language
will be ok. But if we express the constraints in JSON alongside the URL, then
I may request a change to the above to support the additional semantics. 
No change for now, but I wanted to flag this as something we may come back to.

Summary: I am supportive of the change proposal, which I believe addresses
use cases like:
- surcharges and discounts per payment method
- "We accept this amount in USD and this amount in bitcoin."

As we concluded in our proposal, I do not think we should do "multiple currencies for same payment method" as that can be handled separately (see [1]).

Ian

[1] https://github.com/w3c/webpayments/wiki/Support-for-multi-price-and-currency

dlongley:
@ianbjacobs,

-1 to  `PaymentMethodsChange` or any more interactive imperative mechanisms than are absolutely necessary. Declarative mechanisms are better (Rule of Least Power) and allow for more interesting, smarter, innovative systems to be built.

Furthermore, if we're going to start allowing payment instruments to be selected in browser, we will need to consider the ability to select more than one for a particular payment request -- and all the UI and details and complexity that that entails. I think that this should be the domain of Payment Apps, particularly in version 1. Let's get some implementation experience with something simpler before pushing all of the complexity into the user agent.

ianbjacobs:
@dlongley wrote:

"if we're going to start allowing payment instruments to be selected in browser, we will need to consider the ability to select more than one for a particular payment request -- and all the UI and details and complexity that that entails. I think that this should be the domain of Payment Apps, particularly in version 1."

Strong -1 to requirement that browsers or payment apps handle multi-tender payments in v1.

Furthermore, payment apps MAY still do multi-tender payments if they want, and send back the relevant data to the merchant. 

I suggest that if we want to talk more about multi-tender payments, we start a different thread.

Ian

rsolomakhin:
> -1 to PaymentMethodsChange or any more interactive imperative mechanisms than are absolutely necessary. Declarative mechanisms are better (Rule of Least Power) and allow for more interesting, smarter, innovative systems to be built.

I have come around and agree that `PaymentMethodsChange` is not necessary. This pull requests works ~~with~~ without it.

EDIT: `s/works with it/works without it/`

adrianba:
@ianbjacobs, the spec does say that the user agent should update UX based on changes from the update algorithm but I specifically stayed away from describing how an implementation should display the total as payment methods are selected. There are many different ways to do this especially as related to the different ways that payment apps might be implemented. I think it is implicit that a wise implementation would find some way to indicate to the user the different prices per payment method but friendly amendments to make this more explicit welcome.

On the topic of matching payment request identifiers, my position is that we should deal with simple matching in v1, which is what is currently specified. Should a proposal for more complex matching semantics be adopted in the group then any requirements dealing with matching payment methods would need amending. Such a proposal would need to include a comprehensive solution to this issue.

ianbjacobs:
@adrianba wrote:

"Should a proposal for more complex matching semantics be adopted in the group then any requirements dealing with matching payment methods would need amending. Such a proposal would need to include a comprehensive solution to this issue."

We (AHB, MattS, and I) took an action to look into grouping/subclassing based on 26 May discussion. I would like to have a proposal by FTF meeting time. If that proposal does not get traction, I have no problem keeping simple matching semantics in v1.

[1] https://www.w3.org/2016/05/26-wpwg-minutes.html#item04

adrianba:
Merged as 67f4f25f4a0562e7b13c5139e1bb8f8b1d41f53a.

Closes #4.

PR #213, opened 2016-06-08
Remove unnecessary clause about validation.

### LINKED ISSUE #175: Spec prohibits currency validation but doesn't define what it is
In the definition of [`CurrencyAmount`](https://w3c.github.io/browser-payment-api/specs/paymentrequest.html#currencyamount), the spec defines the [`currency`](https://w3c.github.io/browser-payment-api/specs/paymentrequest.html#dfn-currency) field saying:

> user agents MUST not attempt to validate this string.

The spec does not describe what it means to validate the string. It would be better if instead of a negative, the spec could describe positively what MUST happen. Presumably there is something bad that might happen as a result of some kind of validation? The spec is trying to avoid this and make sure that something does happen but it doesn't say what this is.

adrianba:
The spec doesn't define what it means to validate the currency string so it is impossible to know what NOT to do. Additionally, specs should describe what _should_ be done and not what should not be done in order to achieve the desired outcome.

Closes #175.

### DIFF
--- a/index.html
+++ b/index.html
@@ -655,8 +655,7 @@ dictionary PaymentCurrencyAmount {
     <dd>
       <code>currency</code> is a string containing a currency identifier. The most common
       identifiers are three-letter alphabetic codes as defined by [[!ISO4217]] (for example,
-      <code>"USD"</code> for US Dollars) however any string is considered valid and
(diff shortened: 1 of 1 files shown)


ianbjacobs:
+1.

Ian

adrianhopebailie:
LGTM

PR #217, opened 2016-06-15
Add shippingOption to response

adrianba:
When we added shipping address to the response we forgot to include the shipping option. Correcting that here.


rsolomakhin:
+1. This is logical.

PR #220, opened 2016-06-21
Fix two minor typos.

romandev:
There is no 'string' type in WebIDL. Instead we should use DOMString.


yoavweiss:
Marked as non-substantive for IPR from ash-nazg.

Issue #230, opened 2016-08-10
Add Guidance Text for Incognito Mode


adamroach:
This is a recommendation from the Security and Privacy Checklist review. See https://docs.google.com/document/d/1w7ginyzNg-xZUmITK4vzcGUKB4gbMOAvlkWWaRtX14k/edit?usp=sharing for additional detail

When operating in an “incognito” mode, we would expect the Payment Request API to remain available; however, we recommend that any such persistent permission be ignored in such a mode (otherwise, websites with such persistent permission would be able to identify users via their payment details). The user agent would still make stored user information available -- similar to how the web browser assists in filling out form information even when incognito; however, such information would be inaccessible to the merchant web site until submitted by the user. Assistance is expected, automation is not.

When operating in incognito mode, it is probably also advisable to take additional steps, possibly at the expense of usability, to frustrate attempts to determine whether the user has registered payment apps that support specific payment methods. For example, always prompting the user when a payment request is made, even if there are no matching payment apps available, may serve such a purpose. Note, however, that this would need careful consideration, as web sites might determine from such behavior that the user is browsing in an incognito context.

When the Payment Request API is invoked in an incognito context, we suggest that any web-based payment apps also be invoked in an incognito context. This will generally prevent such sites from accessing any previously-stored information; this, in turn, will require users to either log in to the payment app or re-enter payment instrument details.

The Payment Request API specification should thus include discussion on browser behavior in incognito mode.

ianbjacobs:
Ian and AdamR took an action on 19 September to go back to the TAG and let them know we did not understand what to put in the spec since private browsing not a standardized concept.

adrianhopebailie:
A PR, #238, was submitted for this but it was raised by the implementors that there is no standard wrt "private browsing" mode therefor we can't specify behavior in this mode.

@adrianba said:

> browsers should be allowed to compete on what features their private browsing modes provide and how "good" they are - there does not need to be standardization here

The PR has been closed pending feedback from the @w3ctag on what is expected here.

@ianbjacobs and @adamroach are tracking this for the WG

ianbjacobs:
Initiated conversation with TAG:
 https://lists.w3.org/Archives/Public/www-tag/2016Sep/0004.html

Ian

PR #238, opened 2016-08-17
"Incognito mode" notes

ianbjacobs:
Responsive to:
https://github.com/w3c/browser-payment-api/issues/230


rvm4:
This seems pretty reasonable and straight-forward. I cannot see anyone objecting to this language. I'm happy to merge this if no one speaks up to the contrary soon.

index.html
```diff
@@ -1355,12 +1355,18 @@ <h2 id="state-transitions" class="informative">State transitions</h2>
...
     <p>
       The <a>user agent</a> should never share information about the user to the web page
       (such as the shipping address) without user consent.
     </p>
   </section>
+  <section>
+    <h3>Incognito Mode</h3>
+    <ul>
+      <li>This API should remain available in a browser's "incognito mode".</li>
+      <li>Browsers that offer "persistent permission" features for sharing user payment information automatically with merchants should ignore those settings in icognito mode and require explicit user consent.</li>
```
adrianhopebailie, line 1368: icognito -> incognito

adrianba:
I think it's up to implementations to decide how to integrate features into their private browsing modes given the goals they have for those modes. The spec doesn't need to express a preference here. It probably also should not use feature names from particular browsers.

adrianhopebailie:
Closed but reopening #230 to track this

burdges:
We should however take seriously any observations that aspects of the payment spec seem problematic for private browsing mode.

PR #256, opened 2016-09-16
Make UAs throw for invalid monetary amounts

sideshowbarker:


adrianba:
The problem with this change is that it doesn't deal with the updating case where throwing isn't a possibility. Just like in #249 it would have been easy to throw in the constructor in this situation but then the behaviour would be inconsistent in `updateWith()`. We didn't really want the bad update to cause the payment request to be terminated because doing so looks like a user agent bug and not a page bug. Also it will be easier to debug the problem when a user explains what happens.

So for the currency value I'd prefer a different proposal.

adrianba:
Given that we already have the difference I outlined above for negative totals, I've updated the PR to apply the same difference to `updateWith`.

sideshowbarker:
Thanks this looks great—will update the expected results in the corresponding wpt tests to match this.

PR #268, opened 2016-09-17
Add allowpaymentrequest attribute for iframe support

### LINKED ISSUE #2: Should the Payment Request API only be available in a top-level browsing context?
This issue comes from WICG/paymentrequest#30 and was [discussed at the F2F](https://www.w3.org/2016/02/24-wpwg-minutes#item08).

adrianba:
Fixes #2.

### DIFF
--- a/index.html
+++ b/index.html
@@ -152,7 +152,7 @@
       <p>
         A <dfn data-lt="user agents">user agent</dfn> MUST behave as described in this specification
         in order to be considered conformant. In this specification, <a>user agent</a> means a <em>Web
-        browser or other interactive user agent</em> as defined in [[!HTML5]].
+        browser or other interactive user agent</em> as defined in [[!HTML51]].
       </p>
       <p>
         User agents MAY implement algorithms given in this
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1111,6 +1106,38 @@ <h2 id="state-transitions" class="informative">State transitions</h2>
...
 <section>
+  <h2>PaymentRequest and iframes</h2>
+  <p>
+    There are some circumstances where a cross-origin <a>iframe</a> wants to make a payment
+    request. A cross-origin iframe needs explicit permission from the embedding page to invoke
+    the payment request API.
+  </p>
+  <p>
+    The <a>HTMLIFrameElement</a> is extended with an <dfn><code>allowpaymentrequest</code></dfn>
+    content attribue. <a><code>allowpaymentrequest</code></a> is a <a>boolean attribute</a>.
```
adamroach, line 1117: This isn't really what we've discussed. Unless there was subsequent conversation outside of the issue, the proposal (which didn't have any disagreement) was to extend the already-defined "permissions" attribute with a new "payment" value.

halindrome:
@adamroach do you mean the sandbox attribute?  allow-payment would be a sensible additional value for that attribute.

A related question.  Can a script _in_ an iframe query its surrounding iframe element to determine if it has permission to do something?

adamroach:
@halindrome 

>  @adamroach do you mean the sandbox attribute? allow-payment would be a sensible additional value for that attribute.

No, I mean the syntax demonstrated in this comment: https://github.com/w3c/browser-payment-api/issues/2#issuecomment-228463359 (where I also describe why Mozilla's security folks are not okay with using `sandbox` for this kind of thing).

The conversation from there consisted of @zkoch calling the suggestion "great", and then (after consulting with his security team) pointing to the Feature Policy work, which _currently_ defers to [the Permission Delegation API](https://noncombatant.github.io/permission-delegation-api/) -- see [its examples section](https://noncombatant.github.io/permission-delegation-api/#examples) for a quick glance at how this works; but the syntax it shows is in line what I proposed. I understand that the Feature Policy document and the Permission Delegation document may be merged, but my understanding is that this is a document structure change with no protocol implications.

As far as I know, that's the extent of the discussion on this topic, which is why I find the changes that were just merged in to be surprising.

PR #292, opened 2016-10-19
Introduce paymenRequestID

rvm4:


ianbjacobs:
Thanks, Roy. I've put on tomorrow's agenda:
 https://github.com/w3c/webpayments/wiki/Agenda-20161020

One comment: would it make sense to add even more to make clear that the status information is payment method specific, as in change:

 "The state of the payment as a member of <code>PaymentStatus</code>."

to:

 "Optional payment method specific information about status."

Ian

dlongley:
I see some good opportunities to coordinate with the HTTP API here. The text for payment status update/discovery/etc could be defined in the HTTP API spec and the browser payment API could say how to provide a callback URL in the JS API where that service could be taken advantage of.

ianbjacobs:
@dlongley,

I have been thinking about HTTP API and messages spec recently. I am not sure that we need the HTTP spec. It describes how to use (vanilla) HTTPS for a particular use case, but I can think of multiple use cases and don't think we need the standard around the message flow for these use cases. Here are the use cases I was thinking of:
- A pings B to know what to pay. B tells A. A pays B. (That's what HTTP API covers today).
- A knows what to pay B so just pays B. (Use case: recurring payment to known endpoint.)
- B asks A to be paid. A pays B. (Use case: how I pay my utilities today - I am billed and I pay.)

The message spec is common to those use cases and probably others. That's why my current thinking is that the message spec should be the focus. The rest, as I understand it, is "how to get that information between parties" and that might happen in lots of different ways even beyond HTTP (e.g., push notifications) just using existing standard protocols.

dlongley:
@ianbjacobs,

That is the original intent of the messages spec -- to be able to say how these messages are modeled and appear so that you can reuse them with a variety of different protocols. That's exactly the reason for the abstraction.

That being said, that doesn't necessarily mean we don't need a spec to say "If you want to send messages around using HTTP, here's how you do it". The same goes for "If you want to send messages using the payment request API, or with push notifications, or with X, or Y ... here's how you do it." We should only (ideally) need to define the messages once (in the messages spec) and then, when necessary, we can define specifications for using particular protocols for communicating those messages. We should be careful that we don't get too far off topic here though.

In this particular case, we want to make sure we tell app developers what they're supposed to do with the callback URL the merchant provides to them. Since the intent is to use it to communicate information over HTTP, it seems prudent to define that interaction in the HTTP spec as an endpoint a server can provide (and a client/payment app can make use of). Given that there are many different ways to communicate over HTTP, we should say exactly what the payment app should do here (and it doesn't need to be complicated, e.g. POST payment message X to the URL) so there's interoperability. Then, here in the payment request API, we can just define where the merchant would specify the callback in order for the payment app to find it.

adrianhopebailie:
@rvm4 In general I really like this but I'm not sure I see the value of the callback if it's in the `PaymentMethodData`.

As a merchant, the advantage of this for me is that I only implement a single public facing endpoint to process callbacks. These requests are always `PaymentResponse`s so the outward facing code I must deploy will seldom , if ever, change even if I add support for new payment methods.

index.html
```diff
@@ -1106,6 +1132,10 @@ <h2 id="state-transitions" class="informative">State transitions</h2>
       <dt><dfn id="dom-paymentcomplete-fail"><code>"fail"</code></dfn></dt>
       <dd>Indicates that processing of the payment failed. The <a>user agent</a> MAY display UI indicating
         failure.</dd>
+      <dt><dfn id="dom-paymentcomplete-fail"><code>"processing"</code></dfn></dt>
```
zkoch: Probably need to change this dfn id to be "dom-paymentcomplete-processing"

zkoch:
A few thoughts:
- For the callback_url, I'm concerned that since it's the payment method that's going to be returning data to that URL, we won't have any control over the response. It's difficult to build a server-side endpoint when you're not sure of the response. We can recommend something, but we can't make any guarantees.
- At least at the beginning, payment methods that already have their own system for callback support and going to continue to maintain their current systems. This means some methods will support this generic callback_url, and some won't, which is a bit of a pain for developers I would imagine.

Something that @adrianba brought up that I thought was interesting: why not let the merchant create the transaction id and pass it into PR? They're already doing this anyway for non PR-based transactions, and they're going to have to continue doing this since not all browsers will support PR, so it seems easier on the merchant to let them create.

ianbjacobs:
We discussed this at the 27 October call:
 https://www.w3.org/2016/10/27-wpwg-minutes#item03

There seemed to be support on the call for merchants providing transaction IDs (rather than browsers generating them). 

Ian

adrianhopebailie:
PROPOSAL: Make tx id an optional parameter to the `PaymentRequest` constructor and generate it if it's not provided?

adamroach:
@adrianhopebailie 

> PROPOSAL: Make tx id an optional parameter to the PaymentRequest constructor and generate it if it's not provided?

I think that works as long as any provided ID ends up scoped to the requestor's origin in some way.

rvm4:
@zkoch my intention was that the callbackURL be passed something structurally the same as the PaymentResponse. It is true that every payment app will have its own payload, but we have already accounted for that in the PaymentResponse object with a freeform place to put whatever you want.

I think for the purposes of this PR I'm happy just getting the transactionID in there. We can continue talking about how best to accomodate server callbacks in our weekly calls.

index.html
```diff
@@ -1024,12 +1040,14 @@ <h2 id="state-transitions" class="informative">State transitions</h2>
 <section>
   <h2>PaymentResponse interface</h2>
       <pre class="idl">
-        enum PaymentComplete { "success", "fail", "" };
+        enum PaymentStatus { "success", "fail", "processing", "" };
```
stan-stripe: I believe we want to remove the `processing` from here pending a better understanding on how the merchant would get notified/retrieve that the payment is succeeded?

nickjshearer:
Yes, I think the callbackURL requires further discussion. 

I agree with Zach's point re: the merchant generating the ID. I am not massively keen on the browser generating the transaction ID, certainly as it's currently written. "Unique" is not a conformance key word. Do we mean globally unique? Unique to the merchant? What's its format?

Requiring user agents to generate a unique identifier and then not defining the parameters of that unique identifier will cause problems. Developers may make assumptions (like the length, encoding, use of non ASCII chars) about the format. They might assume, say, the user agent will always return a 128-bit UUID and make that into their database, based on what the major user agents are doing. Then that user agent subsequently changes it to be a 256-bit UUID, _which is within the specification_ - but the developer now has a broken implementation.

So my preference would be to simply drop the requirement for the user agent to generate a transaction ID if none exists. But if we do keep it, the specification should at the very least provide a clear definition of what that ID should be made up of (not necessarily the formula, but length, permitted characters, etc).

halindrome:
I am okay with having a generated Transaction ID and requiring that it be a 128bit UUID.

stan-stripe:
I'm a bit worried that a 128bit UUID will be hard to leverage on statement descriptors of a large number (if not all) payment methods, which would jeopardize some of the value of such IDs: being a customer visible value that can be used for support.

I also think they should be generated by the merchants, as they are incentivized to have IDs that can be properly displayed to customers on the payment methods they support.

halindrome:
Apologies if I was unclear.  I ABSOLUTELY believe that the transaction ID has a domain that is the merchant and a range that is whatever the merchant chooses.  Of course the merchant should generate a meaningful transaction ID.  The point of an API-generated one is to provide traceability in the event that the merchant fails to provide one.  There does not seem to be support for just requiring a Transaction ID from the merchant, so this is a reasonable compromise IMHO.

As to things like data type, encoding, size, etc...  I am okay saying it is a UTF-8 encoded string that is no more than some number of octets long (256?).  That will assist people who have databases, etc.  However, since we have no control over merchants unless things like size limits are enforced, I would be okay requiring there be an exception thrown if the Transaction ID from the merchant does not match the requirements.  Better kill it on the front end than start passing it along and break something elsewhere in the ecosystem.

adrianhopebailie:
I would advocate that we do not have an identifier for the **financial transaction** but for the **payment request**.

These are subtly different.

We have no scope to define transaction ids, as these are very specific to the payment processing, however we are attempting to provide a robust messaging infrastructure in which a request may be passed via the API and responded to via another channel.

I see the request ID (and I recommend changing the name) as a short lived identifier that allows a merchant to reconcile requests and responses on different channels. For this reason it seems like a good idea for the browser to generate this id using a well known format (128-bit UUID seems like a good option) and only accept a merchant provided ID in the same format.

Payments apps, merchants and PSPs should take responsibility for keeping track of which requests (identified by a request id) resulted in actual payments (identified by a tx id).

halindrome:
Well.... hmm.  The problem I was trying to solve in the first place when I brought this up was that the payment app needs a way to help consumers track their transactions.  Not just a short lived way.  A long-term way.  And that way cannot be inside of the payment method data envelope.  We don't look in there!  I don't mind if we call it a payment request ID or whatever.  But it is a way to correlate merchants / requests with actual payments that were made.  Right now the Payment App (a smart wallet, remember) can't really be smart about this at all.

index.html
stan-stripe: My bad you still need it to communicate that the payment is not succeeded yet. Maybe `pending` is more general than `processing` but no strong feeling here.

rvm4:
I believe this was deemed good to go, merging

PR #294, opened 2016-11-10
Combine scope and goal + relocate/reduce conformance

marcoscaceres:


marcoscaceres:
I've added the "User agents MAY implement algorithm..." text to the "tidy" branch. 

This is good to merge!

PR #312, opened 2016-11-16
normative(PaymentComplete): replace '' with 'unknown' (closes #308)

### LINKED ISSUE #308: enum PaymentComplete's ""
Can we please give "" a proper enum value to PaymentComplete's "". I believe we've only used "" as a means to recover from where we screwed up an API in the past... given that this spec is still evolving, there is no reason for `""` to be used, instead of, say, something more meaningful like "unknown" or "dismissed".

marcoscaceres:


marcoscaceres:
Hmm... maybe this should be "abort". @rsolomakhin, do you have any thoughts on this?

rsolomakhin:
I would prefer not to change this part of the spec. If we do change it, `complete('unknown')` matches the meaning of `complete('')`. Let's not use `'abort'`, because it can be confused with `abort()` method, whose meaning and behavior are different.

adrianba:
It's definitely not `'abort'` - that would be a very different meaning. You call this with `''` or `'unknown'` when the transaction is in progress not when you are aborting.

marcoscaceres:
@rsolomakhin, yeah, ignore my "abort" suggestion. 

However, I'd  really like to see `""` changed to `"unknown"` on the Moz side for the reasons I mentioned in #308.

zkoch:
This LGTM

PR #317, opened 2016-11-17
Example update

ianbjacobs:
Uses ‘basic-card’ from:
https://w3c.github.io/webpayments-methods-card/

Uses ‘aFamousBrand’ instead of a trademarked string for now


marcoscaceres:
@ianbjacobs, could you please rebase.

index.html
```diff
@@ -1355,12 +1359,18 @@ <h2 id="state-transitions" class="informative">State transitions</h2>
...
-    <h2>Exposing user information</h2>
+    <h3>Exposing user information</h3>
     <p>
       The <a>user agent</a> should never share information about the user to the web page
       (such as the shipping address) without user consent.
     </p>
   </section>
+  <section>
+    <h3>Incognito Mode</h3>
+    <ul>
```
marcoscaceres, line 1370: Seems odd to have this as an `ul`. Maybe switch back to paragraph. Also, I think the TAG was trying to nail down what to call "incognito mode" (that's quite Chrome specific terminology, I think).

index.html
```diff
@@ -1355,12 +1359,18 @@ <h2 id="state-transitions" class="informative">State transitions</h2>
...
+      <li>This API should remain available in a browser's "incognito mode".</li>
+      <li>Browsers that offer "persistent permission" features for sharing user payment information automatically with merchants should ignore those settings in icognito mode and require explicit user consent.</li>
```
marcoscaceres, line 1372: Persistent permissions should link to the Permission spec. I think it defines what that is.

index.html
```diff
@@ -297,7 +297,11 @@
           <pre class="example highlight">
             [
               {
-                supportedMethods: ["visa","bitcoin"]
+                supportedMethods: ["basic-card"]
+  	        data: {
```
marcoscaceres, line 301: nit: whitespace.

marcoscaceres (changes requested):

index.html
adamroach: > Also, I think the TAG was trying to nail down what to call "incognito mode" (that's quite Chrome specific terminology, I think).

I believe the impetus for the TAG reviewing this issue is because Ian and I brought it to their attention. :)  We are, in any case, tracking it closely, and will adjust the document when they have provided guidance.

marcoscaceres: Ok, cool.

marcoscaceres:
Helping @ianbjacobs with the rebase. Will put this on a new local branch.

PR #319, opened 2016-11-21
chore: xref internal slots

marcoscaceres:
This is only a fix to the markup. No editorial or normative change. 

Makes sure the internal slots link to their definitions. Otherwise it's annoying to have to look up the type of each internal slot in the corresponding "internal slots" tables.


halindrome:
LGTM

PR #329, opened 2016-11-24
editorial: clarify examples

marcoscaceres:
* Add titles to the example, so it's easier to understand what they show
* Use async/await where possible for clarity

### DIFF
--- a/index.html
+++ b/index.html
@@ -250,8 +250,9 @@
           payment.addEventListener("shippingaddresschange", event =&gt; {
               // Process shipping address change
           });
-          const paymentResponse = await payment.show();
+          let paymentResponse;
           try {
+            paymentResponse = await payment.show();
             // paymentResponse.methodName contains the selected payment method.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -240,21 +240,30 @@
...
-    }).catch(function(err) {
-      console.error("Uh oh, something bad happened", err.message);
-    });
-  </pre>
+        async function doPaymentReqest () {
+          const payment = new PaymentRequest(methodData, details, options);
+          payment.addEventListener("shippingaddresschange", event =&gt; {
+              // Process shipping address change
+          });
+          const paymentResponse = await payment.show();
```
adrianba: Doesn't this need to be inside the try block?

marcoscaceres:
rebased

index.html
marcoscaceres: Ah yeah, good catch :) The joys of coding examples without the implementation 💃

PR #355, opened 2016-11-30
editorial: PaymentRequest ctor, describe errors (closes #290)

### LINKED ISSUE #290: PaymentRequest constructor errors should be properly annotated
In the section that details the algorithm for the Payment Request constructor, there are multiple places where a TypeError is thrown. If these errors are not annotated (_TypeError.message_ is set), it will be difficult for a developer to determine why the TypeError was thrown (as there are multiple ways it can be thrown in the constructor algorithm).

https://w3c.github.io/browser-payment-api/#paymentrequest-constructor

After chatting a bit with @rsolomakhin, he noted that Chromium does annotate the TypeErrors. The specification shouldn't make a normative requirement on what the text should be, but should provide hints via "optional messages" much like the DOM4 entry in the Dependencies section: https://w3c.github.io/browser-payment-api/#h-dependencies so that developers working across multiple browsers can depend on more or less the same TypeError annotation text.

@rsolomakhin identified the following _TypeError.message_s that can be thrown in the PaymentRequest constructor in Chromium:

https://chromium.googlesource.com/chromium/src/+/0f7bc8677b8f04c46bbbd574b98faa5a2161769a/third_party/WebKit/Source/modules/payments/PaymentRequest.cpp#166
* [x] Must specify at least one payment method
* [ ] Must specify at least one payment method identifier
* [x] Must specify at least one payment method identifier
* [x] Must specify total
* [x] Currency amount required
* [x] Total amount value should be non-negative
* [ ] <del>Currency code required</del>
* [ ] <del>Must specify at least one payment details modifier</del>
* [ ] <del>Currency value required</del>
* [ ] <del>ShippingOption id required</del>
* [ ] <del>Duplicate shipping option identifiers are not allowed</del>
* [ ] <del>Duplicate payment method identifiers are not allowed</del>
* [ ] Duplicate payment method identifiers are not allowed
* [ ] Unable to parse payment method specific data
* [ ] Data should be a JSON-serializable object
* [ ] Item label  #required

I'll try to submit a PR to fix this if it's agreed that this is a good thing to include in the spec.

marcoscaceres:
Describes to developers why the errors are being thrown.

### DIFF
--- a/index.html
+++ b/index.html
@@ -325,12 +325,20 @@
           The <a>PaymentRequest</a> constructor MUST act as follows:
         </p>
         <ol>
-          <li>If the length of the <code>methodData</code> sequence is zero,
-          then <a>throw</a> a <a>TypeError</a>.
+          <li>If the length of the <var>methodData</var> sequence is zero, then
+          <a>throw</a> a <a>TypeError</a>; optionally informing the developer
+          that at least one <a>payment method</a> is required.
           </li>
-          <li>For each <a>PaymentMethodData</a> dictionary, if the length of
-          the <code>supportedMethods</code> sequence is zero, then <a>throw</a>
-          a <a>TypeError</a>.
+          <li>For each <var>paymentMethod</var> of <var>methodData</var>:
+            <ol>
+              <li>If the length of the <a data-lt=
+              "PaymentMethodData.supportedMethods">supportedMethods</a>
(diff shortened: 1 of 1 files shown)


adrianba (approved):

zkoch:
LGTM

PR #357, opened 2016-12-02
fix: don't validate display items (closes #344)

### LINKED ISSUE #344: What does it mean for a user agent to validate displayItems?
https://w3c.github.io/browser-payment-api/#paymentdetails-dictionary description of displayItems says:

>  The user agent MAY validate that the total amount is the sum of these items

If the UA does that... what does that mean?  What is the corresponding observable behavior or processing model?

marcoscaceres:


zkoch:
LGTM

Issue #332, opened 2016-11-30
The origin restrictions in the PaymentRequest are not nearly strong enough


bzbarsky:
Step 3 of the steps in https://w3c.github.io/browser-payment-api/#constructor (I see no way to link to the steps directly) says:

> If the browsing context of the script calling the constructor is a nested browsing context whose origin is different from the top-level browsing context's origin and the nested browsing context is not allowed to make payment requests, then throw a SecurityError. 

This has at least two problems (not counting the already-filed https://github.com/w3c/browser-payment-api/issues/324 and <https://github.com/w3c/browser-payment-api/issues/323>):

1.  Consider the case when our toplevel origin is A, which loads a subframe from origin B, which loads a subframe from origin A.  This innermost frame would be allowed to make payment requests, as the spec is currently written, even if the B subframe is not.  That seems very odd to me.  The secure thing here would be to only allow payment requests in a document if all its ancestor documents can make payment requests, plus whatever other criteria are desired.  This is how feature policy would work, for example.
2.  It's not clear to me that any of this considers that various of the documents involved may not be active, or may not be fully active.  It's hard to tell, exactly, because the other two issues I cite above mean the processing model is really not defined at all here.

adrianhopebailie:
> Consider the case when our toplevel origin is A, which loads a subframe from origin B, which loads a subframe from origin A. This innermost frame would be allowed to make payment requests, as the spec is currently written, even if the B subframe is not. That seems very odd to me.

Why does that seem odd?

On the contrary I would think it's odd if A had to give permission to call payment request to B as a way to let B embed a frame from A that can.

Use case example:

Merchant A embeds frames from B who is a third-party that manages their advertising across all properties. A wishes to give B the ability to show advertisements with "click-to-buy" functionality but doesn't want B to be able to invoke the payment API directly.

bzbarsky:
> Why does that seem odd?

Because it allows B to trigger payment requests from A by when A might not expect it to.

Note that the proposed feature policy API has exactly the semantics I'm talking about: if a document doesn't have permission to do something, neither do any of its descendants.  And I believe that having the payment permissions model be describable by feature policy is an explicit goal.   So if it can't be, one or the other is wrong and needs to change.

adrianhopebailie:
> Because it allows B to trigger payment requests from A by when A might not expect it to.

I disagree. Both contexts from origin A (the top-level and innermost contexts) can trigger payment requests but the other context (with origin B) cannot.

[...]

Issue #324, opened 2016-11-21
Browsing contexts don't have an "origin"


bzbarsky:
https://w3c.github.io/browser-payment-api/#constructor is talking about "a nested browsing context whose origin is different from the top-level browsing context's origin", but browsing contexts do not have an origin attached to them per se.  They can contain things that have origins, of course.

I _suspect_ what you want is an "allowed to make payment requests" flag on Document or Window and make sure it gets set correctly in the HTML spec.

PR #359, opened 2016-12-05
Reference W3C HTML for allowpaymentrequest definition

sideshowbarker:
Following https://github.com/w3c/html/pull/743, which adds the definition of the `<iframe>` `allowpaymentrequest` attribute to W3C HTML, this change updates the W3C Payment Request API to reference W3C HTML for that definition of `allowpaymentrequest` from W3C HTML.

See also https://github.com/w3c/browser-payment-api/issues/311

### DIFF
--- a/index.html
+++ b/index.html
@@ -1521,44 +1521,23 @@
     </section>
     <section>
       <h2>
-        PaymentRequest and iframes
+        PaymentRequest and <code>iframe</code> elements
       </h2>
       <p>
-        There are some circumstances where a cross-origin <a>iframe</a> wants
-        to make a payment request. A cross-origin iframe needs explicit
-        permission from the embedding page to invoke the payment request API.
+        There are some circumstances where a cross-origin
+        <a><code>iframe</code></a> wants to make a payment request. A
+        cross-origin <a><code>iframe</code></a> needs explicit permission
+        from the embedding page to invoke the payment request API.
       </p>
       <p>
-        The <a>HTMLIFrameElement</a> is extended with an
-        <dfn><code>allowpaymentrequest</code></dfn> content attribute.
-        <a>allowpaymentrequest</a> is a <a>boolean attribute</a>. When
-        specified, it indicates that scripts in the iframe element's browsing
-        context are <dfn>allowed to make payment requests</dfn> (if it's not
-        blocked for other reasons, e.g., there is another ancestor iframe
-        without this attribute set).
+        When an <a><code>iframe</code></a> element has an
+        <a><code>allowpaymentrequest</code></a> attribute, and does not
+        have any ancestor <a><code>iframe</code></a> element without an
+        <a><code>allowpaymentrequest</code></a> attribute, then
+        <code>Document</code> objects in the <a><code>iframe</code></a>
+        element's <span>browsing context</span> are <dfn>allowed to
+        make payment requests</dfn>.
       </p>
-      <section data-dfn-for="HTMLIFrameElement">
-        <h2>
-          <code>HTMLIFrameElement</code> extension
-        </h2>
-        <p>
-          The iframe DOM interface is extended as follows:
-        </p>
-        <pre class="idl">
-          partial interface HTMLIFrameElement {
-              attribute boolean allowPaymentRequest;
-          };
-        </pre>
-        <dl>
-          <dt>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1518,44 +1518,23 @@ <h2 id="state-transitions" class="informative">
...
-        The <a>HTMLIFrameElement</a> is extended with an
-        <dfn><code>allowpaymentrequest</code></dfn> content attribute.
-        <a>allowpaymentrequest</a> is a <a>boolean attribute</a>. When
-        specified, it indicates that scripts in the iframe element's browsing
-        context are <dfn>allowed to make payment requests</dfn> (if it's not
-        blocked for other reasons, e.g., there is another ancestor iframe
-        without this attribute set).
+        When an <a><code>iframe</code></a> element has an
+        <a><code>allowpaymentrequest</code></a> attribute, and does not
+        have any ancestor <a><code>iframe</code></a> element without an
```
zcorpan: This seems wrong. This reads like it's talking about ancestor the DOM concept. But you should hook into "allowed to use" instead, which handles nested browsing contexts. See Fullscreen.

sideshowbarker: > hook into "allowed to use" instead, which handles nested browsing contexts

Thanks, see https://github.com/w3c/browser-payment-api/pull/359/commits/4f59cf90ea05ecb40729453193e7c248a90bef0f, which attempts that.

index.html
```diff
@@ -343,8 +343,11 @@
           <li>If the <a>browsing context</a> of the script calling the
           constructor is a <a>nested browsing context</a> whose origin is
           different from the <a>top-level browsing context</a>'s origin and the
-          nested browsing context is not <a>allowed to make payment
-          requests</a>, then <a>throw</a> a <a>SecurityError</a>.
+          nested <a>browsing context</a>'s <a>browsing context container</a> is
+          an <a><code>iframe</code></a> element whose <a>node document<a> is
+          not allowed to use the feature indicated by attribute name
```
zcorpan: `<a>allowed to use</a>`?

index.html
```diff
@@ -343,8 +343,11 @@
           <li>If the <a>browsing context</a> of the script calling the
           constructor is a <a>nested browsing context</a> whose origin is
           different from the <a>top-level browsing context</a>'s origin and the
-          nested browsing context is not <a>allowed to make payment
-          requests</a>, then <a>throw</a> a <a>SecurityError</a>.
+          nested <a>browsing context</a>'s <a>browsing context container</a> is
+          an <a><code>iframe</code></a> element whose <a>node document</a> is
```
zcorpan: So if it's not an `iframe` (e.g. it's `frame`/`object`/`embed`), it wouldn't throw. That's bad. Remove `iframe` check here I think.

index.html
```diff
@@ -343,8 +343,11 @@
           <li>If the <a>browsing context</a> of the script calling the
           constructor is a <a>nested browsing context</a> whose origin is
```
zcorpan: I think this PR should be blocked on fixing these issues:
https://github.com/w3c/browser-payment-api/issues/324
https://github.com/w3c/browser-payment-api/issues/332

The whole paragraph doesn't any make sense right now and does not seem actually secure.

zcorpan (changes requested):

index.html
rsolomakhin: Do we plan to allow nested `frame`/`object`/`embed` to use `PaymentRequest`? I thought not.

zcorpan: No. But the current text would allow that.

index.html
sideshowbarker: > I think this PR should be blocked on fixing these issues:
> #324
> #332

Agreed. Though I wasn’t sure that this PR itself was the right place to fix those. I guess I imagined this PR branch could be rebased once those fixes got made. But I suppose we could instead make the changes on this branch as part of this PR.

index.html
sideshowbarker: > Remove `iframe` check here I think.

OK, removed in https://github.com/w3c/browser-payment-api/pull/359/commits/681837dea0c5232d2ef3c97fa87dfc447fbac479

index.html
zcorpan: Either way 😊

PR #368, opened 2016-12-13
Minor typographical fixes

### LINKED ISSUE #353: "store" value in [[internalSlot]]
The spec uses terminology like:

> Store details into request@[[details]]. 

But it's not clear what "store" means. We should just "set" internal slots to things.

domenic:
- Fix #353: set slots to values, don't store in slots
- Use `<var>` instead of `<em>` for variables
- Use `<i>` instead of `<em>` for state names
- Don't use `<em>` or `<code>` for literal values like null, undefined, false, and true.
- Run HTML Tidy

### DIFF
--- a/index.html
+++ b/index.html
@@ -356,23 +356,21 @@
               </li>
             </ol>
           </li>
-          <li>If the <a>current settings object</a> has a <a>responsible browsing
-            context</a> that is not a <a>top-level browsing context</a>, then
+          <li>If the <a>current settings object</a> has a <a>responsible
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Looks good! thanks @domenic.

marcoscaceres:
D'oh, I'd assume this was the first part - as it come in first.

domenic:
This one first is good! I've rebased #369 and #370, although for some reason #370 is still showing merge conflicts.

marcoscaceres:
Yeah, #370 confused me. Cool.

Issue #335, opened 2016-11-30
A PaymentDetailsModifier may not have an additionalDisplayItems value present


bzbarsky:
https://w3c.github.io/browser-payment-api/#constructor step 13 says:

> For each PaymentDetailsModifier in details.modifiers, if the additionalDisplayItems sequence contains any PaymentItem objects with an amount that is not a valid decimal monetary value, then throw a TypeError. 

But there may not be a value present for additionalDisplayItems at all in the dictionary.

Furthemore, the `amount` of a `PaymentItem` is not a string, so you can't check whether it's a "valid decimal monetary value".

Issue #321, opened 2016-11-21
"contains a value for" isn't defined for dictionaries


bzbarsky:
The IDL spec defines the concept of a dictionary member being "present" or not at <https://heycam.github.io/webidl/#dfn-present>.  Please use that instead of making up new terminology here.

bzbarsky:
Same thing for various places that use "is supplied" when they mean "is present".

domenic:
This is now fixed and can be closed, through a combination of a couple PRs.

sideshowbarker:
Closing per https://github.com/w3c/browser-payment-api/issues/321#issuecomment-270523227

Issue #371, opened 2016-12-13
waitForUpdate is not defined


domenic:
I can't tell who is supposed to hold this. Is it one per document? One per PaymentRequest?

domenic:
This still remains mysterious. @adrianba @rsolomakhin any ideas what [[waitForUpdate]] is about?

dlongley:
@domenic,

`[[waitForUpdate]]` is an internal slot in `PaymentRequestUpdateEvent`. I believe it gets flipped when someone calls `updateWith` from within an event handler so that:

1. An invalid state exception will be thrown if `updateWith` is called more than once.
2. As a guard that will prevent event handling from completing until unset again.

https://github.com/domenic/browser-payment-api/blob/85074dda44c64c088f87f961ac503783dc22114b/index.html#L1623-L1624

I believe its analog is the "[wait to respond flag](https://www.w3.org/TR/service-workers/#wait-to-respond-flag)" used in the fetch spec's [respondWith method](https://www.w3.org/TR/service-workers/#fetch-event-respondwith-method).

Does that answer your question?

domenic:
Yes, that's very helpful; thank you!

PR #372, opened 2016-12-13
Various fixes to updateWith

### LINKED ISSUE #345: The shippingOptions field is used in more cases than the spec claims, as far as I can tell
https://w3c.github.io/browser-payment-api/#dom-paymentdetails says this in the `shippingOptions` section:

> The shippingOptions field is only used if the PaymentRequest was constructed with PaymentOptions requestShipping set to true. 

but this doesn't match the normative text in updateWith, as far as I can tell: that will use the shippingOptions no matter how the PaymentRequest was constructed...

### LINKED ISSUE #347: The target of a PaymentRequestUpdateEvent may not be a PaymentRequest object
https://w3c.github.io/browser-payment-api/#updatewith-method step 1 says:

> Let target be the PaymentRequest object that is the target of the event. 

But target can be null, or some EventTarget that is not a PaymentRequest.  Behavior in that case needs to be defined (presumably to throw).

### LINKED ISSUE #348: It's probably a good idea to use https://www.w3.org/2001/tag/doc/promises-guide#shorthand-reacting in updateWith
Currently updateWith is explicitly poking at Promise internal state.  It's probably better to use the terminology in https://www.w3.org/2001/tag/doc/promises-guide#shorthand-reacting for consistency with other specifications.

This is particularly important because the two options  are not actually black-box identical as far as I can tell.  UAs are going to have a codepath that looks like the shorthands defined in https://www.w3.org/2001/tag/doc/promises-guide#shorthand-reacting but may not have one for doing "Wait until d settles" without actually invoking the initial value of `Promise.prototype.then`.

### LINKED ISSUE #349: The "if d is resolved" section of updateWith needs to be more explicit about how it gets a PaymentDetails dictionary
Right now it says:

> If d is resolved with details and details is a PaymentDetails dictionary, then

but what d is resolved with is an ES value.  What you probably want to do is to convert this value to a dictionary using <https://heycam.github.io/webidl/#es-to-dictionary>.  This can throw exceptions; the spec needs to say what happens if an exception is thrown.

domenic:
- Fixes #348 by using the promises guide terms
- Fixes a bug related to #335 by appropriately iterating over the additionalDisplayItems
- <ins>Fixes #347 by throwing on non-PaymentRequest targets</ins>
- <ins>Fixes #349 by performing dictionary conversion</ins>
- <ins>Fixes part of #321 (not the constructor part yet) by using "is present" appropriately for dictionaries</ins>
- <ins>Fixes #345 by only looking at shippingOptions when requestShipping is true</ins>
- Fixes the algorithm to correctly reset [[waitForUpdate]] / target.[[updating]] / the UI even when the promise is rejected, instead of aborting
- Various phrasing fixes around how to use and access dictionaries

---

On top of #368 and should be merged after that. <del>Will clash with #370 so whichever lands first I'll rebase the other.</del> <ins>Also on top of #370 so should be merged after that.</ins>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1631,8 +1631,12 @@
             The <dfn>updateWith</dfn> method MUST act as follows:
           </p>
           <ol>
-            <li>Let <var>target</var> be the <a>PaymentRequest</a> object that
-            is the target of the event.
+            <li>Let <var>target</var> be the value of this
+            <code>PaymentRequestUpdateEvent</code> event's <code>target</code>
+            attribute.
+            </li>
+            <li>If <var>target</var> is not a <code>PaymentRequest</code>
+            object, then <a>throw</a> a <a>TypeError</a>.
             </li>
             <li>If the <a>dispatch flag</a> is unset, then <a>throw</a> an <a>
               InvalidStateError</a>.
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@domenic, can you rebase this one please?

domenic:
Done

index.html
```diff
@@ -1692,115 +1699,178 @@ <h2 id="state-transitions" class="informative">
...
-            <li>Set <var>target</var>.<a>[[\updating]]</a> to false.
-            </li>
-            <li>The <a>user agent</a> should update the user interface based on
-            any changed values in <var>target</var>. The user agent SHOULD
-            re-enable user interface elements that might have been disabled in
-            the steps above if appropriate.
+            <li>In either case, run the following steps, after either the upon
+            rejection or upon fulfillment steps have concluded:
+              <ol>
+                <li>Set [[\waitForUpdate]] to false.
```
marcoscaceres, line 1867: nit: missing `<a>` around waitForUpdate

marcoscaceres (approved):

index.html
domenic: It's the same everywhere in the spec. See #371.

marcoscaceres: Ah, ok. Will see what that belongs to. @adrianba?

zkoch:
LGTM

PR #374, opened 2016-12-13
Various fixes to the PaymentRequest constructor

### LINKED ISSUE #333: PaymentDetails may not have values present for various sequences, but the PaymentRequest constructor assumes that it does
https://w3c.github.io/browser-payment-api/#constructor step 7 says:

> If the details.displayItems sequence contains any PaymentItem objects with an amount that is not a valid decimal monetary value, then throw a TypeError. 

But there may not be  a displayItems value present at all in `details`.  Similar in step 8 for shippingOptions.

### LINKED ISSUE #334: A `PaymentItem` can't be a "valid decimal monetary value"
Step 11 of https://w3c.github.io/browser-payment-api/#constructor says:

> For each PaymentDetailsModifier in details.modifiers, if the total field is supplied and is not a valid decimal monetary value, then throw a TypeError. 

But the "total" member of `PaymentDetailsModifier` is a `PaymentItem`, not a string.  So checking whether it's a "valid decimal monetary value" is nonsensical.

### LINKED ISSUE #335: A PaymentDetailsModifier may not have an additionalDisplayItems value present

### LINKED ISSUE #373: PaymentRequest constructor validation ordering does not match Blink at all
Compare https://cs.chromium.org/chromium/src/third_party/WebKit/Source/modules/payments/PaymentRequest.cpp?sq=package:chromium&rcl=1481631594&l=378 with the current spec.

I will work to align the spec with Blink, since it's open-source and easy to read off. I guess technically some of these aren't testable since they all throw the same TypeError, but it'd be observable to users based on the message thrown.

domenic:
- Fixes #373: align ordering of steps with Blink's implementation, except check for allowed-to-use first since that makes more sense
- Fixes #334: properly get the total.amount.value instead of pretending total is a string
- Fixes #333 and fixes #335: do not assume various dictionaries are present
- Fixes part of #321: use proper dictionary terminology
- Ensures request.[[details]].shippingOptions/modifiers is always a sequence, instead of sometimes left as not present
- Fixes various algorithm structure and typographic issues

---

This is on top of #368 and #370 and can only be merged after them.

### DIFF
--- a/index.html
+++ b/index.html
@@ -346,21 +346,6 @@
           <li>
             If a <code>paymentRequestID</code> was not provided during construction, generate a <code>paymentRequestID</code>.  
           </li>
-          <li>If the length of the <var>methodData</var> sequence is zero, then
-          <a>throw</a> a <a>TypeError</a>; optionally informing the developer
-          that at least one <a>payment method</a> is required.
-          </li>
-          <li>For each <var>paymentMethod</var> of <var>methodData</var>:
-            <ol>
-              <li>If the length of the <a data-lt=
-              "PaymentMethodData.supportedMethods">supportedMethods</a>
-              sequence is zero, then <a>throw</a> a <a>TypeError</a>;
-              optionally informing the developer that each <a>payment
-              method</a> needs to include at least one <a>payment method
-              identifier</a>.
-              </li>
-            </ol>
-          </li>
           <li>If the <a>current settings object</a> has a <a>responsible
           browsing context</a> that is not a <a>top-level browsing context</a>,
           then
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@domenic, this one also now needs rebase. So. Close.

domenic:
Just hit refresh :)

marcoscaceres:
ah 👍

adrianba:
Is there a need to reorder the checks to align with Blink? I think this might make our implementation out of compliance with the spec. Need to check on the impact of this but would probably prefer we don't change ordering if we don't need to.

domenic:
In general some ordering fixes would be good. E.g. the current spec splits up checking methodData into two locations of the algorithm, interspersing checks and processing of other arguments between them, which is just strange (and requires looping over the sequence twice).

If there's a different ordering to the top-level "sections" desired, that'd be fine too. But the grouping here makes more sense than the current spec.

In other words, this PR's ordering is:

- allowed to use check
- process payment methods
- process the total
- process displayItems
- process shipping options
- process payment details modifiers
- process error

Any reordering of these seems reasonable, although I think the allowed to use check should be first. But the current spec, which splits up several of these steps into two or three separate substeps which are then sprinkled around the algorithm, seems bad.

marcoscaceres:
@domenic can you rebase? 

I also think @domenic refactoring is an improvement here - if only from an editorial perspective. We can do any additional reordering after that.

domenic:
@marcoscaceres done

marcoscaceres:
@adrianba, want your ok to merge this (and that Edge is willing to make the necessary changes to implementation, if any).

adrianba (approved):
I can live with this.

PR #381, opened 2016-12-14
Validate payment method data

domenic:
This is part of w3c/webpayments-methods-card#20. What remains is for the basic-card specification to define its validate payment method data algorithm.

---

This is on top of #374 and should not be merged until that is.


domenic:
Closing since per https://github.com/w3c/browser-payment-api/issues/307#issuecomment-267146564 this cannot be done synchronously; it needs to happen in another process.

Issue #346, opened 2016-11-30
The "data" member of  PaymentDetailsModifier doesn't seem to be used for anything


bzbarsky:
There's no processing model defined for it that I can tell.

nickjshearer:
Isn't it simply intended to be passed back to the payment instrument?

ianbjacobs:
@nickjshearer wrote:

"Isn't it simply intended to be passed back to the payment instrument?"

Yes, except I would say "to the payment app" that handles that payment method. The payment app could be the browser or a third party payment app.

Ian

bzbarsky:
You can't pass an object to a third-party payment app.  See also https://github.com/w3c/browser-payment-api/issues/307

rsolomakhin:
We can run `JSON.stringify(data)` and pass the result to the third-party payment app.

bzbarsky:
OK, that's fine.  The spec then needs to:

1. Define exactly when JSON.stringify runs, because it can have observably side-effects.  Again, see https://github.com/w3c/browser-payment-api/issues/307

2. Define that the resulting string is passed somewhere, wherever that is.

domenic:
Can someone (@rsolomakhin?) help me understand exactly in which steps the data is passed to the payment app? For example, we determined that the methodData (parsed into [[parsedMethodData]]) is passed to the payment app in canMakePayment and in show. Is the idea that the payment apps should receive not only the method data, but also the payment details modifier data, at both of those points?

rsolomakhin:
Both `.canMakePayment()` and `.show()` should be sending the method-specific `data` to the payment app, because that can include some important filtering information. For example, `data` for `"basic-card"` payment method includes `supportedNetworks`. If merchant specifies `supportedNetworks: ["jcb"]`, but the only installed payment app is not aware of JCB, then `.canMakePayment()` should resolve with `false` and `.show()` should reject with `NotSupportedError`.

As for modifiers, that should be used for altering the price, which is important when completing a transaction, not when determining whether a transaction can procede. Therefore, modifiers should be sent to the payment app with `.show()`, but not in `.canMakePayment()`.

PR #382, opened 2016-12-14
Straighten out the payment method data flow

### LINKED ISSUE #307: The myth of JSON-serializable object
I might be wrong, but I recall "JSON-serializable object" coming up in the past in other specs and being problematic: 

> The term JSON-serializable object used in this specification means an object that can be serialized to a string using JSON.stringify and later deserialized back to an object using JSON.parse with no loss of data.

But I'm pretty sure it's a bit of a 🦄 - or it should be actually defined as to how one actually works that out. Right now, the spec basically implies:
```JS
try {
   JSON.parse(JSON.stringify(obj)); 
} catch(err){
   throw new Error("Not a JSON-serializable object");
}
```
 
That seems nasty. Could we check with public-script-coord what the right thing to do here is? 

There might be an actual object type that we can use here.

### LINKED ISSUE #338: The "data" member of PaymentMethodData doesn't seem to actually be used for anything
The PaymentRequest constructor enforces it being "JSON-serializable", whatever that means. .

But after that, nothing references that member.  Is it supposed to actually get serialized to JSON at some point, perhaps?

### LINKED ISSUE #354: Ambiguous conformance requirement in constructor
The following seems to be an authoring requirement: 

> The methodData supplied to the PaymentRequest constructor SHOULD be in the order of preference of the caller.

The following assertion seems to be out of place in the algorithm: 

> Implementations MAY show payment methods in this order if possible but should prioritize the preference of the user when presenting payment methods.

The above should be placed somewhere closer to when they payment request is actually shown.

domenic:
This fixes #338 and helps with #307 by stating more explicitly how the data member of PaymentMethodDetails is serialized to a string, stored, and later passed to the appropriate payment app.

Along the way, this fixes #354.

This does not completely fix #307, as the problematic JSON-serializable concept is still used for PaymentDetailsModifier (and is still not used by any part of the processing model; that's #346).

---

This is on top of #374 and should be merged only after that is.

If this looks correct, then to clarify how the basic card spec actually interacts with the main spec, we'll edit the basic card spec to say:

- Payment apps for basic card must determine if they support the requested payment method by first performing steps equivalent to parsing the supplied data using JSON.parse, then converting the resulting object to a BasicCardRequest dictionary, and inspecting the results.
- Payment apps for basic card must supply a result object back to the payment request flow in the form of a BasicCardResponse dictionary.

This will then solve https://github.com/w3c/webpayments-methods-card/issues/20.

### DIFF
--- a/index.html
+++ b/index.html
@@ -125,11 +125,18 @@
         every transaction: the merchant (e.g., an online web store), the buyer
         represented by the <a>user agent</a> (e.g., the user buying from the
         online web store), and the <dfn data-lt=
-        "payment method|payment methods">Payment Method</dfn> (e.g., credit
+        "payment method|payment methods">payment method</dfn> (e.g., credit
         card). Information necessary to process and confirm a transaction is
-        passed between the <a>Payment Method</a> and the merchant via the
+        passed between the <a>payment method</a> and the merchant via the
         <a>user agent</a> with the buyer confirming and authorizing as
         necessary across the flow.
+      </p>
+      <p>
+        The details of how to fulfill a payment request for a given <a>payment
+        method</a> are handled by <dfn data-lt=
+        "payment app|payment apps">payment apps</dfn>. In this specification,
+        these details are left up to the <a>user agent</a>, but future
+        specifications may expand on the processing model in more detail.
       </p>
       <p>
         In addition to better, more consistent user experiences, this also
@@ -276,13 +283,15 @@
         </h2>
         <p>
           The <a>PaymentRequest</a> is constructed using the supplied
-          <code>methodData</code> list including any <a>payment method</a>
-          specific <code>data</code>, the payment <code>details</code>, and the
-          payment <code>options</code>.
+          <var>methodData</var> list including any <a>payment method</a>
+          specific <a data-lt="PaymentMethodData.data">data</a>, the payment
+          <var>details</var>, and the payment <var>options</var>. The
+          <var>methodData</var> supplied to the <a>PaymentRequest</a>
+          constructor SHOULD be in the order of preference of the caller.
         </p>
         <div class="note">
           <p>
-            The <code>methodData</code> sequence contains
+            The <var>methodData</var> sequence contains
             <a>PaymentMethodData</a> dictionaries containing the <a>payment
             method identifiers</a> for the <a>payment methods</a> that the web
             site accepts and any associated <a>payment method</a> specific
@@ -305,7 +314,7 @@
             const request = new PaymentRequest(methodData, details, options);
           </pre>
           <p>
-            The <code>details</code> object contains information about the
+            The <var>details</var> object contains information about the
             transaction that the user is being asked to complete such as the
             line items in an order.
           </p>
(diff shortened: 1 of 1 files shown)


adrianba:
I think this is generally fine but I'd prefer that we don't _require_ data be passed in JSON string form. In our implementation we might present the data to apps already "deserialized" so it won't look like a string to them. I think the spec ought not be so explicit in saying it will be that string form - an equivalent should also be okay. We likely to use an alternative RPC mechanism that isn't a string and this should be acceptable.

domenic:
Well, in general,

> Conformance requirements phrased as algorithms or specific steps may be implemented in any manner, so long as the end result is equivalent.

The important part is that JSON.stringify is performed at a very particular point in the algorithm.

adrianba:
> Well, in general,
> > Conformance requirements phrased as algorithms or specific steps may be implemented in any manner, so long as the end result is equivalent.

>The important part is that JSON.stringify is performed at a very particular point in the algorithm.

That's true, but it looks like the end result here is passing off a JSON string to a payment app.

domenic:
How would that be observably any different from passing along some other serialization format derived from or containing the equivalent information to that JSON string?

adrianba:
My concern is that the way this is written implies that that Payment App spec will expect a JSON string.

domenic:
Yes, I think it would. It can then transform that JSON string into whatever format is most appropriate for it.

What other serializable data structure would you expect a spec to be able to deal in?

adrianba:
I think the transport format should be an implementation detail as long as it isn't exposed. The current language leaves ambiguous whether it would be exposed. I would like to close down that ambiguity.

We discussed in the editors call just now and Marcos is reviewing.

domenic:
Can you explain how it's ambiguous that it would be exposed? No spec exposes it except to other specs.

We need a well-defined transport format for specs to be able to talk to each other. Specifying an entirely new one seems superfluous.

index.html
```diff
@@ -125,13 +125,20 @@
...
+        passed between the <a>payment method</a> and the merchant via the
         <a>user agent</a> with the buyer confirming and authorizing as
         necessary across the flow.
       </p>
       <p>
+        The details of how to fulfill a payment request for a given <a>payment
+        method</a> are handled by <dfn data-lt=
+        "payment app|payment apps">payment apps</dfn>. In this specification,
+        these details are left up to the <a>user agent</a>, but future
+        specifications may expand on the processing model in more detail.
```
marcoscaceres, line 139: I think this is fine for now - but we should have at least 1 base thing here (probably "basic-card").

marcoscaceres (approved):

marcoscaceres:
Small conflicts

index.html
domenic: I'm not sure I completely understand the ask here, but maybe this is the same as my "If this looks correct, then to clarify how the basic card spec actually interacts with the main spec, we'll edit the basic card spec to say:" from the PR's OP?

domenic:
Rebased with conflicts fixed.

adrianba (approved):

PR #396, opened 2017-01-17
feat(payment resp algo): phone# SHOULD be E.164 (closes #390)

### LINKED ISSUE #390: Phone number format
What format should we use for phone numbers for `payerPhone` and `PayerAddress.phone`? I propose E.164, which looks like this: `"+15555555555"`. (For UI, Chrome will probably use E.123, which shows that same phone number in more human-readable format of `"+1 (555) 555-5555"`.)

marcoscaceres:


marcoscaceres:
@rsolomakhin r?

rsolomakhin:
lgtm

PR #398, opened 2017-01-18
Editorial: refer to methods with parentheses (closes #378)

### LINKED ISSUE #378: Editorial: refer to methods with parentheses for clarity, and spell out their params in their definition
E.g. instead of

> abort can only be called after the show method has been called

say

> abort() can only be called after the show() method has been called

and instead of

> The complete method MUST act as follows: 

say

> The complete(result) method MUST act as follows:

marcoscaceres:


marcoscaceres:
@domenic r?

domenic (changes requested):
Nice.

The intro paragraph for abort is not updated; it refers to abort and show without parens.

Also, I would remove the mention of "result" before the actual method definition in the definition of complete.

marcoscaceres:
> Also, I would remove the mention of "result" before the actual method definition in the definition of complete.

@domenic, I agree, but to clarify. I think it might be best to do the following here:

  * remove the paragraph that starts with "The complete method takes a string argument from the PaymentComplete enum (result)." 
  * split out `enum PaymentComplete` into its own section.

domenic:
Yeah, that sounds pretty good to me!

marcoscaceres:
Ok, I think this is good to go too.

index.html
```diff
@@ -1492,13 +1490,44 @@ <h2 id="state-transitions" class="informative">
...
+          "unknown",
+        };
+      </pre>
+      <dl>
+        <dt>
+          "<dfn>fail</dfn>"
+        </dt>
+        <dd>
+          Indicates that processing of the payment failed. The <a>user
+          agent</a> MAY display UI indicating failure.
```
domenic, line 1510: It's a bit weird that these normative processing model requirements are given in the definition of the enum, instead of in the algorithm that processes it, but that's a separate issue.

index.html
```diff
@@ -1492,13 +1490,44 @@ <h2 id="state-transitions" class="informative">
         </dd>
       </dl>
     </section>
+    <section data-dfn-for="PaymentComplete" data-link-for="PaymentComplete">
+      <h2><dfn>PaymentComplete</dfn> enum</h2>
```
domenic: Maybe add the missing "enum" to the PaymentShippingType header while you're here

domenic (commented):
Oh, I caught one more missing one. updateWith!

I guess with that in mind we now have competing conventions of using `<var>` (updateWith) vs. `<code>` (complete) for argument names, inside the method algorithms. Either seems fine but we should stick with one...

index.html
marcoscaceres: Agree.

index.html
marcoscaceres: fixed.

marcoscaceres:
... just fixing a few more `updateWith()`s ... well spotted!

marcoscaceres:
@domenic, btw... does this seem odd to you too?:

```
void updateWith(Promise<PaymentDetails> d);
```

That should probably be:

```
void updateWith(Promise<PaymentDetails> d or PaymentDetails d);
```

Not sure why it needs details needs to always be wrapped in a promise.

marcoscaceres:
ok, think we are good to go here too.

index.html
```diff
@@ -1802,10 +1804,10 @@ <h2 id="state-transitions" class="informative">
         </p>
         <section>
           <h2>
-            <dfn>updateWith()</dfn> method
+            <dfn data-lt="updateWith(details)">updateWith()</dfn> method
           </h2>
           <p>
-            The <a>updateWith</a> method MUST act as follows:
+            The <a>updateWith(details)</a> method MUST act as follows:
```
domenic: Now it is inconsistent as the rest of the algorithm references `<var>d</var>`. This also doesn't address the `<var>`-in-updateWith vs. `<code>`-in-complete issue.

domenic (changes requested):
> btw... does this seem odd to you too?:

No, you cannot overload with a promise argument in Web IDL, because whatever the developer passes in gets converted to a promise.

index.html
marcoscaceres: Ah, darn. I missed the <var>d</var> and code stuff. BRB!

marcoscaceres: (also, I prefer `<var>details</var>` ... more clear)

marcoscaceres:
I keep forgetting about the magic promise wrapping - that's super useful.

index.html
marcoscaceres: Argh... there is a lot of other `<code>` to fix...

marcoscaceres:
Ok ... third (fifth?) try is a charm! ✨

index.html
marcoscaceres: I s/code/var in complete()...

domenic:
So now there are two variables named details in the algorithm... see "Let details be the result of converting value to a PaymentDetails dictionary." Getting close!?

marcoscaceres:
effin hell... brb :)

marcoscaceres:
Ok, I changed it to "detailsPromise" - I think it makes it much nicer to read in prose. 

For example: "The remaining steps are conditional on the <var>detailsPromise</var> settling"

domenic (approved):

domenic:
Probably more feedback then you bargained for, but I think overall this is a great improvement; thanks for sticking with it!!

marcoscaceres:
Np! thank you for catching all those things.

PR #404, opened 2017-01-20
State how payment details modifier data is serialized and used

### LINKED ISSUE #307: The myth of JSON-serializable object

### LINKED ISSUE #346: The "data" member of  PaymentDetailsModifier doesn't seem to be used for anything

domenic:
This closes #346, by making it clear that the data is serialized and stored in the PaymentRequest for further usage by show(), and closes #307, by finally stating all the points at which JSON-serialization happens explicitly.

A couple of points worth mentioning:

- I assumed we wanted the same swallow-all-exceptions behavior for updateWith for the JSON serialization process there.
- It's still a bit vague exactly what data is to be sent to the payment app by show(). I put in a sentence "The payment app should be sent the appropriate data from request in order to guide the user through the payment process. This includes the various attributes and internal slots of request." This is meant to convey how the data is transported. But getting this fleshed out in more detail would probably be a good idea.

### DIFF
--- a/index.html
+++ b/index.html
@@ -474,6 +474,8 @@
               </li>
             </ol>
           </li>
+          <li>Let <var>serializedModifierData</var> be an empty list.
+          </li>
           <li data-link-for="PaymentDetails">Process payment details modifiers:
             <ol>
               <li>Let <var>modifiers</var> be an empty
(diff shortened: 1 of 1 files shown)


marcoscaceres:
> I assumed we wanted the same swallow-all-exceptions behavior for updateWith for the JSON serialization process there.

Yes, that seems fine. 

> It's still a bit vague exactly what data is to be sent to the payment app by show(). I put in a sentence "The payment app should be sent the appropriate data from request in order to guide the user through the payment process. This includes the various attributes and internal slots of request." This is meant to convey how the data is transported. But getting this fleshed out in more detail would probably be a good idea.

I think this is pretty good, and we should flesh it out if ambiguities comes up during implementation, testing, and usage. It maybe "good enough"(tm) to not cause interop issues.

index.html
```diff
@@ -834,13 +859,30 @@ <h2 id="state-transitions" class="informative">
...
               The current <a>PaymentDetails</a> for the payment request
               initially supplied to the constructor and then updated with calls
               to <a data-lt=
-              "PaymentRequestUpdateEvent.updateWith">updateWith()</a>.
+              "PaymentRequestUpdateEvent.updateWith">updateWith()</a>. Note
+              that all <a data-lt="PaymentDetailsModifier.data">data</a>
+              members of <a>PaymentDetailsModifier</a> instances contained in
+              the <a data-lt="PaymentDetails.modifiers">modifiers</a> member
+              will be removed, as they are instead stored in serialized form in
+              the [[\serializedModifierData]] internal slot.
```
marcoscaceres: nit: link to `<a>[[\serializedModifierData]]</a>`. I'll eventually teach ReSpec about internal slots :(

marcoscaceres (approved):
LGTM, bar nit.

PR #405, opened 2017-01-20
Queue tasks to change developer-visible properties

### LINKED ISSUE #339: Setting internal slots of objects from asynchronous steps is not OK
https://w3c.github.io/browser-payment-api/#abort-method step 7 says:

> Set the value of the internal slot request@[[state]] to closed. 

But this is running async.  That means its only interaction with observable objects should be via tasks queued on a task queue, to avoid violating run-to-completion semantics.

### LINKED ISSUE #340: Is it intended that abort() rejects the acceptPromise and resolves its return value in separate tasks?
https://w3c.github.io/browser-payment-api/#abort-method steps 8 and 9 are running async, so those two steps will queue two separate tasks to perform the actual promise manipulation.  This is observably different from performing both operations in a single task.

If this is _is_ the intended behavior, please make sure you have a test for this in the test suite, because I expect implementors will miss the subtlety here.

### LINKED ISSUE #351: The shipping address/option changed algorithms should not be synchronously poking internal slots
As far as I can tell these algorithms are invoked by the browser UI while things are running in parallel.  That is, they do not seem to run off a task.  It's a bit hard to say, because nowhere in the spec does it say how the browser goes about invoking these algorithms.

Anyway, assuming these are running in parallel, they should not be modifying values of attributes on objects, because that breaks run-to-completion.  They should be doing the value modification off a task; probably the same one that fires the relevant change event.

domenic:
Fixes #339. Fixes #351. Fixes #340.

### DIFF
--- a/index.html
+++ b/index.html
@@ -692,17 +692,24 @@
           <li>Try to abort the current user interaction and close down any
           remaining user interface.
           </li>
-          <li>If it is not possible to abort the current user interaction, then
-          reject <var>promise</var> with "<a>InvalidStateError</a>"
-          <a>DOMException</a> and abort this algorithm.
-          </li>
-          <li>Set the value of the internal slot
-          <var>request</var>.<a>[[\state]]</a> to <i>closed</i>.
-          </li>
-          <li>Reject the promise <var>request</var>.<a>[[\acceptPromise]]</a>
-          with an "<a>AbortError</a>" <a>DOMException</a>.
-          </li>
-          <li>Resolve <var>promise</var> with undefined.
+          <li>
+            <a>Queue a task</a> on the <a>user interaction task source</a> to
+            perform the following steps:
+            <ol>
+              <li>If it is not possible to abort the current user interaction,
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2319,6 +2338,9 @@ <h2 id="state-transitions" class="informative">
               <dfn>queue a task</dfn>
             </li>
             <li>
+              <dfn>user interaction task source</dfn>
```
marcoscaceres, line 2341: nit: `<dfn data-lt="!HTML5#user-interaction-task-source">`

marcoscaceres (approved):

domenic:
Generally not a fan of linking to that spec with it's outdated/inaccurate copies-of-definitions.

marcoscaceres:
I also find it unpalatable. Maybe we can cite the canonical one and note that we will reference the other one if it becomes accurate. We are indeed doing everyone a disservice by referencing the outdated spec (as was also the case with WebIDL).

marcoscaceres:
Ok, so we don't get bogged down, I'll fix the citations separately.

PR #406, opened 2017-01-20
Move state transitions diagram into a note

### LINKED ISSUE #342: Is section 3.4 informative or normative?
Section 3.4 doesn't say whether it's informative or normative.  That means that presumably it's normative, but it's not clear to me what the normative content of it is....

I _assume_ this section is meant to be informative and should be annotated as such.

domenic:
Closes #342.

### DIFF
--- a/index.html
+++ b/index.html
@@ -766,17 +766,6 @@
           </li>
         </ol>
       </section>
-      <section>
-        <h2 id="state-transitions" class="informative">
-          State transitions
-        </h2>
-        <p>
-          The internal slot <a>[[\state]]</a> follows the following state
-          transitions:
-        </p><img alt=
-        "Transition diagram for internal slot state of a PaymentRequest object"
-        src="images/state-transitions.svg" width="518" height="125">
-      </section>
       <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
         <h2>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -881,6 +870,14 @@ <h2 id="state-transitions" class="informative">
             </td>
           </tr>
         </table>
+        <div class="note" id="state-transitions">
+          <p>
+            The internal slot <a>[[\state]]</a> transitions between its various
+            potential values as follows:
+          </p><img alt=
```
marcoscaceres, line 877: Why not use a figure/figcaption here? All the alt info seems valuable to everyone.

marcoscaceres (changes requested):
@domenic, let me know if you want me to do this (add figure/figcaption). I don't mind.

index.html
domenic: It's strictly redundant with the image, describing the exact same information.

domenic: If you think we should repeat it, then we need to do `alt=""` since then the image is redundant.

marcoscaceres: yeah, the alt is redundant... don't want to repeat it. Sorry for not being clear.

marcoscaceres (approved):
I'll add fig caption if/when I do another pass of at fixing the markup.

Issue #27, opened 2016-03-09
Should API support billing address capture (for tax computation)?


ianbjacobs:
During the Feb FTF meeting [1], Vincent Kuntz pointed out that in Europe, billing address is required to compute VAT. The use case is that you purchase something in one country (where VAT is applied) but it is being shipped to another. 

Thus, for some tax computations, do we need billing address (or some portion of it, like "country")?

[1] https://www.w3.org/2016/02/24-wpwg-minutes

djackso3:
The method by which many payment systems work (especially in card) is to verify against billing address so billing address is required for some payment types and providers. In addition, some items will require the KYC address (government functions like BMV transactions). Delivery address is not always interesting to the payments process itself, but could be used if there is an AML issue with respect to shipping location of product that requires filtering by the payments provider. Lastly, a growing number of consumers provide KYC address -- because it is legally mandated -- and billing address for convenience -- yet have a fluctuating shipping address which generally in not germane to the payments process -- however, could be for filtering requirements as mentioned above. There are excellent examples of payment systems which cease working because of confusion of KYC / billing address having a mismatch against shipping address at the merchant. Personally over the last 6 months I have been working with Bank of America on this exact issue. The bank made an IT change which confused KYC and billing address. Once that happened, Delta Airlines refused to accept payment for a ticket because of the mismatch issue. My "delivery" address at the merchant (Delta) was equal to the billing address. So when I changed to the KYC address for approval of the payment, Delta refused again because they had a merchant mismatch of their delivery address of record with the physical (KYC) address I attempted to use. For this reason -- until finally adjusted by BofA (after first flat refusing the issue) -- I could not use my card at Delta because I could not figure out what address they believed they were verifying against. It took months for me to get them to believe it an issue and they claimed that KYC = billing address (which is of course false). I explain in more detail because the issue is with respect to lack of consistent definition. There is another further example of a bank confusing "mailing" address and "billing" address. In short, the bank verified against "billing" address but their database did not equate mailing=billing so the bank rejected the verification because no "billing" address was on file and there was a mismatch against KYC.

I believe that there is reason that all three -- with consistent definition -- may need to be part of the protocol because of the fluidity for people with various reasons to maintain a shipment location for goods and "letter" or USPS location for "billing". And both of these are not the KYC address because the merchant never has a reason to collect the actual KYC address.  Help?

djackso3:
[...]

PR #409, opened 2017-01-26
Add billing shipping type

rvm4:
There are cases (such as with digital goods), where an address is needed, but there is no notion of shipping. I would prefer to be more bold in this edit by changing the name of this enum to PaymentAddressType and changing requestShipping to requestAddress, but I wanted to be on the safe side to start. Please let me know if the renaming would be agreeable as well.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1210,7 +1210,8 @@
         enum PaymentShippingType {
           "shipping",
           "delivery",
-          "pickup"
+          "pickup",
+          "billing",
         };
       </pre>
       <dl>
--- a/index.html
+++ b/index.html
@@ -1211,7 +1211,7 @@
           "shipping",
           "delivery",
           "pickup",
-          "billing",
+          "billing"
         };
       </pre>
       <dl>


index.html
```diff
@@ -1210,7 +1210,8 @@ <h2 id="state-transitions" class="informative">
         enum PaymentShippingType {
           "shipping",
           "delivery",
-          "pickup"
+          "pickup",
+          "billing",
```
domenic: Invalid Web IDL (trailing commas)

rvm4: I will address that once I get some quick signal around the naming.

rsolomakhin:
Billing address requirement depends on payment methods. So, bitcoin, for example, would not need it, but credit cards do need it. Therefore, I'd prefer that we keep billing address in method-specific data. I suggest that you send a patch to "Basic Card" spec, which has billing address in https://w3c.github.io/webpayments-methods-card/#basiccardresponse. I'm in support of something like this:
```javascript
new PaymentRequest([{
    supportedMethods: ['basic-card'],
    data: {requiredFields: ['billingAddress']}
  }], shoppingCart);
```

rvm4:
I disagree with your premise, the merchant will often need an address in order to compute the taxes owed. If I accept your premise, I don't understand how this would work. When would the browser surface an address selection?

ianbjacobs:
I believe we discussed billing address in issue 27:
 https://github.com/w3c/browser-payment-api/issues/27

(which may not have concluded cleanly)

Ian

rsolomakhin:
I would prefer that merchants estimate the tax and indicate that to the user by placing a `pending: true` on the tax in `displayItems`. The merchant is also able to confirm the updated price after they receive the `PaymentResponse`.

If we must calculate the tax precisely during the PaymentRequest flow, user agent needs to provide billing address to the website before the user clicks "Pay" in browser UI. Is the country enough for tax purposes? A user agent generally should require a user action for this to happen in a privacy-preserving manner. Therefore, we need:

1. New option `"requestBillingAddress"` (or `"requestBillingCountry"`). This is similar to existing option `"requestShipping"`.
1. New user interface in implementing user agents to prompt the user to send their billing address or country to the website.
1. A new event on `PaymentRequest` object: `"billingAddressChange"` or `"billingAddressChange"`, for example. This would work similarly to `"shippingAddressChange"`.

This is a fairly involved addition to the API. Therefore, I prefer the estimation workaround. How strong is the preference for this?

marcoscaceres:
> On 27 Jan 2017, at 4:53 am, Rouslan Solomakhin <notifications@github.com> wrote:
> 
> Billing address requirement depends on payment methods. So, bitcoin, for example, would not need it, but credit cards do need it. 
> 

But only in some countries (Australia doesn't require a billing address to process a credit card payment online, afaik).

dlongley:
In the US (for states with sales taxes that are destination-based), the tax applies to the shipping address not the billing address. That would apply even for digital goods; the tax is calculated based on where the customer "takes possession" of the product, not whatever their billing address is. This would seem to mean that a "shipping address" should be gathered for digital goods that need to be taxed, not a "billing address", at least for tax purposes. Taxes are fun.

dlongley:
So, making this a shipping type as @rvm4 suggested probably makes sense, however, perhaps the name should be something like "tax" or "taxation" or "usage" rather than "billing" so as to avoid confusing it with a "billing address" which has to do with payment methods, as @rsolomakhin noted.

rvm4:
The canonical example here is for digital goods. Shipping address doesn't quite fit because nothing is shipping.

Why would you have to send a merchant the address prior to Pay being clicked? You could be given a list of items that does not include tax, but specify in the options that I need a billing/taxation address and when that event is received, append another item to the list that captures the applicable tax.

zkoch:
I think this is actually one of our oldest issues, see #27.

We didn't have a good solution to this for a while, but I agree that this is probably the right place to build this in. There are a few things to consider, though:

* VAT rules vary per country. Some are billing, some are where the good is being shipped, and others are the home address of the purchaser, which is independent of where the good is being shipped or the billing address used. Fun.

* As such, perhaps just calling this "vat" or something similar might be simpler. It gives some flexibility to implementors about how they want to handle the tricky situation.

* We would need to spec out a behavior change here. Presumably a type of "vat" doesn't require shipping options, so that should probably be spec'd out.

ianbjacobs:
@zkoch,

Could the editors discuss this at their next meeting and suggest something on this thread?

Ian

zkoch:
I think our suggestion is similar to what I mentioned above:

* We should figure out if "billingAddress" is the right label for this, or if we should make more generic as "vat"

* We need to spec out the behavior change, since this probably won't require shipping options

rvm4:
Do we need to hold up this PR for figuring out the label? The enum name is immaterial to what gets rendered in the browser, is it not? Let's hash this out in the F2F if need be.

nschonni:
Sent from my BlackBerry 10 smartphone.                                                                                                                                                                                                                From: rvm4Sent: Tuesday, February 21, 2017 9:35 PMTo: w3c/browser-payment-apiReply To: w3c/browser-payment-apiCc: SubscribedSubject: Re: [w3c/browser-payment-api] Add billing shipping type (#409)Do we need to hold up this PR for figuring out the label? The enum name is immaterial to what gets rendered in the browser, is it not?

—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/w3c/browser-payment-api","title":"w3c/browser-payment-api","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/w3c/browser-payment-api"}},"updates":{"snippets":[{"icon":"PERSON","message":"@rvm4 in #409: Do we need to hold up this PR for figuring out the label? The enum name is immaterial to what gets rendered in the browser, is it not?"}],"action":{"name":"View Pull Request","url":"https://github.com/w3c/browser-payment-api/pull/409#issuecomment-281550363"}}}

zkoch:
I don't think we can merge until we clarify what happens with shipping options when this is selected.

stan-stripe:
Remarks from the 03/02 call:
- User agent being able to properly differentiate and store billing addresses from shipping addresses is important to the user experience, but I'm a bit worried that this PR incentivize user agents to not do that
- From a developer experience perspective, using `billing_address` feels really wonky, if the goal really is vat calculation for digital goods, I think we should call it `vat_address` instead of `billing_address`. The goal is also to clearly separate billing address tied to payment instrument and the address that have influence on the total amount.

ianbjacobs:
We discussed on today's WPWG call:
 https://www.w3.org/2017/03/02-wpwg-minutes#item03

The group supports support for VAT info collection in v1 of the spec. We ask the
editors to evaluate how to do so and the cost of doing so.

Thanks!

Ian

zkoch:
Closing this to continue discussion in #27. We've agreed we would like this, but needs a bit more work to figure out interactions with shippingOptions, etc

PR #418, opened 2017-02-09
Make updateWith() return a promise signaling any update failures

### LINKED ISSUE #350: updateWith will silently ignore data in hard-to-debug ways
There are various cases in which updateWith will ignore the data its argument promise is resolved with.  For example, it will ignore various negative amounts, onvalid decimal monetary values, etc.

When this happens, there is no indication to the caller of updateWith that anything went wrong.  Should there be (e.g. via a Promise that updateWith returns)?  I assume UAs are free to report things to the console here, of course, but that only helps if the page author is reproducing the problem themselves...

domenic:
This closes #350. It is a slightly backward-incompatible change in that consumers which were previously try-catching errors will no longer see them. But consumers that were always passing valid data, or passing invalid data and not try-catching the errors, will see no changes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1818,7 +1818,7 @@
         <pre class="idl">
           [Constructor(DOMString type, optional PaymentRequestUpdateEventInit eventInitDict),SecureContext]
           interface PaymentRequestUpdateEvent : Event {
-            void updateWith(Promise&lt;PaymentDetails&gt; detailsPromise);
+            Promise&lt;void&gt; updateWith(Promise&lt;PaymentDetails&gt; detailsPromise);
           };
         </pre>
         <p>
@@ -1830,6 +1830,15 @@
           call <a>updateWith()</a> and provide a promise that will resolve with
           a <a>PaymentDetails</a> dictionary containing changed values that the
           <a>user agent</a> SHOULD present to the user.
+        </p>
+        <p>
+          The promise returned by <a>updateWith()</a> will be fulfilled with
+          undefined if all of the data is accepted, or rejected with a
+          <a>TypeError</a> if the update was complete but some data was
+          rejected as invalid. It can also be rejected if it is called
+          inappropriately, e.g. on a <a>PaymentRequestUpdateEvent</a> event
+          that is not being dispatched or is has already been updated, or if
+          the passed promise rejects.
         </p>
         <p>
           The <a>PaymentRequestUpdateEvent</a> constructor MUST set the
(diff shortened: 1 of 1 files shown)


marcoscaceres:
This is great, @domenic! I've pinged folks from Microsoft to see if they can live with this change.

index.html
```diff
@@ -1832,6 +1832,15 @@
           <a>user agent</a> SHOULD present to the user.
         </p>
         <p>
+          The promise returned by <a>updateWith()</a> will be fulfilled with
+          undefined if all of the data is accepted, or rejected with a
+          <a>TypeError</a> if the update was complete but some data was
+          rejected as invalid. It can also be rejected if it is called
+          inappropriately, e.g. on a <a>PaymentRequestUpdateEvent</a> event
+          that is not being dispatched or is has already been updated, or if
```
marcoscaceres: typo: is has

marcoscaceres (approved):

zcorpan:
Would be good to update web-platform-tests for this.

marcoscaceres:
we are just waiting for one more OK from Microsoft. But should be all good.

mollyd16:
@marcoscaceres @zkoch 
I’ve been working through this scenario and it's fairly complicated and additionally also seems like a pretty narrow edge case…

Here’s what I’ve been thinking through:

1.	The feedback provided by the rejection of the promise returned from updateWith() does not seem to provide actionable feedback because what will a merchant do with the feedback? If it was an error that should be shared with the user, how would the merchant do so? And if it is from malformed merchant generated data (bug/exploit), how would the information benefit the merchant?

2.	What does this additional promise offer that’s better than failing the transaction? 

3.	If a merchant has malformed data via a bug/exploit, shouldn’t we just veer on the safe side and fail the request anyways. 

4.	Just generally, this seems unlikely that a merchant will have malformed information, and a large code change doesn’t seem hugely beneficial

5.	If we are unable to make this change in the short-term, then Microsoft Edge will throw an exception if .then() or .catch() is called after updateWith()


We would prefer to keep the implementation as is and think that having invalid data reject the entire transaction is a good solution to the problem

marcoscaceres:
@domenic - you are probably in the best position to answer @molly26dalton's questions above.

domenic:
I think @zkoch is probably the best person to answer from Google's side actually :). Failing makes sense to me, but I have to assume there was some rationale behind the original design to ignore invalid data...

> We would prefer to keep the implementation as is and think that having invalid data reject the entire transaction is a good solution to the problem

I'm confused: do you want to keep the implementation as-is, or do you want to have invalid data reject the transaction? Those are mutually exclusive.

rsolomakhin:
By the way, Chrome does not ignore invalid data in `updateWith()`. It validates the provided `PaymentDetails` in the same way as the constructor for `PaymentRequest`. If `PaymentDetails` is invalid, then `updateWith()` throws the same exceptions as the constructor. The only difference is that the constructor prohibits use of `PaymentDetails.error`, which is OK to use in `updateWith()`.

If validation of `PaymentDetails` is the only concern for `updateWith()`, should we stick with exceptions, because that validation can be performed synchronously.

domenic:
Oh, that is great to hear! OK, so @marcoscaceres and @molly26dalton can you confirm Mozilla and Microsoft would be OK with switching the spec from ignoring invalid data to throwing an exception? Since that apparently matches Chrome, and would solve the original problem, that seems like a great solution, and I'm happy to update this PR to do so once I get confirmation.

rsolomakhin:
Oh no, spoke too soon. After double-checking [the code](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/modules/payments/PaymentRequest.cpp?l=742-775), I notice that invalid data in `updateWith()` will reject the `show()` promise. I can change it to throw instead, if that's more desirable. Sorry!

domenic:
I realized that we can't throw synchronously, because updateWith takes a promise for PaymentDetails, instead of taking a PaymentDetails directly. So we won't have the data available synchronously, in time to throw.

We could still reject the show() promise though as the way of signaling errors. The current spec will ignore any bad data, performing a partial update for any good data that is present, and then always fulfill the show() promise. We could change that to reject the show() promise (and not do any update) if any of the data is bad.

mollyd16:
@domenic LGTM

domenic:
I started working on this but realized it's going to conflict with #425 so let's get that merged first.

marcoscaceres:
#425 no longer blocks this 🎉

domenic:
Superceded by https://github.com/w3c/browser-payment-api/pull/451 which implements the semantics discussed here of validating the data ahead of time.

PR #419, opened 2017-02-13
BREAKING CHANGE: s/paymentRequestID/paymentRequestId (closes #414)

### LINKED ISSUE #414: paymentRequestID is spelled wrong :(
Per https://w3ctag.github.io/design-principles/#casing-rules it should be "paymentRequestId" (lowercase).

I assume this is too late to change now, unless paymentRequestID/d is one of the things that are in the spec but nobody has shipped yet? If it is indeed too late, we should add a note to the spec saying essentially "sorry for creating an API inconsistent with the rest of the platform".

marcoscaceres:
Implementers agreed that this is ok to change, as none of us shipped it. The naming aligns with other things in the platform.

### DIFF
--- a/index.html
+++ b/index.html
@@ -213,7 +213,7 @@
           Promise&lt;void&gt; abort();
           Promise&lt;boolean&gt; canMakePayment();
 
-          readonly attribute DOMString? paymentRequestID;
+          readonly attribute DOMString? paymentRequestId;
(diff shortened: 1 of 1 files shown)


zcorpan:
Test for this and other earlier IDL changes:
https://github.com/w3c/web-platform-tests/pull/4806

domenic (approved):

PR #420, opened 2017-02-14
Define payment option filtering algorithm.

rsolomakhin:
Define the algorithm for filtering the payment options in the payment
app. These payment options are used to invoke the payment app that the
user selects.

### DIFF
--- a/index.html
+++ b/index.html
@@ -628,30 +628,21 @@
           <li>Return <var>acceptPromise</var> and perform the remaining steps
           <a>in parallel</a>.
           </li>
-          <li>For each <var>paymentMethod</var> in
-          <var>request</var>.<a>[[\serializedMethodData]]</a>:
-            <ol>
-              <li>Consult the appropriate <a>payment apps</a> to see if they
-              support any of the <a>payment method identifiers</a> given by the
-              first element of the <var>paymentMethod</var> tuple, passing
-              along the data string given by the second element of the tuple in
-              order to help them determine their support.
-              </li>
-            </ol>
+          <li>Let <var>matchingOptions</var> be the result of running the
+            <a>filter payment options algorithm</a>.
           </li>
-          <li>If this consultation produced no supported method of paying, then
-          reject <var>acceptPromise</var> with a "<a>NotSupportedError</a>" <a>
-            DOMException</a>, and abort this algorithm.
-          </li>
+          <li>If <var>matchingOptions</var> is empty, then reject
+            <var>acceptPromise</var> with a "<a>NotSupportedError</a>" <a>
+              DOMException</a>, and abort this algorithm. </li>
           <li>
             <p>
               Otherwise, show a user interface to allow the user to interact
-              with the payment request process, using those <a>payment apps</a>
-              and <a>payment methods</a> which the above step identified as
-              feasible. The user agent MAY show payment methods in the order
-              given by <var>supportedMethods</var>, but SHOULD prioritize the
-              preference of the user when presenting payment methods and
-              applications.
+              with the payment request process, using those <a>payment apps</a>,
+              <var>matchingOptions</var>, and <a>payment methods</a> which the
+              above step identified as feasible. The user agent MAY show payment
+              methods in the order given by <var>supportedMethods</var>, but
+              SHOULD prioritize the preference of the user when presenting
+              payment methods and applications.
             </p>
             <p>
               The <a>payment app</a> should be sent the appropriate data from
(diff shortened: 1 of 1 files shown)


rsolomakhin:
@adamroach: This is essentially your algorithm, but without the addition of the new `filters` member. Does it fairly represent your idea?

@domenic: This pull request is especially tricky because it intends to interact with the payment app spec. Please take a look.

index.html
```diff
@@ -2342,6 +2333,50 @@
...
+        </p>
+        <ol>
+          <li>Let <var>matchingOptions</var> be an empty list.</li>
+          <li>Let <var>request</var> be the <a>PaymentRequest</a> object that
+          the user is interacting with.
+          <li>For each <var>requestMethod</var> in
+            <var>request</var>.<a>[[\serializedMethodData]]</a>:
+            <ol type="I">
+              <li>Let <var>filters</var> be <var>requestMethod.data</var>
+                members whose values are lists of strings.</li>
```
adamroach, line 2354: Hrm. I thought, after our last discussion on this point, we had reached agreement that this was problematic, in that it precluded payment methods from defining `data` parameters whose values were a list of strings, but which were *not* filters.

index.html
```diff
@@ -2342,6 +2333,50 @@
...
+              <li>For each <var>option</var> of the installed
+                <a>payment apps</a>:
+                <ol type="i">
+                  <li>Let <var>optionMatch</var> be true.</li>
+                  <li>For <var>filterName</var> in the keys of
+                    <var>requestMethod</var>:
```
adamroach, line 2360: Shouldn't this be in the keys of `filters` rather than keys of `requestMethod`?

index.html
```diff
@@ -628,30 +628,21 @@
           <li>Return <var>acceptPromise</var> and perform the remaining steps
           <a>in parallel</a>.
           </li>
-          <li>For each <var>paymentMethod</var> in
```
adamroach, line 631: You're removing the part of the current document that deals with `paymentMethod` matching. The algorithm I thumbnailed in [issue 96](https://github.com/w3c/webpayments-payment-apps-api/issues/96) was intended to supplement `paymentMethod` matching rather than replace it. Basically, you need to put the *filter payments options algorithm* into a loop that evaluates it once for each `paymentMethod` in the request, and only those options that match the `paymentMethod` should be evaluated on each iteration.

adamroach (changes requested):
Thanks! The text looks mostly good, but I think you cut a bit too deep. I'd also like to get on the same page about how to indicate filters in the PaymentRequest object.

index.html
```diff
@@ -628,30 +628,21 @@
...
-              first element of the <var>paymentMethod</var> tuple, passing
-              along the data string given by the second element of the tuple in
-              order to help them determine their support.
-              </li>
-            </ol>
-          </li>
-          <li>If this consultation produced no supported method of paying, then
-          reject <var>acceptPromise</var> with a "<a>NotSupportedError</a>" <a>
-            DOMException</a>, and abort this algorithm.
+          <li>Let <var>matchingOptions</var> be the result of running the
```
domenic, line 631: It's a little unclear to me what the result of this algorithm is, i.e. what a "matching option" is. The algorithm steps seem to only add _option_ to the return value, and _option_ is a payment app. So is _matchingOptions_ a list of payment apps? But then the subsequent sentence

> show a user interface to allow the user to interact with the payment request process, using those payment apps, matchingOptions, and payment methods which the above step identified as feasible

is confusing, since the above step only identified payment apps = _matchingOptions_, whereas this seems to be saying that there are three different lists determined by the above step: the payment apps, _matchingOptions_, and payment methods.

index.html
```diff
@@ -2342,6 +2333,50 @@
...
+          following steps:
+        </p>
+        <ol>
+          <li>Let <var>matchingOptions</var> be an empty list.</li>
+          <li>Let <var>request</var> be the <a>PaymentRequest</a> object that
+          the user is interacting with.
+          <li>For each <var>requestMethod</var> in
+            <var>request</var>.<a>[[\serializedMethodData]]</a>:
+            <ol type="I">
+              <li>Let <var>filters</var> be <var>requestMethod.data</var>
```
domenic, line 2353: This step doesn't make sense to me. requestMethod is a tuple containing supportedMethods, a `sequence<DOMString>`, and `data`, a string of serialized JSON. So `requestMethod.data` does not have any members.

index.html
```diff
@@ -2342,6 +2333,50 @@
...
+                members whose values are lists of strings.</li>
+              <li>For each <var>option</var> of the installed
+                <a>payment apps</a>:
+                <ol type="i">
+                  <li>Let <var>optionMatch</var> be true.</li>
+                  <li>For <var>filterName</var> in the keys of
+                    <var>requestMethod</var>:
+                    <ol type="a">
+                      <li>If <var>option.capabilities</var> does not have key
```
domenic, line 2362: Where is option.capabilities defined? I thought a payment app was an abstract concept, not an object which had properties.

domenic (commented):
I'm rather confused, so I tried to ask clarifying questions...

ianbjacobs:
We discussed the proposal today at the WPWG call:
 https://www.w3.org/2017/02/23-wpwg-minutes#item02

Rouslan is planning to make some updates regarding (1) explicit identification of filters with (2) limited backwards compatibility for managing the way basic card has been implemented up to now. We also want to check with MS on the proposal.

zkoch:
I think we've decided on calls that this particular direction is not what the group wants, so going to go ahead and close this PR.

adrianhopebailie:
@zkoch for completeness, can you link to the minutes where this decision was made as I think this may be premature (unless there is another issue tracking this?)

ianbjacobs:
Hi @adrianhopebailie,

Here are the most recent minutes:
 https://www.w3.org/2017/03/09-wpwg-minutes#item02

In the payment handler spec we will define a way for payment apps to provide capabilities to the user agent. 

My proposal was first to ensure that PR API explicitly takes those payment app capabilities into account in the matching algorithm(s). The spec has been edited to say that.

The specification does not provide details about how that is done (e.g., by string set matching). We can make suggestions in the payment app spec.

Ian

PR #421, opened 2017-02-16
editorial: s/character/code unit/ (closes #322)

### LINKED ISSUE #322: DOMStrings don't contain "characters"
There are various places in this specification that talk about things like "If the first character of details.total.amount.value is U+002D HYPHEN-MINUS" where the thing under conisideration (in this case details.total.amount.value) is a DOMString.  DOMStrings contain code units, not characters.  You either need to talk about the numeric values of those code units, or use USVString and talk about characters or something.

marcoscaceres:


zcorpan:
I think this doesn't address @bzbarsky's comment since DOMStrings contain code **units**. So we'd need to use https://heycam.github.io/webidl/#dfn-obtain-unicode before talking about code points. Or instead talk about code units...

marcoscaceres:
D'oh - will change to code unit. I do love these web "holly trinities" of "character", "code point", an "code unit".

marcoscaceres:
Ok, used "code unit" instead.

zcorpan:
Except this still doesn't make sense. 😂 "U+002D HYPHEN-MINUS" is a code point. It would need to be something like "0x002D code unit". Maybe in parenthesis "(Unicode HYPHEN-MINUS (-))". Or some such. As @domenic said, it is common for specs to talk about code points in DOMStrings even though that is technically wrong, and we haven't established what we should do instead...

marcoscaceres:
My suggestion is to drop "code unit" - and just say things like "if it starts with a U+002D HYPHEN-MINUS" or whatever.

domenic:
That doesn't really address the original issue though.

My plan was to ignore this until we figure out a coherent story over in whatwg/infra.

marcoscaceres:
> On 18 Feb 2017, at 2:54 am, Domenic Denicola <notifications@github.com> wrote:
> 
> That doesn't really address the original issue though.
> 

Well, it does in as far that it removes use of the word "character" - but yes, it doesn't address the core of the problem...

> My plan was to ignore this until we figure out a coherent story over in whatwg/infra.
> 

Ok, let's close this for now.

PR #423, opened 2017-02-16
editorial: fix linking to various specs

marcoscaceres:


index.html
```diff
@@ -342,7 +342,7 @@
           construction, generate a <code>paymentRequestId</code>.
           </li>
           <li>If the <a>current settings object</a>'s <a data-cite=
-          "!html5#responsible-document">responsible document</a> is not
+          "!HTML52#responsible-document">responsible document</a> is not
```
domenic: If you're fixing the links anyway, might as well go for the one that has the right definitions of e.g. allowed to use/allowpaymentrequest.

marcoscaceres: I'll cite "[[!HTML]]" - but you better have my back.

Issue #415, opened 2017-02-08
It is never stated how the payment request ID is used


domenic:
It appears that it's meant to be (a) set during construction, see #413, then (b) PaymentResponse is supposed to set its paymentRequestID to the same value of the PaymentRequest that generated it.

This latter part (b) never happens in the spec, so it is spec-complaint right now for paymentResponse.paymentRequestID to return anything at all.

"user accepts the payment request algorithm" needs to have a step that copies over the paymentRequestID value.

This should be done after #413 since it looks like #413 is going to introduce an internal slot to hold the paymentRequestID on the PaymentRequest object.

marcoscaceres:
closed via #426

Issue #412, opened 2017-02-07
PaymentRequestID initialization


rvm4:
I (a colleague) realized that in a previous PR I introduce the paymentRequestID field on PaymentRequest without any way to actually set this value. I would like to take this value in as the first parameter to the constructor of PaymentRequest since none of PaymentMethodData, PaymentDetails and PaymentOptions makes much sense for holding this value.

Thoughts?

rsolomakhin:
As long as it's optional :+1:

rvm4:
Great, I'll PR it.

marcoscaceres:
closed via #426

PR #426, opened 2017-02-22
feat: put request id in details dictionary

### LINKED ISSUE #388: Providing/Generating paymentRequestID during construction
The specification says that "If a paymentRequestID was not provided during construction, generate a paymentRequestID". 

Specification is not clear about providing paymentRequestID while construction. 
Constructor accepts PaymentMethodData, PaymentDetails and PaymentOptions.

Constructor(sequence<PaymentMethodData> methodData, PaymentDetails details, optional PaymentOptions options).

rvm4:
* closes #388, #412, #415.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/request_id_construct.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/3c3a100...113fb60.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1644,6 +1644,7 @@
         interface PaymentResponse {
           serializer = { attribute };
 
+          readonly attribute DOMString requestId;
           readonly attribute DOMString methodName;
           readonly attribute object details;
           readonly attribute PaymentAddress? shippingAddress;
--- a/index.html
+++ b/index.html
@@ -347,11 +347,11 @@
           <li>Establish the request's id:
             <ol>
               <li>If <var>details</var>.<a data-lt="PaymentDetails.id">id</a>
-              is missing, set the value of <var>details</var>.<a data-lt=
-              "PaymentDetails.id">id</a> to a string that uniquely identifies
-              this payment request. It is RECOMMENDED that the string be a
-              <abbr title="Universally Unique Identifier">UUID</abbr>
-              [[!RFC4122]].
+              is missing, add an <a data-lt="PaymentDetails.id">id</a> member
+              to <var>details</var> and set its value to string that uniquely
+              identifies this payment request. It is RECOMMENDED that the
+              string be a <abbr title=
+              "Universally Unique Identifier">UUID</abbr> [[!RFC4122]].
               </li>
             </ol>
           </li>
--- a/index.html
+++ b/index.html
@@ -1040,7 +1040,7 @@
       </h2>
       <pre class="idl">
         dictionary PaymentDetails {
-          DOMString paymentRequestId;
+          DOMString id;
           PaymentItem total;
           sequence&lt;PaymentItem&gt; displayItems;
           sequence&lt;PaymentShippingOption&gt; shippingOptions;
--- a/index.html
+++ b/index.html
@@ -344,6 +344,19 @@
           </li>
           <li>Let <var>serializedMethodData</var> be an empty list.
           </li>
+          <li>Establish the request's id:
+            <ol>
+              <li>Let <var>requestId</var> be the empty string.
+              </li>
+              <li>If <var>details.id</var> is present, set <var>requestId</var>
+              to the value of <var>details</var>.<a data-lt=
+              "PaymentDetails.id">id</a>. Otherwise, set <var>requestId</var>
+              to a string that uniquely identifies this payment request. It is
+              RECOMMENDED that the string be a <abbr title=
+              "Universally Unique Identifier">UUID</abbr> [[!RFC4122]].
+              </li>
+            </ol>
+          </li>
           <li>Process payment methods:
             <ol data-link-for="PaymentMethodData">
               <li>If the length of the <var>methodData</var> sequence is zero,
@@ -456,9 +469,6 @@
             </ol>
           </li>
           <li>Let <var>serializedModifierData</var> be an empty list.
-          <li>If <var>details.id</var> was not provided during
-          construction, populate the field with an [[RFC4122]] UUID.
-          </li>
           </li>
           <li data-link-for="PaymentDetailsBase">Process payment details
           modifiers:
@@ -537,6 +547,8 @@
             </ol>
           </li>
           <li>Let <var>request</var> be a new <a>PaymentRequest</a>.
+          </li>
+          <li>Set <var>request</var>.<a>id</a> to <var>requestId</var>.
           </li>
           <li>Set <var>request</var>.<a>[[\options]]</a> to <var>options</var>.
           </li>
@@ -1056,12 +1068,12 @@
           <dfn>id</dfn>
         </dt>
         <dd>
-          <a>id</a> is a freeform identifier for this
-          payment request. This field is intended to hold a unique
-          identifier for the purposes of recovering from failures and/or as a
-          foreign key into other systems. If a <a>id</a>
-          is not provided by the client then a UUID will be generated and stored
-          instead.
+          <a>id</a> is a free-form identifier for this payment request.
+          <p class="note">
+            If a <a>id</a> member is not present, then the UA will generate a
+            unique identifier for the payment request during <a data-lt=
+            "PaymentRequest.PaymentRequest()">construction</a>.
+          </p>
         </dd>
         <dt>
           <dfn>total</dfn>
@@ -1739,6 +1751,13 @@
           <a>PaymentRequest</a> constructor, then <a data-lt=
           "PaymentResponse.payerPhone">payerPhone</a> will be the phone number
           chosen by the user.
+        </dd>
+        <dt>
+          <dfn>requestId</dfn>
+        </dt>
+        <dd>
+          The corresponding payment request <a data-lt=
+          "PaymentRequest.id">id</a> that spawned this payment response.
         </dd>
       </dl>
       <section data-dfn-for="PaymentResponse" data-link-for="PaymentResponse">
--- a/index.html
+++ b/index.html
@@ -346,8 +346,8 @@
           </li>
           <li>Establish the request's id:
             <ol>
-              <li>If <var>details</var>.<a data-lt="PaymentDetails.id">id</a>
-              is missing, add an <a data-lt="PaymentDetails.id">id</a> member
+              <li>If <var>details</var>.<a data-lt="PaymentDetailsInit.id">id</a>
+              is missing, add an <a data-lt="PaymentDetailsInit.id">id</a> member
               to <var>details</var> and set its value to string that uniquely
               identifies this payment request. It is RECOMMENDED that the
               string be a <abbr title=
@@ -600,7 +600,7 @@
         <p>
           When getting, the <a>id</a> attribute returns this
           <a>PaymentRequest</a>'s <a>[[\details]]</a>.<a data-lt=
-          "PaymentDetails.id">id</a>.
+          "PaymentDetailsInit.id">id</a>.
         </p>
       </section>
       <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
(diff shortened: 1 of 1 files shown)


rsolomakhin:
Looks like you're passing in a `paymentRequestID` to the constructor, but remove its accessors and don't mention where this identifier is going.

If you expect the merchant website to read this identifier, please expose it through an accessor in `PaymentResponse` interface. Depending on your use-case, you may want to expose the identifier through an accessor in `PaymentRequest` interface as well.

If you expect the payment app to read the identifier, please add a note somewhere that it should be sent to the payment app.

index.html
```diff
@@ -301,6 +300,7 @@
           </p>
           <pre class="example" title="The 'details' argument">
             const details = {
+              paymentRequestID: "1234567890123",
```
marcoscaceres: s/paymentRequestID/paymentRequestId

index.html
```diff
@@ -1014,6 +1014,7 @@
       </h2>
       <pre class="idl">
         dictionary PaymentDetails {
+          DOMString paymentRequestID;
```
marcoscaceres: s/paymentRequestID/paymentRequestId/g

index.html
```diff
@@ -1033,6 +1034,17 @@
       </p>
       <dl>
         <dt>
+          <dfn>paymentRequestID</dfn>
```
marcoscaceres: s/paymentRequestID/paymentRequestId

index.html
```diff
@@ -1033,6 +1034,17 @@
       </p>
       <dl>
         <dt>
+          <dfn>paymentRequestId</dfn>
+        </dt>
+        <dd>
+          <a>paymentRequestId</a> is a freeform identifier for this
+          payment request. This field is intended to hold a unique
```
marcoscaceres: Re: "intended", as it reads, it could be confused that the browser is supposed to use it for error recovery. I would remove the second sentence.

index.html
```diff
@@ -1033,6 +1034,17 @@
...
+          identifier for the purposes of recovering from failures and/or as a
+          foreign key into other systems. If a <a>paymentRequestId</a>
+          is not provided by the client then a UUID will be generated and stored
```
marcoscaceres: The third sentence should be a non-normative note. Also, there is no requirement that this shall be a UUID in the payment request constructor.

index.html
```diff
@@ -204,7 +204,6 @@
           Promise&lt;void&gt; abort();
           Promise&lt;boolean&gt; canMakePayment();
 
-          readonly attribute DOMString? paymentRequestId;
```
marcoscaceres, line 205: Did you delete this by accident from the interface?

marcoscaceres: Also, the constructor algorithm doesn't ever set this to null - so maybe it should not be nullable. It's also unclear what happens if paymentRequestId is set to just "" (the empty string). 

Should setting it to `null` be treated as a special case (i.e., treat null as undefined instead of being converted to the word "null")?

index.html
```diff
@@ -460,6 +457,9 @@
             </ol>
           </li>
           <li>Let <var>serializedModifierData</var> be an empty list.
+          <li>If <var>details.paymentRequestId</var> was not provided during
+          construction, populate the field with an [[RFC4122]] UUID.
```
marcoscaceres: Oh, I see. I don't think we have consensus to make this a hard requirement. Also, avoid modifying the incoming data, create a new variable instead.

marcoscaceres: Drafting some alternative text...

index.html
```diff
@@ -333,9 +333,6 @@
           The <a>PaymentRequest</a> constructor MUST act as follows:
         </p>
         <ol data-link-for="PaymentDetails" class="algorithm">
-          <li>If a <code>paymentRequestId</code> was not provided during
```
marcoscaceres: Here, I think we want this instead:

```HTML
          <li>Establish the request's id:
            <ol>
              <li>Let <var>requestId</var> be the empty string.
              </li>
              <li>If <var>details.id</var> is present, set <var>
                requestId</var> to the value of
                <var>details.paymentRequestId</var>. Otherwise, generate a
                string to uniquely identify this payment request and set
                <var>requestId</var> to that value. It is RECOMMENDED that the
                generated string be a <abbr title=
                "Universally Unique Identifier">UUID</abbr> as per
                [[!RFC4122]].
              </li>
            </ol>
          </li>
```

Then, after the object is actually constructed:

```HTML
          <li>Set <var>request</var>.<a>paymentRequestId</a> to
          <var>requestId</var>.
```

marcoscaceres (dismissed):

index.html
marcoscaceres: Sent some text for the comments above.

marcoscaceres:
Fixed the nits via a commit. Let me know if you agree with the changes I proposed. I can add them in.

domenic:
I realize this is a bit late in the process, so feel free to ignore me, but is there a reason it's `paymentRequest.paymentRequestId` instead of `paymentRequest.id`? Same for the dictionary member I guess.

marcoscaceres:
I was thinking the same. It's not too late to change it.

marcoscaceres:
(no one shipped this, so we can change still)

marcoscaceres:
@domenic, so, in practice you would end up with (not ideal, IMO):

``` js
const request = new PaymentRequest(x, {paymentRequestId: "123121"},z);
// stuff happens... 
if (request.id === response.paymentRequestId ) {

}
```
What might be best is just call it `requestId` across all three: 

```JS
const request = new PaymentRequest(x, {requestId: "123312"}, z);
// stuff happens... 
if (request.requestId === response.requestId ) {

}
```

WDYT?

domenic:
`requestId` could be reasonable, although I was thinking something more like

```js
const request = new PaymentRequest(x, {id: "123121"},z);
// stuff happens... 
if (request.id === response.requestId) {

}
```

marcoscaceres:
@domenic, yeah, that works well actually... it's only the response that needs the qualifier.

marcoscaceres:
I've updated the PR to reflect discussion above:

 * `PaymentRequest` gains "id" attribute.
 * `PaymentResponse` gains "requestId" attribute.
 * `PaymentDetailsBase` gains "id" member.
 * PaymentRequest constructor algo gains "Establish the request's id:" sub-algo.  
 * UUID is RECOMMENDED.
 * PaymentResponse's .requestId gets set on construction.

marcoscaceres:
Question: do we want an internal slot for the [id] for when it's copied over from Request to Response? Can it be forged if the attribute value is just copied over? I guess having the internal slot is nicer - but I don't have an strong opinion.

marcoscaceres:
so, decided internal slots are fun - so added one. PR now reflects that.

index.html
```diff
@@ -830,6 +850,14 @@
           </tr>
           <tr>
             <td>
+              <dfn>[[\requestId]]</dfn>
+            </td>
+            <td>
+              A string, which can be empty, that identifies the a payment request.
```
rsolomakhin: s/the a/the/

rsolomakhin (approved):
LGTM

index.html
```diff
@@ -344,6 +342,17 @@
           </li>
           <li>Let <var>serializedMethodData</var> be an empty list.
           </li>
+          <li>Establish the request's id:
+            <ol>
+              <li>If <var>details.id</var> is present, set <a>[[\requestId]]</a>
```
domenic: Elsewhere (e.g. for details.shippingOptions) we've used the approach of normalizing the details.whatever, instead of creating a separate slot. I think that'd work best here too, so no separate internal slot for the ID. A separate slot leads to potential confusion where algorithms could accidentally reach into both places and get different values.

index.html
```diff
@@ -591,6 +602,15 @@
       </section>
       <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
         <h2>
+          <dfn>id</dfn> attribute
+        </h2>
+        <p>
+          When getting, the <a>id</a> attribute reflects the value of
+          the payment request's <a>[[\requestId]]</a> internal slot.
```
domenic: "Reflects" is a term of art that isn't quite appropriate here. You want "When getting, returns this PaymentRequest's [[requestId]] internal slot" (or, per the above advice, "this PaymentRequest's [[details]].id")

index.html
```diff
@@ -1033,6 +1062,17 @@
       </p>
       <dl>
         <dt>
+          <dfn>id</dfn>
+        </dt>
+        <dd>
+          <a>id</a> is a free-form identifier for this payment request.
+          <p class="note">
+            If a <a>id</a> member is not present, then the UA will generate a
```
domenic: s/a/an; maybe link "user agent" instead of using "UA"

index.html
```diff
@@ -1033,6 +1062,17 @@
...
+            unique identifier for the payment request during <a data-lt=
+            "PaymentRequest.PaymentRequest()">construction</a>.
```
domenic: This is a broken link according to ReSpec.

index.html
domenic: We should not treat `null` as a special case. It is no more special than `false` or `0` (all of them are "things that are vaguely negative, but are not the default-triggering special value of `undefined`").

domenic (changes requested):
Processing model looks good, just a few nits and the issue of where to store it.

index.html
marcoscaceres: Yep, I've got a patch to fix that in ReSpec. I'll try to land that soon.

marcoscaceres:
Rebased

index.html
```diff
@@ -583,6 +595,16 @@
       </section>
       <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
         <h2>
+          <dfn>id</dfn> attribute
+        </h2>
+        <p>
+          When getting, the <a>id</a> attribute returns this
+          <a>PaymentRequest</a> <a>[[\details]]</a>.<a data-lt=
```
domenic: Nit: missing `'s` after PaymentRequest

index.html
```diff
@@ -1035,7 +1057,9 @@
         <dfn>PaymentDetailsBase</dfn> dictionary
       </h2>
       <pre class="idl">
-        dictionary PaymentDetailsBase {
+        dictionary PaymentDetails {
```
domenic: Rebasing seems to have gone bad here.

index.html
```diff
@@ -1047,6 +1071,29 @@
...
+        <dd>
+          <a>id</a> is a free-form identifier for this payment request.
+          <p class="note">
+            If an <a>id</a> member is not present, then the <a>user agent</a>
+            will generate a unique identifier for the payment request during
+            <a data-lt="PaymentRequest.PaymentRequest()">construction</a>.
+          </p>
+        </dd>
+        <dt>
+          <dfn>total</dfn>
```
domenic: and here

marcoscaceres:
@domenic, sorry, can you recheck this one (request.id)? I'm not if everything was addressed or not.

domenic (changes requested):
Bad rebase but overall looking good to me.

index.html
marcoscaceres: Oh, we'll spotted. How odd :(

marcoscaceres:
Ok, I think I fixed the rebase issues.

domenic:
Oh, I think this also needs handling in updateWith()? Unless it should move to PaymentDetailsInit...

marcoscaceres:
I'm not sure it makes sense in `.updateWith()`, so PaymentDetailsInit would be the right place.

index.html
```diff
@@ -1131,6 +1154,17 @@
...
       <dl>
         <dt>
+          <dfn>id</dfn>
+        </dt>
+        <dd>
+          <a>id</a> is a free-form identifier for this payment request.
+          <p class="note">
+            If an <a>id</a> member is not present, then the <a>user agent</a>
+            will generate a unique identifier for the payment request during
+            <a data-lt="PaymentRequest.PaymentRequest()">construction</a>.
```
marcoscaceres, line 1164: ReSpec will link this soon...

domenic (approved):

PR #429, opened 2017-02-23
editorial: dfn payment request states as strings (closes #376)

### LINKED ISSUE #376: Editorial: define and link to the different states
That is, every time we mention a state like <i>interactive</i> it should link to a definition that describes it somewhere.

Also we could consider switching to strings for state names as that seems a bit more straightforward.

marcoscaceres:


domenic (approved):
LGTM, I did a number of Ctrl+Fs of various sorts and didn't find anything you missed :)

marcoscaceres:
thanks for the quick review!

PR #432, opened 2017-02-23
fix: s/user interaction/transaction (closes #402)

### LINKED ISSUE #402: `PaymentInstrument.complete(result)` wording
The spec says:

> Calling the complete() method tells the user agent that the user interaction is over (and should cause any remaining user interface to be closed).

This phrasing is confusing, because user is not interacting with the webpage before `PaymentInstrument.complete(result)` is called. The user agent would normally block user interaction with a "Processing transaction..." message before this call. Therefore, the phrasing should be changed to:

> Calling the complete() method tells the user agent that the _transaction_ is over (and should cause any remaining user interface to be closed).

marcoscaceres:


zkoch (commented):
This isn't quite as factually correct IMO. We don't know if the transaction is over or not. We only know the user interaction with the PaymentRequest sheet is over

marcoscaceres:
@zkoch, ok. So close this or can you recommend some improvements?

zkoch:
That moment when you realized you've contradicted yourself... I blame the jet lag.

Re-reading this and the original issue, I agree this looks good. Let's do it.

PR #433, opened 2017-02-23
Editorial: firing event not done correctly (closes #401)

### LINKED ISSUE #401: firing an event is not done correctly
E.g.

> 1. Let event be a new PaymentRequestUpdateEvent.
> 2. Queue a task to fire an event named name of type event at request. 

should be

> 1. Queue a task to fire an event named _name_ at _request_ using `PaymentRequestUpdateEvent`.

Note that this requires referencing the DOM Standard, which defines fire an event correctly, instead of DOM4, which does not allow using non-`Event` interfaces, or DOM 4.1, which does not define it at all.

marcoscaceres:


domenic (changes requested):
Since the original issue was filed, the "queue a task" was moved to the caller, so it should not be added here. Otherwise LGTM, feel free to merge after addressing that.

marcoscaceres:
Removed redundant "queue a task".

PR #435, opened 2017-02-23
editorial: s/shown/presented (closes #428)

### LINKED ISSUE #428: "Shown" should be "presented" in prose
Although we have `.show()`, the modality of the presentation of the payment request depends on the user's preferences and abilities. That is, a visually impaired user might hear the request for payment - instead of seeing it. Thus, we should really say "presented" where we say "shown".

marcoscaceres:


zkoch (approved):
I think this works fine either way, but I don't object to the change

PR #438, opened 2017-02-23
editorial: attr/method intros are notes (closes #379)

### LINKED ISSUE #379: Introductory explanations of methods should be non-normative notes
Each method has a paragraph explaining what it does. Unfortunately it often contains normative terms like must/should/etc.. They should be rephrased to explain instead of dictate, and should be explicitly marked as non-normative notes. Maybe they should even be formatted in the "domintro" style used in various specs (e.g. HTML/DOM/IndexedDB/...)

marcoscaceres:
***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/intro-notes.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/712e10b...217b760.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -121,7 +121,7 @@
         authorizing as necessary across the flow.
       </p>
       <p>
-        The details of how to fulfil a payment request for a given <a>payment
+        The details of how to fulfill a payment request for a given <a>payment
         method</a> are handled by <dfn data-lt="payment app">payment
         apps</dfn>. In this specification, these details are left up to the
         <a>user agent</a>, but future specifications may expand on the
@@ -836,7 +836,7 @@
         <h2>
           <dfn>onshippingoptionchange</dfn> attribute
         </h2>
-        <p class="note">
+        <p>
           A <a>PaymentRequest</a>'s <a>onshippingoptionchange</a> attribute is
           an <code>EventHandler</code> for an <code>Event</code> named
           <code>shippingoptionchange</code>.
@@ -1722,7 +1722,7 @@
         </h2>
         <p>
           The <a>payment method identifier</a> for the <a>payment method</a>
-          that the user selected to fulfil the transaction.
+          that the user selected to fulfill the transaction.
         </p>
       </section>
       <section>
@@ -1962,9 +1962,9 @@
         </p>
         <p>
           If the web page wishes to update the payment request then it should
-          call <a>updateWith()</a> and provide a promise that will resolve with
-          a <a>PaymentDetailsUpdate</a> dictionary containing changed values
-          that the <a>user agent</a> SHOULD present to the user.
+          call <a>updateWith()</a> and provide a <a>PaymentDetailsUpdate</a>
+          dictionary, or a promise for one, containing changed values that the
+          <a>user agent</a> SHOULD present to the user.
         </p>
         <p>
           The <a>PaymentRequestUpdateEvent</a> constructor MUST set the
@@ -1975,10 +1975,10 @@
             <dfn data-lt="updateWith(detailsPromise)">updateWith()</dfn> method
           </h2>
           <p class="note">
-            If the web page wishes to update the payment request then it can
-            call <a>updateWith()</a> and provide a promise that will resolve
-            with a <a>PaymentDetailsUpdate</a> dictionary containing changed
-            values that the <a>user agent</a> presents to the user.
+            If the web page wishes to update the payment request then it should
+            call <a>updateWith()</a> and provide a <a>PaymentDetailsUpdate</a>
+            dictionary, or a promise for one, containing changed values that
+            the <a>user agent</a> presents to the user.
           </p>
           <p>
             The <a>updateWith(detailsPromise)</a> method MUST act as follows:
--- a/index.html
+++ b/index.html
@@ -1482,94 +1482,6 @@
           readonly attribute DOMString phone;
         };
       </pre>
-      <dl>
-        <dt>
-          <dfn>serializer</dfn>
-        </dt>
-        <dd>
-          Each attribute is <a data-cite=
-          "WEBIDL-LS#dfn-convert-to-serialized-value">converted to serialized
-          values</a> as per [[!WEBIDL-LS]].
-        </dd>
-        <dt>
-          <dfn>country</dfn>
-        </dt>
-        <dd>
-          This is the [[CLDR]] (Common Locale Data Repository) region code. For
-          example, US, GB, CN, or JP.
-        </dd>
-        <dt>
-          <dfn>addressLine</dfn>
-        </dt>
-        <dd>
-          This is the most specific part of the address. It can include, for
-          example, a street name, a house number, apartment number, a rural
-          delivery route, descriptive instructions, or a post office box
-          number.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -787,28 +790,42 @@
         <h2>
           <dfn>shippingAddress</dfn> attribute
         </h2>
-        <p>
-          <a>shippingAddress</a> is populated when the user provides a shipping
-          address. It is null by default. When a user provides a shipping
-          address, the <a>shipping address changed algorithm</a> runs.
+        <p class="note">
```
domenic: Hmm. I am less sure about the attributes than about the methods. Since their behavior isn't defined in these sections at all, maybe it is OK to leave the text as non-note, as long as we don't use any normative keywords. The risk with methods is that people might think the notes are what specs their behavior.

index.html
```diff
@@ -787,28 +790,42 @@
...
+          populated when the user provides a shipping address. It is null by
+          default. When a user provides a shipping address, the <a>shipping
+          address changed algorithm</a> runs.
         </p>
+      </section>
+      <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
+        <h2>
+          <dfn>onshippingaddresschange</dfn> attribute
+        </h2>
         <p>
```
domenic, line 818: This is inconsistent with the rest (missing class="note"). But as per above I'd be OK with omitting class="note" for all the attributes...

index.html
domenic: A totally different style you might consider:

- Get rid of the separate sections for each attribute.
- Add normative text to the overall PaymentRequest interface section, or maybe to an attributes subsection: "The shippingAddress and shippingOption attributes must return the values they are initialized to. The onshippingaddresschange and onshippingoptionchange attributes are event handle IDL attributes for the shippingaddresschange and shippingoptionchange events, respectively."
- Introduce "domintro" style notes for web developers. Examples: https://html.spec.whatwg.org/#pixel-manipulation https://w3c.github.io/IndexedDB/#request-api. (You'd have separate domintro boxes for each method + the constructor, plus one for the attributes, I think.)

I think this would help especially with the attributes, to separate the normative parts from the web developer info (e.g. "default value is null" is for the web developer; the normative algorithm that sets it to null takes place in the constructor).

index.html
```diff
@@ -1070,10 +1087,10 @@
...
           This <a>PaymentItem</a> contains the total amount of the payment
           request.
-          <p>
-            <code>total</code> MUST be a non-negative value. This means that
-            the <code>total.amount.value</code> field MUST NOT begin with a
-            U+002D HYPHEN-MINUS character.
+          <p class="note">
+            The <a>total</a>.<a data-lt=
+            "PaymentItem.amount">amount</a>.<a data-lt=
+            "PaymentCurrencyAmount.value">value</a> can't be a negative number.
```
domenic: Maybe "Algorithms in this specification that accept a PaymentDetails dictionary will throw if the total.amount.value is a negative number."

index.html
```diff
@@ -1704,14 +1721,14 @@
...
+        <p class="note">
           The <a>complete()</a> method is called after the user has accepted
           the payment request and the <a>[[\acceptPromise]]</a> has been
           resolved. Calling the <a>complete()</a> method tells the <a>user
-          agent</a> that the user interaction is over (and SHOULD cause any
+          agent</a> that the user interaction is over (and can cause any
           remaining user interface to be closed).
         </p>
-        <div class="note">
+        <div class="note" title="Dealing with blocked browser UI">
```
domenic: Honestly this feels like a normative requirement, not a note.

domenic (changes requested):
No need to go for the whole domintro style, but there are a few minor things that should be fixed before merging.

index.html
domenic: Oh, also note that PaymentResponse currently uses a completely different style for its attributes, instead of having a heading per attribute :-/.

marcoscaceres: > Hmm. I am less sure about the attributes than about the methods.

Agree. Removed "note".  Making statements of fact with non-RFC2119 words seems good (even if they mostly benefit developers). 

> Oh, also note that PaymentResponse currently uses a completely different style for its attributes, instead of having a heading per attribute :-/.

We should switch those to be headings (separate bug) - personally, I find it frustrating when attributes are not in the ToC - one should be able to see/access the API shape there too.

index.html
marcoscaceres: Agree. I've changed the second "may" to an RFC2119 "MAY" and un-noted these two paragraphs.

index.html
adrianhopebailie: > We should switch those to be headings (separate bug) - personally, I find it frustrating when attributes are not in the ToC - one should be able to see/access the API shape there too.

+1

marcoscaceres: @adrianhopebailie I filed: https://github.com/w3c/browser-payment-api/issues/442 

Kindly please also note you can "react" to a comment via github's interface:

screenshot 2017-02-27 18 54 16

It saves generating an email with only "+1" in it, while allowing you to throw support behind a proposal.

index.html
marcoscaceres: Leaving attributes without notes.

marcoscaceres: I'll convert the PaymentResponse attributes into sections/headings to match.

index.html
```diff
@@ -121,7 +121,7 @@
         authorizing as necessary across the flow.
       </p>
       <p>
-        The details of how to fulfill a payment request for a given <a>payment
+        The details of how to fulfil a payment request for a given <a>payment
```
domenic: The original word was correct here

index.html
```diff
@@ -802,27 +805,41 @@
...
         </p>
-        <p>
-          <dfn>onshippingoptionchange</dfn> is an <code>EventHandler</code> for
-          an <code>Event</code> named <code>shippingoptionchange</code>.
+      </section>
+      <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
+        <h2>
+          <dfn>onshippingoptionchange</dfn> attribute
+        </h2>
+        <p class="note">
```
domenic: This should not be a note, to be consistent

index.html
```diff
@@ -1918,6 +1974,12 @@
           <h2>
             <dfn data-lt="updateWith(detailsPromise)">updateWith()</dfn> method
           </h2>
+          <p class="note">
+            If the web page wishes to update the payment request then it can
+            call <a>updateWith()</a> and provide a promise that will resolve
+            with a <a>PaymentDetailsUpdate</a> dictionary containing changed
```
domenic: Maybe point out that you don't have to provide a promise. E.g. "provide a PaymentDetailsUpdate dictionary, or a promise for one, containing..."

domenic (approved):
Some nits to fix, but this is quite a bit nicer, yay.

index.html
marcoscaceres: Ah right, my editor is misconfigured I think: https://en.wiktionary.org/wiki/fulfil

We should change those to en-us. Or is there a difference between "fulfill" and "fulfil" that I don't know about?

PR #445, opened 2017-02-28
Create pr-preview config file.

tobie:
(Using https://tobie.github.io/pr-preview/config.html)

This still requires someone from @w3c to install the [PR preview integration](https://github.com/integration/pr-preview). Either for all org repos (opt-in to using it by adding a config file like this one), or just for this one.

### DIFF
--- a/.pr-preview.json
+++ b/.pr-preview.json
@@ -1,0 +1,4 @@
+{
+    "src_file": "index.html",
+    "type": "respec"
+}


marcoscaceres:
@dontcallmedom or @ianbjacobs, would you be so kind as to install it: https://github.com/integration/pr-preview

marcoscaceres:
(would be awesome if it was globally installed for the whole W3C organization, btw... the PR diff tools seems very useful)

sideshowbarker:
> would you be so kind as to install it: https://github.com/integration/pr-preview

Installed and showing up at https://github.com/w3c/browser-payment-api/settings/installations

Also changed your perms for the repo to Admin so if the PR preview tool doesn’t seem to be work as expected, feel free to troubleshoot ti

marcoscaceres:
Thanks @sideshowbarker!

tobie:
FYI: there were huge AWS outages yesterday which affected pr-preview greatly. Not sure how stable things are right now. So do expect some disturbance.

marcoscaceres:
Yeah, it's still pretty bad out there... Travis CI is still struggling ATM.

sideshowbarker:
Amazing distributed system we’ve spent all these years creating technologies for! Would be a crazy situation if instead our only choice in practice was among just two or three giant vendors and they all run into different major outages relatively regularly. Imagine that!

PR #450, opened 2017-03-07
Section for A11Y Considerations

halindrome:
As per an APA Resolution ages ago [1] , this PR integrates text to help clarify that there are no explicit A11Y requirements but that implementors should ensure the UI they end up providing is one where A11Y is taken into account.

[1] http://lists.w3.org/Archives/Public/public-apa/2016Aug/0080.html

***
[Preview](https://s3.amazonaws.com/pr-preview/halindrome/browser-payment-api/apa-considerations.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/dc6350c...halindrome:6d9c9f1.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2428,58 +2428,71 @@
         </ol>
       </section>
     </section>
-    <section class='informative'>
-      <h2>
-        Security Considerations
-      </h2>
-      <p class="ednote">
-        This section is a placeholder to record security considerations as we
-        gather them through working group discussion.
-      </p>
-      <section>
+    <section class="informative">
+      <h2>Special Considerations</h2>
+      <section class='informative'>
         <h2>
-          Encryption of data fields
+          Security Considerations
         </h2>
-        <p>
-          The <a>PaymentRequest</a> API does not directly support encryption of
-          data fields. Individual <a>payment methods</a> may choose to include
-          support for encrypted data but it is not mandatory that all
-          <a>payment methods</a> support this.
+        <p class="ednote">
+          This section is a placeholder to record security considerations as we
+          gather them through working group discussion.
         </p>
+        <section>
+          <h2>
+            Encryption of data fields
+          </h2>
+          <p>
+            The <a>PaymentRequest</a> API does not directly support encryption of
+            data fields. Individual <a>payment methods</a> may choose to include
+            support for encrypted data but it is not mandatory that all
+            <a>payment methods</a> support this.
+          </p>
+        </section>
       </section>
-    </section>
-    <section class='informative'>
-      <h2>
-        Privacy Considerations
-      </h2>
-      <p class="ednote">
-        This section is a placeholder to record privacy considerations as we
-        gather them through working group discussion.
-      </p>
-      <section>
+      <section class='informative'>
         <h2>
-          Exposing user information
+          Privacy Considerations
         </h2>
-        <p>
-          The <a>user agent</a> MUST NOT share information about the user to
-          the web page (such as the shipping address) without user consent.
+        <p class="ednote">
+          This section is a placeholder to record privacy considerations as we
+          gather them through working group discussion.
         </p>
+        <section>
+          <h2>
+            Exposing user information
+          </h2>
+          <p>
+            The <a>user agent</a> MUST NOT share information about the user to
+            the web page (such as the shipping address) without user consent.
+          </p>
+        </section>
+        <section>
+          <h2>
+            Exposing available payment methods
+          </h2>
+          <p>
+            A page might try to call the payment request API repeatedly with only
+            one payment method identifier to try to determine what payment
+            methods a <a>user agent</a> has installed. There may be legitimate
+            scenarios for calling repeatedly (for example, to control the order
+            of payment method selection). The fact that a successful match to a
+            payment method causes a user interface to be displayed mitigates the
+            disclosure risk. Implementations may also require a user action to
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2428,58 +2428,71 @@
...
         <h2>
-          Encryption of data fields
+          Security Considerations
         </h2>
-        <p>
-          The <a>PaymentRequest</a> API does not directly support encryption of
-          data fields. Individual <a>payment methods</a> may choose to include
-          support for encrypted data but it is not mandatory that all
-          <a>payment methods</a> support this.
+        <p class="ednote">
```
marcoscaceres, line 2437: I think this should be an "issue" pointing to the relevant bug in the issue tracker.

index.html
```diff
@@ -2428,58 +2428,71 @@
...
-    <section class='informative'>
-      <h2>
-        Security Considerations
-      </h2>
-      <p class="ednote">
-        This section is a placeholder to record security considerations as we
-        gather them through working group discussion.
-      </p>
-      <section>
+    <section class="informative">
```
marcoscaceres, line 2431: I'm not sure there is much value in grouping these into a larger section.

index.html
```diff
@@ -2428,58 +2428,71 @@
...
-          of payment method selection). The fact that a successful match to a
-          payment method causes a user interface to be displayed mitigates the
-          disclosure risk. Implementations may also require a user action to
-          initiate a payment request or they may choose to rate limit the calls
-          to the API to prevent too many repeated calls.
-        </p>
+      <section class="informative">
+          <h2>Accessibility Considerations</h2>
+          <p>
+          This specification has no defined user interface. In addition, there are no specific accessibility requirements on implementations. However, to the extent that an
```
marcoscaceres, line 2490: I don't agree with this. There are absolutely accessibility requirements on the user interface:  for example, where the line items are shown, those much be accessible to people. Same with totals, etc.

index.html
```diff
@@ -2428,58 +2428,71 @@
...
+          implementation provides user interactions in support of this specification, the implementation must ensure that the interface for those interactions is exposed to the
+          platform accessibility API. Moreover, implementors should take into consideration the needs of their users with varying abilities when designing solutions that
```
marcoscaceres, line 2492: I don't think these generic statements are very helpful, tbh. I think we should be prescriptive where we can be.

marcoscaceres (changes requested):

index.html
halindrome: Each item was so small as to beg the question why would a top level section be in place for them.  Moreover, as they are all related to W3C horizontal review requirements, it felt sensible to gather them together to me.

index.html
halindrome: I am not sure why this showed up in this PR, but this is not content related to the change request and I don't have a preference.  It isn't my text.

index.html
halindrome: No.  The specification explicitly has no UI requirements.  There is no requirement that any of that information be presented at all.  And it is outside of the charter of the group to specify UI requirements.  It is up to user agents and payment apps.

index.html
halindrome: Understood.  But without specific UI requirements that is basically impossible.  The APA has indicated that this sort of guidance is sufficient for their current purposes (although they are also producing a Payment Accessibility User Requirements document that will have more specific examples).

index.html
chaals: The specification doesn't explicitly define a UI, however, it can and as Marcos suggests should define functional interface requirements for accessibility.

index.html
chaals: This is not consistent with itself. There is a concrete accessibility requirement named: That anything exposed in the UI is exposed to the accessibility API.

This seems necessary but not sufficient. Information needs to meet accessibility requirements for users other than those who rely on the accessibility API, too.

ianbjacobs:
@halindrome, @chaals,

Is there a good example of a W3C spec that includes a requirement to ensure that data sent through the API is available to the user agent's accessibility API? 

Ian

chaals:
@ianbjacobs the [HTML AAM spec](https://www.w3.org/TR/html-aam-1.0/)

>  …defines how user agents map HTML 5.1 [HTML51] elements and attributes to platform accessibility application programming interfaces (APIs).

You might not want to constrain your own spec quite so tightly. The goal is to be useful, after all.

ianbjacobs:
@chaals and @halindrome,

From @chaals' reference to HTML AAM [1] I am gathering that the idea would be to map the data flowing through the system (e.g., address, currency, amount, and so forth) to familiar platform-specific accessibility APIs.  

@halindrome, would you be able to start a companion document that does so for the data elements of PR API?

Ian

https://www.w3.org/TR/html-aam-1.0/

marcoscaceres:
Given that there has been no movement on this for the last few months, closing.

chaals:
(Sorry, I thought I had said this when it was timely :( Apparently not.)

> From @chaals' reference to HTML AAM [1] I am gathering that the idea would be to map the data flowing through the system (e.g., address, currency, amount, and so forth) to familiar platform-specific accessibility APIs.

Not at all. The idea would be to say, as noted in @marcoscaceres' comments on the PR, that it is vital that the data presented is accessible to users. Pushing them to an accessibility API is not going to achieve that for most users, and is quite possibly unnecessary even for the few users who can potentially benefit from it.

PR #451, opened 2017-03-08
Make updateWith() reject the show() promise on validation failures

### LINKED ISSUE #350: updateWith will silently ignore data in hard-to-debug ways

domenic:
Previously, it would ignore any bad data. Now it validates the data ahead of time before performing any updates.

Closes #350. Supercedes #418.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/new-updatewith.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/361e720...943a457.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1137,14 +1137,14 @@
             If an item in the sequence has the <a data-lt=
             "PaymentShippingOption.selected">selected</a> member set to true,
             then this is the shipping option that will be used by default and
-            <a>shippingOption</a> will be set to the <a data-lt=
-            "PaymentShippingOption.id">id</a> of this option without running
-            the <a>shipping option changed algorithm</a>. Authors SHOULD NOT
-            set <a data-lt="PaymentShippingOption.selected">selected</a> to
-            true on more than one item. If more than one item in the sequence
-            has <a data-lt="PaymentShippingOption.selected">selected</a> set to
-            true, then <a>user agents</a> MUST select the last one in the
-            sequence.
+            <a data-lt="PaymentRequest.shippingOption">shippingOption</a> will
+            be set to the <a data-lt="PaymentShippingOption.id">id</a> of this
+            option without running the <a>shipping option changed
+            algorithm</a>. Authors SHOULD NOT set <a data-lt=
+            "PaymentShippingOption.selected">selected</a> to true on more than
+            one item. If more than one item in the sequence has <a data-lt=
+            "PaymentShippingOption.selected">selected</a> set to true, then
+            <a>user agents</a> MUST select the last one in the sequence.
           </p>
           <p>
             The <a>shippingOptions</a> member is only used if the
@@ -2057,8 +2057,8 @@
                 The remaining steps are conditional on the
                 <var>detailsPromise</var> settling. If
                 <var>detailsPromise</var> never settles then the payment
-                request is blocked. Users should always be able to cancel a
-                payment request. Implementations may choose to implement a
+                request is blocked. Users SHOULD always be able to cancel a
+                payment request. Implementations MAY choose to implement a
                 timeout for pending updates if <var>detailsPromise</var>
                 doesn't settle in a reasonable amount of time. If an
                 implementation chooses to implement a timeout, they must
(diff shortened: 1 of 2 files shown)


domenic:
Remaining questions I have:

- Is the behavior for the `error` member correct? Or should the presence of that lead to the UA ignoring everything else and just showing the error?
- The constructor does not allow negative modifier totals, but updateWith() does. Is that correct?
- Maybe I can refactor the constructor and updateWith() to share some logic here.

marcoscaceres:
> Is the behavior for the error member correct? Or should the presence of that lead to the UA ignoring everything else and just showing the error?

Hmm... good question... it feels like you could do both: update things, but also show an error as needed.  Not sure what implementations are doing today. 

> The constructor does not allow negative modifier totals, but updateWith() does. Is that correct?

`updateWith()` also should forbid a negative total.   

> Maybe I can refactor the constructor and updateWith() to share some logic here? 

If you feel it's worth it. I'd be ok with some minor duplication, but up to you.

domenic:
> Hmm... good question... it feels like you could do both: update things, but also show an error as needed. Not sure what implementations are doing today.

@rsolomakhin, thoughts?

rsolomakhin:
>> Is the behavior for the error member correct? Or should the presence of that lead to the UA ignoring everything else and just showing the error?
>
> Hmm... good question... it feels like you could do both: update things, but also show an error as needed. Not sure what implementations are doing today.

Current behavior is the following:

1. If shipping options are empty and error is present:
    1. Re-enable interaction with user interface.
    1. Show the specified error.
    1. Gray out and unselect the previously selected shipping address.
1. If shipping options are empty, but the error is not present.
    1. Re-enable interaction with user interface.
    1. Show a generic error string about this address being invalid.
    1. Gray out and unselect the previously selected shipping address.
1. If shipping options are not empty:
    1. Re-enable interaction with user interface.
    1. Ignore any error, even if it's present.
    1. Update the shipping options in user interface according to what was received from the website.

We've had some discussions about this algorithm being confusing, because the shipping options dictate what to do with the error. I'm open to suggestion to improve this.

domenic:
Oh, hmm, that is confusing. I thought `error` was just a generic error signifier, not specifically related to shipping addresses. Also this idea that empty shipping options is bad is new to me: it's not captured anywhere in the spec, really.

I think it might be less confusing if you changed 3.ii. to show the error anyway. Then you end up treating shippingOptions and error as separate concerns, although authors will likely use them together.

Alternately, we could make the spec very clear that `error` is only to be used for shipping option-related errors, and will be ignored if requestShipping is false or shippingOptions is non-empty. In this case I'd ideally like to rename `error` to `shippingOptionsError` or similar, but that's not required, and is probably disruptive.

Thoughts between those two?

I guess in any case we should make it clear in the spec that if you pass `shippingOptions: []` to updateWith(), the UA should display an error, using `error` if provided.

rsolomakhin:
I'd prefer to make it clear that `error` is used only for shipping option related errors and will be ignored if `requestShipping` is `false` or `shippingOptions` is non-empty.

The error message is for the user selected shipping address, so `shippingAddressError` would be a better name. However, I'd prefer to avoid such disruptive rename. Can we keep the `error` name, but clearly annotate it in the spec instead?

domenic:
> The error message is for the user selected shipping address, so shippingAddressError would be a better name. However, I'd prefer to avoid such disruptive rename. Can we keep the error name, but clearly annotate it in the spec instead?

Yep, I'm fine with that. I'll try to write up something capturing this now. Hopefully at least from the spec it'll then be clear to developers.

domenic:
OK, updated. Please take a look! We should be sure to mention the change to error semantics in the commit message when merging.

index.html
```diff
@@ -2069,194 +2074,239 @@
...
+                                <li>Let <var>amountValue</var> be
+                                <var>item</var>.<a data-lt=
+                                "PaymentItem.amount">amount</a>.<a data-lt=
+                                "PaymentCurrencyAmount.value">value</a>.
+                                </li>
+                                <li>If <var>amountValue</var> is not a <a>valid
+                                decimal monetary value</a>, then <a>abort the
+                                update</a> with a <a>TypeError</a>.
+                                </li>
+                                <li>If the first character of
```
rsolomakhin: It should be OK to have negative line items in `additionalLineItems`. This can be "Debit card discount," for example. Let's reomve this `<li>...</li>`.

rsolomakhin (changes requested):
Let's allow negative amounts for `additionalDisplayItems`. Only `total` values should be non-negative.

domenic:
@rsolomakhin thanks for catching; I definitely inserted the check in the wrong spot there. Moved it.

rsolomakhin (approved):
LGTM

zkoch:
I think we got LGTMs from everyone, so merging this in

PR #452, opened 2017-03-08
Fix constructor handling of shippingType

domenic:
If shippingType is provided, it will always be a valid shipping type, now that we are using an enum (as of 7e9a3323ac7ed932597fe4b3963331f9fa883041). But we also need to correctly handle the case where it is not provided, or where requestShipping is false, in which case it should be null.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/fix-shipping-type.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/361e720...cfe929b.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -567,10 +567,6 @@
           <li>Set the value of the <a data-lt="PaymentRequest.shippingAddress">
             shippingAddress</a> attribute on <var>request</var> to null.
           </li>
-          <li>Set the value of the <a data-lt=
-          "PaymentRequest.shippingType">shippingType</a> attribute on
-          <var>request</var> to null.
-          </li>
           <li>If <var>options</var>.<a data-lt=
           "PaymentOptions.requestShipping">requestShipping</a> is set to true,
           then set the value of the <a data-lt=
--- a/index.html
+++ b/index.html
@@ -576,18 +576,8 @@
           then set the value of the <a data-lt=
           "PaymentRequest.shippingType">shippingType</a> attribute on
           <var>request</var> to <var>options</var>.<a data-lt=
-          "PaymentOptions.shippingType">shippingType</a>. If
-          <var>options</var>.<a data-lt=
-          "PaymentOptions.shippingType">shippingType</a> is not a valid
-          <a>PaymentShippingType</a> value then set the value of the
-            <a data-lt="PaymentRequest.shippingType">shippingType</a> attribute
-            on <var>request</var> to "<code>shipping</code>".
-            <div class="note">
-              This behavior allows a page to detect if it supplied an
-              unsupported shipping type. This will be important if new shipping
-              types are added to a future version of this specification but a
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):

domenic:
Wait, oops, it's always provided, since it has a default value... one sec.

marcoscaceres:
Ah yeah, forgot about that. So I guess no null then?

domenic:
Well, it's nullable, so I assume it was meant to be null in at least some situation.... and when requestShipping = false seems like a good guess as to when that was. I don't have a strong opinion myself, but it kind of makes sense as null in that case. @rsolomakhin, thoughts?

rsolomakhin:
Does not make much difference to me. If shipping option is null or absent, then Chrome defaults to "shipping" anyway.

zkoch:
I agree it should be null when requestShipping = false

domenic:
OK cool. I think this is ready to merge then.

zkoch:
Sorry, want to clarify what's happening here.

Technically if requestShipping = true, then shippingType default to "shipping". But in this case, doesn't it default to "null"?

Or am I misreading?

domenic:
Yeah, it's a bit confusing because this is defined partially in Web IDL and partially in the prose. Here is the breakdown of how things would work after this spec PR:

```js
const p = new PaymentRequest(m, d);
p.shippingType === null;
```

```js
const p = new PaymentRequest(m, d, { });
p.shippingType === null;
```

```js
const p = new PaymentRequest(m, d, { requestShipping: false });
p.shippingType === null;
```

```js
const p = new PaymentRequest(m, d, { requestShipping: false, shippingType: "shipping" });
p.shippingType === null;
```

```js
const p = new PaymentRequest(m, d, { requestShipping: true });
p.shippingType === "shipping";
```

```js
const p = new PaymentRequest(m, d, { requestShipping: true, shippingType: "delivery" });
p.shippingType === "delivery";
```

```js
new PaymentRequest(m, d, { requestShipping: true, shippingType: "asdf" });
// throws an exception
```

```js
new PaymentRequest(m, d, { requestShipping: true, shippingType: null });
// throws an exception
```

Currently what the spec says doesn't make any sense, so it needs to be fixed. But we could fix it so that the first 4 cases have `p.shippingType === null`, or `p.shippingType === "shipping"`. This patch goes with `null` but we could go either way.

ALSO: I just noticed this PR needs some tweaks because I forgot there's a step before this that already sets it to null. So I should combine those two steps. So please nobody merge yet. But sign off on the behavior would be good :)

domenic:
OK I fixed up the PR so it's ready to merge if we are OK with the behavior. I also realized my above post with all the examples was a bit wrong because I didn't realize there was no `paymentRequest.requestShipping`---i.e., it's only an option to the constructor, not an attribute accessible on PaymentRequest instances. So I edited it to remove such lines.

PR #453, opened 2017-03-10
Remove trailing comma from IDL enums

zcorpan:
28d01b4fdca2da26efe622a0082f8f3956815328 introduced trailing commas
in IDL enums. From what I can tell from
https://heycam.github.io/webidl/#idl-enums that is not allowed.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/zcorpan/enum-trailing-comma.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/361e720...3066910.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1340,7 +1340,7 @@
         enum PaymentShippingType {
           "shipping",
           "delivery",
-          "pickup",
+          "pickup"
         };
       </pre>
       <dl>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Sucks that it's not allowed. We should get that fixed in WebIDL.

PR #455, opened 2017-03-14
l10n: adds Localizable dict + TextDirection enum

### LINKED ISSUE #327: Add dir/lang member on dictionaries whose content is displayed in browser UI
As we can't mark-up the labels in dictionaries (using HTML), we might need to add `dir` ("ltr" and "rtl") members on dictionaries that contain things that are displayed to the end-user (defaulting to "ltr" when missing). So:

```JS
[
  displayItems: [{
    label: "البند الخاص (للبيع!)",
    dir: "rtl",
    lang: "ar-AE",
    amount: { },
  }]
]
```

Have the i18n folks looked at this API? They will likely ask us to add that. We might also need `lang`. 

cc @r12a.

marcoscaceres:
*  closes #327
* `PaymentShippingOption` and `PaymentItem` inherits from WebIDL `Localizable`
* The `label` members of the above are defined as localizable members

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/l10n.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/035d129...7da7748.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1504,7 +1504,7 @@
       </h2>
       <pre class="idl">
         dictionary PaymentItem {
-          required DOMString label;
+          required LocalizableString label;
           required PaymentCurrencyAmount amount;
           boolean pending = false;
         };
@@ -1680,7 +1680,7 @@
       <pre class="idl">
         dictionary PaymentShippingOption {
           required DOMString id;
-          required DOMString label;
+          required LocalizableString label;
           required PaymentCurrencyAmount amount;
           boolean selected = false;
         };
@@ -2847,6 +2847,10 @@
           Web IDL
         </dt>
         <dd>
+          <p>
+            The <code><dfn>LocalizableString</dfn></code> <a data-cite=
+            "!WEBIDL-LS#dfn-typedef">typedef</a> is defined by [[!WEBIDL-LS]].
+          </p>
           <p>
             When this specification says to <dfn data-cite="!WEBIDL#dfn-throw"
             data-lt="throws">throw</dfn> an error, the <a>user agent</a> must


marcoscaceres:
WIP, seeking feedback. It's just missing a statement about how to use the information when presenting the strings in a UI.

index.html
```diff
@@ -1108,7 +1108,7 @@
         <dfn>PaymentDetailsBase</dfn> dictionary
       </h2>
       <pre class="idl">
-        dictionary PaymentDetailsBase {
+        dictionary PaymentDetailsBase : Localizable {
```
marcoscaceres: I want this to apply to `PaymentDetailsUpdate.error`, which is why  put it here.

index.html
```diff
@@ -1258,10 +1258,10 @@
         <dd>
           When the payment request is updated using <a data-lt=
           "PaymentRequestUpdateEvent.updateWith">updateWith()</a>, the
-          <a>PaymentDetailsUpdate</a> can contain a message in the <a>error</a>
-          member that will be displayed to the user. For example, this might
-          commonly be used to explain why goods cannot be shipped to the chosen
-          shipping address.
+          <a>PaymentDetailsUpdate</a> can contain a human-readable message in
```
marcoscaceres: I'm using "human-readable" as the key that binds together all these "localizable" members. Need to formally define "human-readable" as being potentially affected by dir/lang when present.

marcoscaceres: ("human-readable" was already in use in the spec, which is why I reused it - maybe we can come up with something nicer)

domenic:
Would be good to land after https://github.com/w3c/browser-payment-api/pull/451 to avoid conflicts.

Notably this does not allow serving up multiple options with different locales and allowing the browser to choose between them. I guess that is a non-goal.

It's also noteworthy that this matches NotificationOptions already, so that's good. Maybe we can consolidate this `Localizable` type into Web IDL or similar. /cc @annevk @beverloo @tobie. Notifications defaults `lang` to `""` BTW to represent, I guess, no language selected. That makes sense.

Also, this standard says it represents a BCP 47 language tag, but that's not quite accurate; a developer could pass in "asdf" and there's no normative text rejecting that. You'll instead want something similar to the verbiage at https://notifications.spec.whatwg.org/#language.

marcoscaceres:
> Notably this does not allow serving up multiple options with different locales and allowing the browser to choose between them. I guess that is a non-goal.

Yeah. Non-goal I think. 

> It's also noteworthy that this matches NotificationOptions already, so that's good. Maybe we can consolidate this Localizable type into Web IDL or similar. /cc @annevk @beverloo @tobie.

I was thinking the same.... would be great to have this in Web IDL.  Maybe we can craft `Localizable` here first, and then push it to WebIDL? 

>  Notifications defaults lang to "" BTW to represent, I guess, no language selected. That makes sense.

Will fix. 

> Also, this standard says it represents a BCP 47 language tag, but that's not quite accurate; a developer could pass in "asdf" and there's no normative text rejecting that. You'll instead want something similar to the verbiage at https://notifications.spec.whatwg.org/#language.

Yeah, that's some good wording.

annevk:
Uplifting to IDL seems reasonable. We already put some shared concepts and processing models there.

tobie:
Yeah, seems this could go in [§4 Common definitions](https://heycam.github.io/webidl/#common).

ianbjacobs:
@domenic wrote:

"Notably this does not allow serving up multiple options with different locales and allowing the browser to choose between them."

If I understand your comment, it is about support for multiple strings in multiple languages.
A long time back we discussed whether the API needed to support this. Our conclusion at the time was that it did not, because people could adjust content based on global directives like HTTP language negotiation.

(If you were talking about something else; sorry for the noise.)

Ian

domenic:
Yeah, it makes sense. So this is mostly about allowing people to markup certain messages as having a different lang/dir than the rest of the content on the page.

index.html
```diff
@@ -1925,6 +1925,68 @@
...
+      <p>
+        The text-direction values are the following, implying that the value of
+        the human-readable members is by default:
+      </p>
+      <dl>
+        <dt>
+          <dfn>auto</dfn>
+        </dt>
+        <dd>
+          Directionality is determined by the [[!BIDI]] algorithm.
```
zcorpan: The bidi algorithm should be used for `ltr` and `rtl` as well... right?

I think we basically want to hook into http://www.unicode.org/reports/tr9/#Explicit_Directional_Isolates similarly to what CSS does in https://drafts.csswg.org/css-writing-modes-3/#unicode-bidi (for `isolate` or `plaintext` -- not sure which is more appropriate here).

zcorpan:
cc @w3c/i18n-reviewers

index.html
annevk: https://notifications.spec.whatwg.org/#direction has the text you need for this.

marcoscaceres: Thanks, will try to draft it soon.

index.html
domenic: I don't think it should go here if it only applies to PaymentDetailsUpdate. Instead you're going to have to just duplicate the two things (or use a partial dictionary PaymentDetailsUpdate).

If you put it here then PaymentDetailsInit gets it, and in particular that means passing `{ get dir() { throw new Error("boo"); }` to places that expect a PaymentDetailsInit must throw.

index.html
```diff
@@ -1925,6 +1925,68 @@
...
+      <pre class="idl">
+        enum TextDirection {
+          "auto",
+          "ltr",
+          "rtl"
+        };
+      </pre>
+      <p>
+        The text-direction values are the following, implying that the value of
+        the human-readable members is by default:
```
domenic: This sentence doesn't make sense to me.

domenic (changes requested):

marcoscaceres:
blocked on https://github.com/heycam/webidl/pull/358

marcoscaceres:
argh, that's not right. Seems I screwed up a rebase.

marcoscaceres:
better...

marcoscaceres:
Still blocked on discussions over in the WebIDL repo. Probably take another few days to resolve those.

index.html
```diff
@@ -2810,6 +2810,10 @@
         </dt>
         <dd>
           <p>
+            The <code><dfn>LocalizableString</dfn></code> <a data-cite=
+            "!WEBIDL-LS#dfn-typedef">typedef</a> is defined by [[!WEBIDL-LS]].
```
tobie, line 2852: You can go ahead and just use [[WEBIDL]], btw.

marcoscaceres: Will fix globally! Thanks!

zcorpan: [[WEBIDL]] is informative reference, [[!WEBIDL]] is normative reference.

tobie: I meant WEBIDL vs WEBIDL-LS, sorry if that caused confusion. :(((

marcoscaceres: It's all good. I sent https://github.com/w3c/browser-payment-api/pull/524

marcoscaceres:
Closing for now... will reopen if we want to add this in v2

PR #465, opened 2017-03-20
Editorial changes

ianbjacobs:
Some editorial changes:

 * Payment methods describe data; they are not participants in a transaction.
   Therefore, recast the description of parties as payers/payees/providers
   of payment methods. (Both in Intro and Goals)

 * The word "buyer" was used in the intro and not used elsewhere in the
   spec. Therefore, I use payer/payee which are used further down.

 * The first examples the reader should encounter in the PaymentRequest
   interface description should be those most likely found in practice.
   The "Buy" button should come first. The other examples of in-app
   and POS are much further out and so I have listed them after.

 * Updated example 2 as follows:

   - We have network id values; we should use those instead of fictitious
     ones.

   - bobpay.com does not confirm (I believe) to the PMI spec. So I have
     used an example URL instead.

***
[Preview](https://s3.amazonaws.com/pr-preview/ianbjacobs/browser-payment-api/editorial20170320.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/654d0f7...ianbjacobs:08a0dc8.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,7 +6,8 @@
     </title>
     <meta charset='utf-8'>
     <script src='https://www.w3.org/Tools/respec/respec-w3c-common' class=
-    'remove'></script>
+    'remove'>
+    </script>
     <script class='remove'>
     var respecConfig = {
       shortName: "payment-request",
@@ -123,14 +124,15 @@
         transaction:
       </p>
       <ul>
-        <li>the payee, such as a merchant with an online store,
+        <li>the payee: the merchant that runs an online store, or other party
+        that requests to be paid.
         </li>
-        <li>the payer, such as a person making a purchase on that online store,
-        and who authenticates and authorizes payment as required,
+        <li>the payer: the party that makes a purchase at that online store,
+        and who authenticates and authorizes payment as required.
         </li>
-        <li>the provider of a <dfn data-lt="payment methods">payment
-        method</dfn> (e.g., credit card) that the payer uses to pay, and that
-        is accepted by the payee.
+        <li>the <dfn data-lt="payment methods">payment method</dfn> provider:
+        the party that provides the means (e.g., credit card) that the payer
+        uses to pay, and that is accepted by the payee.
         </li>
       </ul>
       <p>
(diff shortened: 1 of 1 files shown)


marcoscaceres (changes requested):
Please run tidy over the doc

ianbjacobs:
@marcoscaceres,

I have just installed HTML Tidy version 5.5.10. What options should I run it with? I used '-i' but the result seems very different from the input file. That is: a lot more is changed by tidy than I changed via my edit. Please advise.

Ian

marcoscaceres:
@ianbjacobs, run:

```
tidy -config tidyconfig.txt -o index.html index.html
```

index.html
```diff
@@ -119,15 +119,17 @@
...
-        method</dfn> (e.g., credit card). Information necessary to process and
-        confirm a transaction is passed between the <a>payment method</a> and
-        the merchant via the <a>user agent</a> with the buyer confirming and
-        authorizing as necessary across the flow.
-      </p>
+        (e.g., browsers) to act as an intermediary between three parties in
+        a transaction:</p>
+
+      <ul>
+	<li>the payee, such as a merchant with an online store,</li>
```
marcoscaceres: I'd prefer these ended in full stop.

marcoscaceres: *each of these

index.html
```diff
@@ -119,15 +119,20 @@
...
-        online web store), and the <dfn data-lt="payment methods">payment
-        method</dfn> (e.g., credit card). Information necessary to process and
-        confirm a transaction is passed between the <a>payment method</a> and
-        the merchant via the <a>user agent</a> with the buyer confirming and
-        authorizing as necessary across the flow.
+        (e.g., browsers) to act as an intermediary between three parties in a
+        transaction:
       </p>
+      <ul>
+        <li>the payee, such as a merchant with an online store,
```
marcoscaceres: I'd still like this to be:

 * The payee: bla bla bla. 
 * The payer: bla bla bla.
 * The provider: bla bla bla.

marcoscaceres (changes requested):

marcoscaceres (approved):

PR #488, opened 2017-03-30
Use "NotAllowedError" instead of "QuotaExceededError"

### LINKED ISSUE #485: Supplement QuoteExceededError with NotAllowedError for canMakePayments
I think `QuotaExceededError` is the wrong type of error for `canMakePayments`, since the user agent can return it at its discretion - not just when a quota limit has been reached.

For example, a user agent may decide to not service a `canMakePayments` request when in a private browsing mode or similar, or may allow the user the option to decline to reveal their payment methods to the merchant.

In those cases a quota hasn't been reached, the call has simply been denied. I wonder if it would therefore be better to permit user agents to return `NotAllowedError` instead. It's somewhat of a semantic argument, but otherwise merchants could end up with a scenario when they call that API a single time and get a quota exceeded error, which doesn't make much sense.

domenic:
Fixes #485.

@nickjshearer is this what you had in mind?

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/notallowederror.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/6421d94...2428ffe.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -770,7 +770,7 @@
           any of the desired <a>payment methods</a> supplied to the
           <a>PaymentRequest</a> constructor, and false if none are supported.
           If the method is called too often, the user agent might instead
-          return a promise rejected with a "<a>QuotaExceededError</a>"
+          return a promise rejected with a "<a>NotAllowedError</a>"
           <a>DOMException</a>, at its discretion.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):
I'll note we've made this change on the Gecko side.

PR #489, opened 2017-03-30
fix: change Events section to normative (closes #466)

### LINKED ISSUE #466: Why are two sections non-normative?
I revisited PR API today. I am wondering why sections 17 (iframe) and 18.1 (summary of shipping events) are non-normative.

For the iframe section, it seems to me more than non-normative if we are saying "here's how you do this."

It's not clear to me what the purpose of 18.1 is. There is normative text elsewhere about the two events. Does that not suffice?

marcoscaceres:
***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/normative_sections.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/0838825...dbf3c3c.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1942,12 +1942,12 @@
               Interface
             </th>
             <th>
-              Dispatched when...
+              Dispatched when…
             </th>
           </tr>
           <tr>
             <td>
-              <code>shippingaddresschange</code>
+              <code><dfn>shippingaddresschange</dfn></code>
             </td>
             <td>
               <a>PaymentRequestUpdateEvent</a>
@@ -1958,7 +1958,7 @@
           </tr>
           <tr>
             <td>
-              <code>shippingoptionchange</code>
+              <code><dfn>shippingoptionchange</dfn></code>
             </td>
             <td>
               <a>PaymentRequestUpdateEvent</a>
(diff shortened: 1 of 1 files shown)


domenic:
What normative content does this section contain?

marcoscaceres:
the event summary.

marcoscaceres:
(see the preview)

marcoscaceres:
Oh, I just totally realized that the event names are defined in the wrong place! They should be "dfn"'ed here.

domenic:
What normative content does the summary contain?

Let's put it this way: what tests could you write for the contents of this section?

marcoscaceres:
The string that is the type of the event.

domenic:
That is already defined in the sections that fire the event (19.1 and 19.2).

Note that these event interfaces can be used with arbitrary types, e.g. `new PaymentRequestUpdateEvent("foobar")`.

marcoscaceres:
I know what you are getting at, but there is a balance to be struck here... right now, the spec has:

> `Let <var>name</var> be <dfn>shippingoptionchange</dfn>.`

That's not a definition. I'd rather have a conceptual definition for shippingoptionchange in the summary table.

domenic:
How would you write tests for that definition? It's just a non-normative conceptualization that developers can use to reason about when it fires, but when it fires is already normatively specified elsewhere.

domenic:
If it helps, https://html.spec.whatwg.org/#events-2 is non-normative also (and it contains even more information).

marcoscaceres:
The spec is also inconsistent, because elsewhere:

> `<li>Let <var>name</var> be "<a>shippingaddresschange</a>".` 

> If it helps, https://html.spec.whatwg.org/#events-2 is non-normative also (and it contains even more information).

Ok, so let's go with non-normative - but have the `dfn`s in the table. Yes?

domenic:
Works for me!

domenic (approved):

PR #492, opened 2017-04-01
Add extension point

### LINKED ISSUE #470: Add explicit extension point to matching algo
^^

adrianhopebailie:
Added extension point for payment app matching.
Based on https://www.w3.org/TR/appmanifest/#dfn-extension-point

Resolves #470

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/extension-point.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/3a79df3...81bc51e.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -623,13 +623,6 @@
           hook themselves into this specification at this point in the
           algorithm.
         </p>
-        <div class="note">
-          <p>
-            The <a>extension point</a> is meant to help avoid <a href=
-            "https://annevankesteren.nl/2014/02/monkey-patch">issues related to
-            monkey patching</a>.
-          </p>
-        </div>
         <p>
           The <a>show()</a> method MUST act as follows:
         </p>
@@ -679,22 +672,15 @@
           <li>Return <var>acceptPromise</var> and perform the remaining steps
           <a>in parallel</a>.
           </li>
-          <li><p><a>Extension point</a>: process any proprietary and/or other
-            supported algorithms, for determining the available methods of paying,
-            at this point in the algorithm.</p>
-            <p>Alternatively, the following generic algorithm may be applied:</p>
-            <p>For each <var>paymentMethod</var> in <var>request</var>.
-            <a>[[\serializedMethodData]]</a>:
-              <ol>
-                <li>Determine which <a>payment apps</a> support any of the
-                <a>payment method identifiers</a> given by the first element of
-                the <var>paymentMethod</var> tuple. For each resulting payment
-                app, if payment method specific capabilities supplied by the
-                payment app match those provided by the second element of the
-                tuple, the payment app matches.
-                </li>
-              </ol>
-            </p>
+          <li>
+            <p><a>Extension point</a>: process algorithms for determining the 
+          available methods of paying at this point.</p>
+            <p>The algorithms should process each item in 
+              <var>request</var>.<a>[[\serializedMethodData]]</a> and determine 
+            the set of <a>payment apps</a>, with the appropriate capabilities, 
+            that is available to fulfill the payment request.</p>
+            <p>The output of executing these algorithms is a list of <a>payment apps</a>,
+            and any other methods of paying, able to fulfill the payment request.</p>
           </li>
           <li>If this consultation produced no supported method of paying, then
           reject <var>acceptPromise</var> with a "<a>NotSupportedError</a>" <a>
@@ -704,12 +690,11 @@
           <li>
             <p>
               Otherwise, show a user interface to allow the user to interact
-              with the payment request process, using those <a>payment apps</a>
-              and <a>payment methods</a> which the above step identified as
-              feasible. The user agent MAY show payment methods in the order
-              given by <var>supportedMethods</var>, but SHOULD prioritize the
-              preference of the user when presenting payment methods and
-              applications.
+              with the payment request process, using those methods of paying
+              which the above step identified as supported. The user agent MAY 
+              show these methods of paying in the order given by 
+              <var>supportedMethods</var>, but SHOULD prioritize the
+              preference of the user.
             </p>
             <p>
               The <a>payment app</a> should be sent the appropriate data from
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -617,6 +617,20 @@
           </p>
         </div>
         <p>
+          The following algorithm provides an <dfn>extension point</dfn>: other
+          specifications that define more specific algorithms for matching the available 
+          <a>payment apps</a> to the provided <a>payment methods</a> are encouraged to
+          hook themselves into this specification at this point in the
+          algorithm.
+        </p>
+        <div class="note">
```
marcoscaceres: I suggest we remove this. I know I have it in the manifest spec, but I added this there originally because we were trying to understand the implications and process of monkey patching things back in 2014 - specially across disparate specs and standards orgs. Anne's post was certainly helpful in providing guidance. 

In our case, we are basically adding an extension point for specs that we will work on just within this working group.

index.html
```diff
@@ -665,17 +679,22 @@
...
+              <ol>
+                <li>Determine which <a>payment apps</a> support any of the
+                <a>payment method identifiers</a> given by the first element of
+                the <var>paymentMethod</var> tuple. For each resulting payment
+                app, if payment method specific capabilities supplied by the
+                payment app match those provided by the second element of the
+                tuple, the payment app matches.
+                </li>
+              </ol>
+            </p>
```
marcoscaceres: This markup is invalid. You can't have a `ol` inside a `p`.

index.html
```diff
@@ -665,17 +679,22 @@
...
-              the <var>paymentMethod</var> tuple. For each resulting payment
-              app, if payment method specific capabilities supplied by the
-              payment app match those provided by the second element of the
-              tuple, the payment app matches.
-              </li>
-            </ol>
+          <li><p><a>Extension point</a>: process any proprietary and/or other
+            supported algorithms, for determining the available methods of paying,
+            at this point in the algorithm.</p>
+            <p>Alternatively, the following generic algorithm may be applied:</p>
```
marcoscaceres: There can't be an "alternatively". Otherwise it reads as:
```
if (extension point) {
  // 
} else{
  // alternatively
}
```

What we want is:

```
try {
  async for (const extension  of extensions) {
    await extension.doYourThing();
  }  
} catch(err){
   return reject(err);
}
// do stuff that is already in the spec
// this spec always wins, because extensions points must not override default behavior 
```

marcoscaceres (changes requested):

index.html
marcoscaceres: Consider also that the extension point might require particular inputs and might have outputs (probably exceptions, as above). I've not given a lot of thoughts to what those are. That extension interface is basically what we would want to standardize here.

A place to start for the above is to look at which spec needs this right now. Which spec is going to use this?

index.html
ianbjacobs: +1 to Marcos' suggestion

ianbjacobs:
@adrianhopebailie and @marcoscaceres,

Here is an alternative to Adrian's proposal (and something similar would be included in the canMakePayment() algorithm):

10. For each paymentMethod in request.[[serializedMethodData]]:

    1. Determine which payment apps support any of the payment method identifiers given by the first element of the paymentMethod tuple. For each resulting payment app, if payment method specific capabilities supplied by the payment app match those provided by the second element of the tuple, the payment app matches. Capability matching constitutes an extension point in this specification. User agents will match according to payment app integration approaches and payment method capabilities.

adrianhopebailie:
@ianbjacobs and @marcoscaceres  - I have updated the text to standardize on the terminology "method of paying" to describe payment apps and any other method of paying.

I think this makes the test more readable and consistent.

marcoscaceres:
@ianbjacobs, your proposed text seems pretty good - but this doesn't seem to add much, so suggest dropping it:

> User agents will match according to payment app integration approaches and payment method capabilities.

Will discuss with @zkoch on this week's editor's call.

marcoscaceres:
Where does payment handler hook into this?

adrianba:
The current algorithm should work without needing an extension point here. It doesn't matter where payment method implementations come from, identifying the available ones is handled correctly.

adrianhopebailie:
@adrianba - The purpose of the extension point is to allow another spec to describe payment method specific filters like "SupportedNetworks" for basic card.

We don't want to get into that detail in PR but it leaves it open for payment method specs to define these filters and ways that payment handlers can specify capabilities they have that would result in them being candidates for user selection.

marcoscaceres:
Personally, I think it's a mistake to add these things preemptively. I think we should only add them when we actually have a spec that needs it. We risk adding text that might not actually work for the intended purpose in practice.

adrianhopebailie:
I understood the basic-card payment method spec to need this already.

marcoscaceres:
I don't think so. I'm hooking in somewhere else.

adrianhopebailie:
Where is the algorithm/logic that explains how to filter cards based on `supportedNetworks` and `supportedTypes` captured?

My understanding was that this would be captured in the basic-card payment method spec as an extension to the PR matching algo (i.e. explicitly using that extension point)

marcoscaceres:
Sorry, made a mistake. I was thinking of a different spec. Let me check the basic card spec (though my position stands).

ianbjacobs:
Three payment method specifications so far (Basic Card, Tokenized Card, Credit Transfer) involve filters. The decision was to have browsers do the filter matching, but to leave the spec underspecified as to how that happens. We have proposed concrete algorithms for string set matching (so that we would know moving forward exactly what behavior to rely on) but the preference of implementers was to not specify an algorithm at this time.

The key conclusions we reached are: filter/capability matching will happen in the browser, that payment handler API will talk about how to register capabilities with the browser, and that W3C payment method spec writers can define filters.

adrianhopebailie:
@adrianba will propose an alternative

PR #512, opened 2017-04-27
Reject with SecurityError if the UA rejects the call to show().

### LINKED ISSUE #486: should .show() be user gated?
Followup from #462. 

If the browser crashes and restarts by reopening N number of tabs. A tab in the background calls `.show()` and it now blocks all payment requests (at least, for a time - assuming it does not focus the tab, which is also not in the spec).

To protect from the above, should we then make it that `.show()` is gated on user interaction?

adrianba:
Fixes #486.

***
[Preview](https://s3.amazonaws.com/pr-preview/adrianba/browser-payment-api/issue486.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/e0cf815...adrianba:cd50c27.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -623,6 +623,24 @@
           <li>Let <var>request</var> be the <a>PaymentRequest</a> object on
           which the method is called.
           </li>
+          <li>
+            <p>
+              If the <a>user agent</a> rejects the call to <a>show()</a> to
+              protect the user, then return a promise rejected with a
+              "<a>SecurityError</a>" <a>DOMException</a>. For example, this
+              might occur if the <a>user agent</a> requires a user gesture to
+              allow a page to call <a>show()</a>, or the <a>user agent</a> may
+              limit the rate at which a page can call <a>show()</a>, as
+              described in the the <a href="#privacy">privacy
+              considerations</a> section.
+            </p>
+            <p class="note">
+              Implementations are expected to experiment in this area.
+              Developers using the payment request API should investigate and
+              anticipate such experiments and understand under what
+              circumstances a <a>SecurityError</a> might occur.
+            </p>
+          </li>
           <li>If <var>request</var>.<a>[[\state]]</a> is not "<a>created</a>"
           then return a promise rejected with an "<a>InvalidStateError</a>" <a>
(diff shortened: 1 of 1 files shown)


domenic:
I have some editorial tweaks; @adrianba could you check the box "allow edits from maintainers"? See https://help.github.com/assets/images/help/pull_requests/allow-maintainers-to-make-edits-sidebar-checkbox.png

Note that unlike the "AbortError" case below, this does not close down the PaymentRequest and prevent further use of it. That seems good and intentional, but I wanted to make sure people were aware of the difference.

marcoscaceres (changes requested):
@adrianba, see @domenic's comments. Would like to see his tweaks before adding mine.

marcoscaceres:
@domenic, should be good now. @adrianba clicked the box.

domenic (approved):
My tweaks applied; over to Marcos

index.html
```diff
@@ -623,6 +623,24 @@
...
+            <p>
+              Optionally, if the <a>user agent</a> wishes to disallow the call
+              to <a>show()</a> to protect the user, then return a promise
+              rejected with a "<a>SecurityError</a>" <a>DOMException</a>. For
+              example, the <a>user agent</a> may require the call to be
+              <a>triggered by user activation</a>, or may limit the rate at
+              which a page can call <a>show()</a>, as described in the the
+              <a href="#privacy">privacy considerations</a> section.
+            </p>
+            <p>
```
marcoscaceres, line 636: This should be a Note, stating that: "During the Candidate Recommendation phase, implementations are expected to experiment in this area..." 

And concluding with: "If interoperable behavior emerges amongst user agents, then that behavior will be standardized here before progressing the specification along the W3C Recommendation Track."

marcoscaceres (changes requested):
One small change.

marcoscaceres:
@adrianba please check the box again if you want me to make the modifications I suggested.

marcoscaceres (approved):
Approving and merging to avoid bit-rot. Will add the suggestions I made in a new PR.

PR #515, opened 2017-04-27
editorial: use DOM relatedTarget instead of .target (closes #513)

### LINKED ISSUE #513: Editorial: event `.target` attribute should be relatedTarget
As per the DOM spec, I believe we need to talk about the relatedTarget of the event (as opposed to the value of the target attribute).   

Affects text in `updateWith()` and possibly other places where we perform target type checks.

marcoscaceres:
***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/relatedTarget.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/fbb526e...5bc3ac4.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2034,7 +2034,7 @@
             instance.
             </li>
             <li>Let <var>target</var> be the value of <var>event</var>'s
-            <a data-cite="DOM#dom-event-target">target</a> attribute.
+            <a data-cite="!DOM#event-relatedtarget">relatedTarget</a>.
             </li>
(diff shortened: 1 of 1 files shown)


domenic:
This seems definitely wrong. Only UIEvents defines relatedTarget for its events; it never appears for shippingaddresschange/shippingoptionchange.

marcoscaceres:
@domenic my bad. I misread the dispatch algorithm in the DOM Spec :(

PR #536, opened 2017-05-25
Reject acceptedPromise if converting method data fails

marcoscaceres:
Align with https://github.com/w3c/webpayments-methods-card/pull/23/

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/conversion.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/a3c35fb...5896b12.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,8 +6,7 @@
     </title>
     <meta charset='utf-8'>
     <script src='https://www.w3.org/Tools/respec/respec-w3c-common' class=
-    'remove'>
-    </script>
+    'remove'></script>
     <script class='remove'>
     var respecConfig = {
       github: "https://github.com/w3c/browser-payment-api/",
@@ -704,11 +703,11 @@
               <var>acceptPromise</var> with that error and terminate this
               algorithm.
               </li>
-              <li>Determine which <a>payment handlers</a> support any of the
-              <a>payment method identifiers</a> given <var>identifier</var>.
-              For each resulting payment handler, if payment method specific
-              capabilities supplied by the payment handler match those provided
-              by <var>data</var>, the payment handler matches.
+              <li>Determine which <a>payment handlers</a> support
+              <var>identifier</var>. For each resulting payment handler, if
+              payment method specific capabilities supplied by the payment
+              handler match those provided by <var>data</var>, the payment
+              handler matches.
               </li>
             </ol>
           </li>
--- a/index.html
+++ b/index.html
@@ -6,7 +6,8 @@
     </title>
     <meta charset='utf-8'>
     <script src='https://www.w3.org/Tools/respec/respec-w3c-common' class=
-    'remove'></script>
+    'remove'>
+    </script>
     <script class='remove'>
     var respecConfig = {
       github: "https://github.com/w3c/browser-payment-api/",
@@ -685,8 +686,8 @@
           <li>For each <var>paymentMethod</var> tuple in
           <var>request</var>.<a>[[\serializedMethodData]]</a>:
             <ol>
-              <li>Let <var>identifier</var> be the first element in the <var>
-                paymentMethod</var> tuple.
+              <li>Let <var>identifier</var> be the first element in the
+              <var>paymentMethod</var> tuple.
               </li>
               <li>Let <var>data</var> be the result of <a data-cite=
               "!ECMA-262-2015#sec-json.parse">JSON-parsing</a> the second
@@ -703,9 +704,11 @@
               <var>acceptPromise</var> with that error and terminate this
               algorithm.
               </li>
-              <li>If payment method specific capabilities supplied by the
-              payment handler match those provided by <var>data</var>, the
-              payment handler matches.
+              <li>Determine which <a>payment handlers</a> support any of the
+              <a>payment method identifiers</a> given <var>identifier</var>.
+              For each resulting payment handler, if payment method specific
+              capabilities supplied by the payment handler match those provided
+              by <var>data</var>, the payment handler matches.
               </li>
             </ol>
           </li>
--- a/index.html
+++ b/index.html
@@ -700,8 +700,8 @@
                   <li>If required by the specification that defines the
                   <var>handler</var>, <a data-cite=
                   "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
-                  <var>data</var> to an IDL value. If it results in a error,
-                  reject <var>acceptPromise</var> with that error an terminate
+                  <var>data</var> to an IDL value. If it results in an error,
+                  reject <var>acceptPromise</var> with that error and terminate
                   this algorithm.
                   </li>
                   <li>If payment method specific capabilities supplied by the
--- a/index.html
+++ b/index.html
@@ -693,9 +693,20 @@
               </li>
               <li>Determine which <a>payment handlers</a> support any of the
               <a>payment method identifiers</a> given <var>identifiers</var>.
-              For each resulting payment handler, if payment method specific
-              capabilities supplied by the payment handler match those provided
-              by <var>data</var>, the payment handler matches.
+              </li>
+              <li>For each <var>handler</var> of the resulting payment
+              handlers:
+                <ol>
+                  <li>If required by the specification that defines the
+                  <var>handler</var>, <a data-cite=
+                  "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+                  <var>data</var> to an IDL value. Rethrow any exceptions.
+                  </li>
+                  <li>If payment method specific capabilities supplied by the
+                  payment handler match those provided by <var>data</var>, the
+                  payment handler matches.
+                  </li>
+                </ol>
               </li>
             </ol>
           </li>
--- a/index.html
+++ b/index.html
@@ -688,21 +688,25 @@
               <li>Let <var>identifiers</var> be the first element in the <var>
                 paymentMethod</var> tuple.
               </li>
-              <li>Let <var>data</var> be the second element in
-              <var>paymentMethod</var> tuple.
+              <li>Let <var>data</var> be the result of <a data-cite=
+              "!ECMA-262-2015#sec-json.parse">JSON-parsing</a> the second
+              element in <var>paymentMethod</var> tuple.
               </li>
-              <li>Determine which <a>payment handlers</a> support any of the
-              <a>payment method identifiers</a> given <var>identifiers</var>.
+              <li>Let <var>methods</var> be a list of <a>payment methods</a>
+              that support <var>identifiers</var>.
               </li>
-              <li>For each <var>handler</var> of the resulting payment
-              handlers:
+              <li>For each <var>paymentMethod</var> of <var>methods</var>:
                 <ol>
                   <li>If required by the specification that defines the
-                  <var>handler</var>, <a data-cite=
+                  <var>paymentMethod</var>, then <a data-cite=
                   "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
-                  <var>data</var> to an IDL value. If it results in an error,
-                  reject <var>acceptPromise</var> with that error and terminate
-                  this algorithm.
+                  <var>data</var> to an IDL value. Otherwise, <a data-cite=
+                  "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a> to
+                  <a data-cite="!WEBIDL#idl-object">object</a>.
+                  </li>
+                  <li>If conversion results in an error, reject
+                  <var>acceptPromise</var> with that error and terminate this
+                  algorithm.
                   </li>
                   <li>If payment method specific capabilities supplied by the
                   payment handler match those provided by <var>data</var>, the
--- a/index.html
+++ b/index.html
@@ -700,7 +700,9 @@
                   <li>If required by the specification that defines the
                   <var>handler</var>, <a data-cite=
                   "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
-                  <var>data</var> to an IDL value. Rethrow any exceptions.
+                  <var>data</var> to an IDL value. If it results in a error,
+                  reject <var>acceptPromise</var> with that error an terminate
+                  this algorithm.
                   </li>
                   <li>If payment method specific capabilities supplied by the
                   payment handler match those provided by <var>data</var>, the
--- a/index.html
+++ b/index.html
@@ -685,34 +685,27 @@
           <li>For each <var>paymentMethod</var> tuple in
           <var>request</var>.<a>[[\serializedMethodData]]</a>:
             <ol>
-              <li>Let <var>identifiers</var> be the first element in the <var>
+              <li>Let <var>identifier</var> be the first element in the <var>
                 paymentMethod</var> tuple.
               </li>
               <li>Let <var>data</var> be the result of <a data-cite=
               "!ECMA-262-2015#sec-json.parse">JSON-parsing</a> the second
               element in <var>paymentMethod</var> tuple.
               </li>
-              <li>Let <var>methods</var> be a list of <a>payment methods</a>
-              that support <var>identifiers</var>.
+              <li>If required by the specification that defines the
+              <var>identifer</var>, then <a data-cite=
+              "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+              <var>data</var> to an IDL value. Otherwise, <a data-cite=
+              "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a> to
+              <a data-cite="!WEBIDL#idl-object">object</a>.
               </li>
-              <li>For each <var>paymentMethod</var> of <var>methods</var>:
-                <ol>
-                  <li>If required by the specification that defines the
-                  <var>paymentMethod</var>, then <a data-cite=
-                  "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
-                  <var>data</var> to an IDL value. Otherwise, <a data-cite=
-                  "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a> to
-                  <a data-cite="!WEBIDL#idl-object">object</a>.
-                  </li>
-                  <li>If conversion results in an error, reject
-                  <var>acceptPromise</var> with that error and terminate this
-                  algorithm.
-                  </li>
-                  <li>If payment method specific capabilities supplied by the
-                  payment handler match those provided by <var>data</var>, the
-                  payment handler matches.
-                  </li>
-                </ol>
+              <li>If conversion results in an error, reject
+              <var>acceptPromise</var> with that error and terminate this
+              algorithm.
+              </li>
+              <li>If payment method specific capabilities supplied by the
+              payment handler match those provided by <var>data</var>, the
+              payment handler matches.
               </li>
             </ol>
           </li>


index.html
```diff
@@ -693,9 +693,20 @@
...
-              capabilities supplied by the payment handler match those provided
-              by <var>data</var>, the payment handler matches.
+              </li>
+              <li>For each <var>handler</var> of the resulting payment
+              handlers:
+                <ol>
+                  <li>If required by the specification that defines the
+                  <var>handler</var>, <a data-cite=
+                  "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+                  <var>data</var> to an IDL value. Rethrow any exceptions.
```
domenic: You can't throw an exception from in parallel.

It seems unlikely we want to error if the user installs a payment handler which doesn't accept the provided _data_. Maybe if *none* of their installed handlers do... but that's already covered by the next step.

domenic (changes requested):

index.html
marcoscaceres: > You can't throw an exception from in parallel.

Ah yeah, will reject acceptPromise. 

> It seems unlikely we want to error if the user installs a payment handler which doesn't accept the provided data. 

Sure... but... 

> Maybe if none of their installed handlers do... but that's already covered by the next step.

Ok, but something bad needs to happen with invalid data? I'm confused about where that should go.

domenic: Sure, if the data is invalid for *all* payment apps, then the next step (which already exists) will fail. But we need to give every payment app a chance. If one succeeds, but 3 fail, it's not invalid data.

marcoscaceres:
To be clear, trying to solve for:

```JS 
const methodData = [{
  supportedMethods: ["basic-card", "https://foo.com/pay"],
  data: { supportedTypes: ['THIS IS NOT A VALID ENUM VALUE!!'] }
}];
try {
 // When converting data, should `supportedTypes` cause `acceptPromise` to reject?
} catch(err) {

}
```

domenic:
That is already rejected before this PR, since no payment handlers support that data, so the next step takes care of rejecting the promise.

What this PR starts rejecting is:

```js
const methodData = [{
  supportedMethods: ["basic-card", "bitcoin"],
  data: { supportedTypes: "credit" }
};
```

because "bitcoin" doesn't support "credit".

Is that intentional? It seems wrong to me; instead it makes more sense to me to do what the spec currently says, and display UI for paying with basic-card/credit, and ignore the bitcoin supported method.

EDIT: switched from "visa" to "credit" (the latter is actually a valid supportedType for basic-card).

index.html
marcoscaceres: Ok, got you. So we just ignore bad conversions as "unsupported". I guess as long as we notify the developer of bad conversions in the error console, that should be ok (or else it's gonna be hard to debug).

marcoscaceres:
this still feels super messed up to me. You could potentially end up with completely incompatible data types and property name clashes. 

```
const methodData = [{
  supportedMethods: ["basic-card", "bitcoin"],
  data: { supportedTypes: "bitcoin-specific" }
};
```

domenic:
Yeah, definitely. I don't understand why supportedMethods is an array at all; it seems strange to think that the same `data` would make sense for multiple methods. Usually it doesn't, I assume, so that's why it doesn't make sense to me to error on mismatch, because it seems like any time you have 2+ supportedMethods, you're always going to get at least one mismatch.

In other words, given that we have supportedMethods being an array, the current spec seems better than requiring `data` to work for all supportedMethods.

marcoscaceres:
> Usually it doesn't, I assume, so that's why it doesn't make sense to me to error on mismatch, because it seems like any time you have 2+ supportedMethods, you're always going to get at least one mismatch.

I can only assume this was done with the assumption that `.data` would be the same type for the array of `supportedMethods` - even if potentially false.  

Ideally, you then want: 

```JS
const methodData = [{
  supportedMethods: ["basic-card", "https://bank.com/basic-card"],
  data: { supportedTypes: "visa" }
}, {
  supportedMethods: ["crypto-currency", "https://bit.com/coin"],
  data: { cryptoCurrencyThing: "thing" }
}];
```

So you don't mix types. If you mix data types, this is going to all come crashing down.

marcoscaceres:
@domenic, the more I think I about it, the more I think we should reject `acceptPromise` if there is a `.data` type mismatch during conversion. 

I can't think of any valid use case for mixing `.data` types.

marcoscaceres:
For those watching at home, the following would be all fine: 

```JS
const methodData = [{
  // Must match "basic-card" spec, undergoes conversion
  supportedMethods: ["https://bank.com/any-type", "basic-card"],
  // .data MUST BE `BasicCardRequest`, because "basic-card" present. 
  data: { supportedTypes: ["visa"] } 
}, {
  // Must match "tokenized-thing" spec, undergoes conversion
  supportedMethods: ["tokenized-thing", "https://bit.com/coin", "https://bank.com/any-type"],
  data: { cryptoCurrencyThing: "thing" }
}, {
  // .data treated as a normal object, no dictionary conversion. 
  supportedMethods: ["https://foo.com/bar", "https://bank.com/any-type"],
  data: { foo: "bar", "thumb": "drive", anything: "goes" }
}];
```

But:

```JS
const methodData = [
  // This would cause a rejection, iff "future-card" is supported:
  {
    supportedMethods: ["https://foo.com/future-card", "future-card"],
    data: { supportedTypes: ['THIS IS NOT A VALID VALUE!!'] }
  }, 
  // iff "future-card" is not supported, this one is still ok:
  {
    supportedMethods: ["https://foo.com/tokenized-thing", "tokenized-thing"],
    data: {   tokensAreFancy: true   }
  },
  {
   // This also throws
    supportedMethods: ["basic-card", "nonbasic-card"],
    data: { supportedTypes: "credit" }
  } 
]
```

adrianhopebailie:
@marcoscaceres and @domenic this was discussed at some length on a WG call a few weeks ago and has been a niggling issue since pretty much the first payment method spec was written.

My understanding from @rsolomakhin and @zkoch was that the data is treated as an opaque object and no validation is done against the IDL in the payment method spec.

The browser passes the opaque data onto the payment handler which then parses it and uses it. If it is invalid then the response from the handler will indicate this.

The `basic-card` payment method is "special" because the handler is internal to the browser but the implementation of how the data is validated shouldn't be exposed in the Payment Request spec.

Before you say: "But we need to look at the filters like `supportedTypes` before the data is passed to the handler". I know.

In an ideal world each `PaymentMethodData` would have:
1. A set of identifiers it applies to
2. A blob of opaque data that the browser ignores and simply passes to the handler
3. A set of filters used to filter handlers based on their capabilities

The proposal to add a new member `filters` was rejected on the basis that Chrome have already deployed this and moving `supportedTypes` from `data` to `filters` is too disruptive.

Another way to solve this is that we are revisiting the idea of handlers defining a function which is presented with the opaque `data` and can respond `true`/`false` if they will be able to handle the payment.

marcoscaceres:
> On 25 May 2017, at 7:16 pm, Adrian Hope-Bailie <notifications@github.com> wrote:
> 
> @marcoscaceres and @domenic this was discussed at some length on a WG call a few weeks ago and has been a niggling issue since pretty much the first payment method spec was written.
> 
> My understanding from @rsolomakhin and @zkoch was that the data is treated as an opaque object and no validation is done against the IDL in the payment method spec.
> 

This is not correct. Please see:
https://github.com/w3c/webpayments-methods-card/pull/23

If we don't want IDL conversion, then why would we use WebIDL? 

WebIDL is not a data-structure definition language: it's a system for type coercion and moving things between c++ and js. Thus, we can't say on one hand "oh! It's Web IDL..." but then "it's an opaque blob". 

If we really don't want WebIDL, then we should not use it as the basis for BasicCardRequest.

> The browser passes the opaque data onto the payment handler which then parses it and uses it. If it is invalid then the response from the handler will indicate this.
> 

Again, this doesn't match what basic card spec says. 

> The basic-card payment method is "special" because the handler is internal to the browser but the implementation of how the data is validated shouldn't be exposed in the Payment Request spec.
> 

I'm sorry, but again you are mistaken. It's not special and it's not treated as special. Anything can act as a "basic-card" handler (including proprietary payment methods denoted by a URL) if registered to handle that type.

Also, if it's expected to be a basicCardRequest, then it must be treated as WebIDL or we should stop using WebIDL to represent BasicCardRequest. 

> Before you say: "But we need to look at the filters like supportedTypes before the data is passed to the handler". I know.
> 
I don't think this is where we were going. 

> In an ideal world each PaymentMethodData would have:
> 
> A set of identifiers it applies to
> A blob of opaque data that the browser ignores and simply passes to the handler
This seems like a massive security risk. Passing arbitrary data to apps that are expecting well-structured payment requests seems extremely bad. 

I don't think I would want to do that. 

> A set of filters used to filter handlers based on their capabilities

This doesn't match the spec. 

> The proposal to add a new member filters was rejected on the basis that Chrome have already deployed this and moving supportedTypes from data to filters is too disruptive.
> 
> Another way to solve this is that we are revisiting the idea of handlers defining a function which is presented with the opaque data and can respond true/false if they will be able to handle the payment.
> 

No, we are fine to work with the current structure: we just need to define some simple rules to avoid the insanity of arbitrary .data types (at least, that's my opinion). That's why I'm proposing having .data be converted to the PMI's data type (when not null).

domenic:
Neither of the two cases you state are the ones I am worried about. The one I am worried about is

```js
const methodData = [{
  supportedMethods: ["basic-card", "nonbasic-card"],
  data: { supportedTypes: "credit" }
}];
```

where nonbasic-card does not support the "credit" type (instead e.g. only "debit" or "giftcard").

I'd like the WG to discuss and pick one of the following:

1. If data applies to any of the supportedMethods, then show() works, presenting a UI for the subset of supportedMethods that accept the data. (Current spec)
1. Only if data applies to *all* of the supportedMethods does show() work; otherwise it rejects. (Your proposal in this PR)

If a position on this was previously documented please link me to it :). I don't otherwise feel comfortable changing this in an "editorial" PR.

adrianhopebailie:
> This is not correct. Please see:
> w3c/webpayments-methods-card#23

I already commented there. You are currently proposing a lot of spec changes with no context of the conversations that have already gone on relating to this topic. My request is to understand the reasons behind the current design before you merge this in and change it.

> If we don't want IDL conversion, then why would we use WebIDL?

That is what I asked a long time ago. But at the time nobody could see the impact of the decision to stick with WebIDL so the issue went stale and was closed by @ianbjacobs 

Please have a look at: https://github.com/w3c/browser-payment-api/issues/132

> WebIDL is not a data-structure definition language: it's a system for type coercion and moving things between c++ and js. Thus, we can't say on one hand "oh! It's Web IDL..." but then "it's an opaque blob".

> If we really don't want WebIDL, then we should not use it as the basis for BasicCardRequest.

Exactly! Where were you when I logged the above issue over a year ago? :smile:

Note that the only arguments in that thread in favour of using Web IDL essentially boil down to: "It's a nice data-structure definition language".

>> The browser passes the opaque data onto the payment handler which then parses it and uses it. If it is invalid then the response from the handler will indicate this.

> Again, this doesn't match what basic card spec says.

I'm not sure I follow. The basic card spec says nothing about how the browser handles this data.

>> The basic-card payment method is "special" because the handler is internal to the browser but the implementation of how the data is validated shouldn't be exposed in the Payment Request spec.

> I'm sorry, but again you are mistaken. It's not special and it's not treated as special. Anything can act as a "basic-card" handler (including proprietary payment methods denoted by a URL) if registered to handle that type.

Actually I think you are wrong.

How would a "proprietary payment methods denoted by a URL" be registered to handle that type? There is no way for the browser to associate the URL with the type at run-time.

Basic card is unique as it's the only payment method for which browsers ship with built in handlers. That is not expected to be the case for all payment methods that are published by the WG (whether they use WebIDL or not). If it is then that is news to me and we should all agree that is the intent and understand the implications of that.

My assumption to this point has been the only justification for using WebIDL in the basic card spec is because there is a built in handler for it. I.e. it will allow browser vendors to generate bindings between the submitted JS and the code that processes the request (the internal handler) .

> Also, if it's expected to be a basicCardRequest, then it must be treated as WebIDL or we should stop using WebIDL to represent BasicCardRequest.

I think we should stop using WebIDL, I'm yet to hear a good argument why we are besides that it's easy to write specs with in ReSpec. As I suggested in that issue I think we should follow the data modeling language used for web app manifest.

> This seems like a massive security risk. Passing arbitrary data to apps that are expecting well-structured payment requests seems extremely bad.

> I don't think I would want to do that.

So are you suggesting that browsers will not accept any new payment methods in the ecosystem for which they haven't shipped built-in data validation?

That's the opposite of the goal of this system. 

>> A set of filters used to filter handlers based on their capabilities
>This doesn't match the spec.

Correct, like I said it was a proposal that was rejected because the implementation that have shipped don't work that way.

> No, we are fine to work with the current structure: we just need to define some simple rules to avoid the insanity of arbitrary .data types (at least, that's my opinion). That's why I'm proposing having .data be converted to the PMI's data type (when not null).

That simply won't work. The whole point of `.data` is that anyone can invent a new payment method with whatever data model they want in there. If they need browsers to ship new data validation for it then the whole design is broken.

marcoscaceres:
My position is 2. 

> I don't otherwise feel comfortable changing this in an "editorial" PR.

I thought this was "editorial" because I thought we were just not clear in the behaviour - and because the PR from the other spec had not been merged. I didn't realize there would be disagreement on something that's clearly broken.

dlongley:
@adrianhopebailie said:

> I think we should stop using WebIDL, I'm yet to hear a good argument why we are besides that it's easy to write specs with in ReSpec. As I suggested in that issue I think we should follow the data modeling language used for web app manifest.

> So are you suggesting that browsers will not accept any new payment methods in the ecosystem for which they haven't shipped built-in data validation?

> That's the opposite of the goal of this system.

> That simply won't work. The whole point of .data is that anyone can invent a new payment method with whatever data model they want in there. If they need browsers to ship new data validation for it then the whole design is broken.

I agree with @adrianhopebailie here as I did a long while back when this first came up in the WG. This is an important point -- and I believe Web IDL only leaked in here because the browsers were special casing basic-card because they were implementing it directly. This was an issue long ago in the WG where, as I recall, @adrianhopebailie, myself, and a few others were on the losing side of the argument but we foresaw saw this and other extensibility concerns (as mentioned briefly by @adrianhopebailie above).

I don't know what this means moving forward for telling browsers what to do here at this point, but I would very much like to see this information left as opaque as possible with respect to the browser implementation. Whatever we do, the key point is this: we do not want to create an environment where browsers must be upgraded in order to support new payment methods.

marcoscaceres:
> On 27 May 2017, at 12:56 am, Dave Longley <notifications@github.com> wrote:
> 
> @adrianhopebailie said:
> 
> I think we should stop using WebIDL, I'm yet to hear a good argument why we are besides that it's easy to write specs with in ReSpec. As I suggested in that issue I think we should follow the data modeling language used for web app manifest.
> 
> So are you suggesting that browsers will not accept any new payment methods in the ecosystem for which they haven't shipped built-in data validation?
> 

Of course not. 

The data type for URL identifiers is object. For standardized short strings it is a WebIDL dictionary. 

I don't see what's so hard to understand. It's what we've had all along.

> That's the opposite of the goal of this system.
> 
Of course - I never suggested that. That would be ridiculous. 

> That simply won't work. The whole point of .data is that anyone can invent a new payment method with whatever data model they want in there. If they need browsers to ship new data validation for it then the whole design is broken.
> 

I never suggested otherwise. I don't know how Adrian would reach this absurd conclusion. 
> I agree with @adrianhopebailie here as I did a long while back when this first came up in the WG. This is an important point -- and I believe Web IDL only leaked in here because the browsers were special casing basic-card because they were implementing it directly. This was an issue long ago in the WG where, as I recall, @adrianhopebailie, myself, and a few others were on the losing side of the argument but we foresaw saw this and other extensibility concerns (as mentioned briefly by @adrianhopebailie above).
> 
> I don't know what this means moving forward for telling browsers what to do here at this point, but I would very much like to see this information left as opaque as possible with respect to the browser implementation. Whatever we do, the key point is this: we do not want to create an environment where browsers must be upgraded in order to support new payment methods.
> 
If anyone wants to discuss this seriously, then please start presenting actual code examples. I'm not interested in having email style discussions without actual code and tests. 

The reason we are talking past each other is because we are not using code. So stop - look at my code examples above, and prove whatever you want to prove (or disprove) with them.

adrianhopebailie:
> The data type for URL identifiers is object. For standardized short strings it is a WebIDL dictionary.

Wrong. That is what you think it is because you haven't been part of any of the conversations on the WG calls where there have been numerous assurances made that, even for a new short-string based payment method, the only implementation burden on browsers is adding that short string to an enum of known identifiers.

>  The reason we are talking past each other is because we are not using code

Actually I think it's because you joined this conversation about a year late and haven't attended a single WG call (where this issue has been discussed at length). I appreciate that the timezone is a challenge but perhaps it would be prudent to approach the online discussion a little more patiently.

To illustrate (and prove that I'm serious, because apparently that's also important), here is a code example:
  1. It has to be a short-string because there is no authority to host a URL for Bitcoin. 
  2. BIP 70 (the mechanism for doing Bitcoin payments) is designed to be extensible so its unlikely that each new version is going to be supported by a new revision of a payment method specification published by the WG to update the WebIDL.

```javascript
const methodData = [{
  supportedMethods: ["bitcoin"],
  data: { 
    payment_details_version: 1, 
    pki_type: "none", 
    pki_data: "",
    serialized_payment_details: "F1AABE5B69...",
    signature: "1FE56C09..."
  }
};
```

The WG has been tracking Bitcoin's BIP 70 as a potential payment method for some time, we mapped out the flow here: https://github.com/w3c/webpayments-flows/tree/gh-pages/PaymentFlows/Bitcoin
For the full explanation: https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki

While I appreciate that a web, centralized around a small number of browsers that can quickly confer on new developments, is a comfortable position from which to design this system, that's not the world many of us are coming from in payments. 

A truly decentralized payment method has no authority so it can't use a URL as a payment method identifier. The only mechanism therefor is the short-string identifier but that can't only be available if that implies browsers must implement a full request/response data validation and keep it up to date.

Hope that makes things a bit clearer.

marcoscaceres:
The bitcoin use case is easily addressed by simply moving the the version number to the PMI. 

Thus: 

```JS
const v1 = [{
  supportedMethods: ["bitcoin-v1"],
  data: { 
    pki_type: "none", 
    pki_data: "",
    serialized_payment_details: "F1AABE5B69...",
    signature: "1FE56C09..."
  }
};

// V2 added new things, changes the semantics/types of v1 fields
const v2 = [{
  supportedMethods: ["bitcoin-v2"],
  data: { 
    this_is_needed_in_v2: "thing", // Add new field
    pki_type: ["this", "is", "now", "an", "array"], 
    pki_data: "",
    serialized_payment_details: "F1AABE5B69...",
    signature: "1FE56C09..."
  }
};

// Bitcoin v3 recognized the error of their ways, so it's backwards compatible with v1, but adds further information 
const v3 = [{
  supportedMethods: ["bitcoin-v1", "bitcoin-v3"],
  data: { 
    // In v3, we add this field the is needed for a better user experience, 
    // but a v1 processor can safely ignore it: 
    new_field_ignored_in_v1: ["thing"],
    pki_type: "none", 
    pki_data: "",
    serialized_payment_details: "F1AABE5B69...",
    signature: "1FE56C09..."
  }
};
```

And so, this works fine:

```JS
const paymentRequest = new PaymentRequest([v3, v2, v1], details, options)
```

What am I missing? 

If nothing, then option 2 remains.

marcoscaceres:
(fixed a couple of typos in the examples above)

marcoscaceres:
Simplified the example above. 

Note that one then has the flexibility to mix-in and compose PMI versions or treat them independently. 

```JS
// Mixin by bitcoin by version
const v1 = {
  supportedMethods: ["bitcoin-v1"],
  data: {
    /*v1 member*/
  },
};
const v2 = {
  supportedMethods: ["bitcoin-v2"],
  data: Object.assing(
    {},
    v1.data,
    {
      /*v2 stuff, trashes v1 stuff where it needs to - because mistakes were made*/
    }
  ),
};
const v3 = {
  supportedMethods: ["bitcoin-v3"].concat(v1.methods),
  data: Object.assing(
    {},
    v1.data,
    {
      /*v3 stuff, which is backward compatible with v1*/
    }
  ),
};
// Because v3 and v1 are compatible, we only list it once. 
// For legacy systems/reasons, we include v2, but we can 
// always fall back to v1.
const paymentRequest = new PaymentRequest([v3, v2], details, options);

// Or, we can support stricter fallback (although order doesn't matter!):
v3.methods.pop(); // Remove v1. 
const paymentRequest = new PaymentRequest([v3, v2, v1], details, options);
```

Anyway, unless I'm missing something, the using the version in the PMI seems optimal to me: 

 *  it gives complete data-type assurances with graceful degradation,
 * provides extensibility at every level (even in versions, with the addition of optional types) by allowing reuse of IDL dictionaries. You can import v1 things into v2 and vN.
 * and reduces overall redundancy.

Enough said... 
 
obama-mic-drop

marcoscaceres:
@domenic it's basically impossible to reliably do https://github.com/w3c/browser-payment-api/pull/541 without the assurances given here. 

Can we please merge this so we can move forward?

domenic:
I'm not willing to merge a normative change of this magnitude without confirmation from the WG, sorry.

ianbjacobs:
Let's put it on the 8 June agenda [1]
Ian

[1] https://github.com/w3c/webpayments/wiki/Agenda-20170608

marcoscaceres:
Zach understands the problem, so I'm happy for him to represent my position on the call. Otherwise please move the call to a reasonable hour so I can participate.

marcoscaceres:
> I'm not willing to merge a normative change of this magnitude without confirmation from the WG, sorry.

I'm completely supportive of this being a working group decision - but it's absolutely unfair for you to  assume that members of the WG understand the implications or even what WebIDL is (just scroll up to see what I mean!). This is not a political or consensus thing - if the data is not valid, then users could get charged the wrong amount or accidentally misled (see #541 - where if the `.data` is screwed up, then certain display items won't show up or the total won't match!). 

As this directly affects browser engines, it's on browser engine people to explain to the WG who don't know WebIDL what the implications are. Otherwise, it's impossible for the WG to make an informed decision on this - and at least, from a Firefox perspective, we will willfully violate the spec where it could negatively impact users (like in this case).

dlongley:
@marcoscaceres you seem to be saying (paraphrasing):

"You can just mint a new payment method identifier and put out a spec with new Web IDL for every new bitcoin version."

Which seems to ignore @adrianhopebailie's second point, which was:

> BIP 70 (the mechanism for doing Bitcoin payments) is designed to be extensible so **its unlikely that each new version is going to be supported by a new revision of a payment method specification published by the WG to update the WebIDL.**

(emphasis added)

I think the response to that point should be this:

If bitcoin wants to do things that way then they should publish one spec and the Web IDL for their payment method "data" should just be a "plain JS object". We could even say that if a spec doesn't say anything about their "data" that it will fall back to the default Web IDL for a plain JS object (it is understood that it has to move between JS and C++ somehow).

The only question that remains: Will browsers reject simple string payment method identifiers if they have not implemented the Web IDL for those methods?

Will the browser do this:

```js
if(isPaymentMethodIdentifierAString()) {
  if(isPaymentMethodIdentifierKnown()) {
    // we know this, use appropriate Web IDL
  } else {
    // throw error
  }
} else {
  // use plain JS object Web IDL to coerce "data"
}
```

Or will it do this:

```js
if(isPaymentMethodIdentifierAString()) {
  if(isPaymentMethodIdentifierKnown()) {
    // we know this, use appropriate Web IDL
  } else {
    // use plain JS object Web IDL to coerce "data"
  }
} else {
  // use plain JS object Web IDL to coerce "data"
}
```

The problem with the former is that it means that people now can't use bitcoin (without a polyfill) until every browser implements it. They could use some URI (if URIs were permitted not just URLs) like "urn:payment-method:bitcoin" but that would be unfortunate for a variety of reasons.

With the second approach, it's true that you won't get certain types of checks for certain types of payment methods on certain browsers -- but those checks can be progressively added over time. What you will avoid is blocking people from using certain payment methods.

domenic:
> I'm completely supportive of this being a working group decision - but it's absolutely unfair for you to assume that members of the WG understand the implications or even what WebIDL is (just scroll up to see what I mean!).

I agree there is a lot of confusion here that is not really related to this issue. What I want is a simple working group decision on whether we're willing to change the semantics of my example from https://github.com/w3c/browser-payment-api/pull/536#issuecomment-303996517 from working to not working. More accurately, please see my "I'd like the WG to discuss and pick one of the following:"

I don't think this requires the working group to figure out how bitcoin works with payment request, or gain a better understanding of Web IDL. This is a fairly narrowly-scoped question, setting aside the the attempts to several people in this thread to blow it up into a larger discussion.

marcoscaceres:
> I don't think this requires the working group to figure out how bitcoin works with payment request, or gain a better understanding of Web IDL. This is a fairly narrowly-scoped question, setting aside the the attempts to several people in this thread to blow it up into a larger discussion.

Agree (even if I'm guilty of the blow up). @domenic, please also make a choice.  

I'll delete any comments that are not choice related.

dlongley:
In the future, let's not delete comments (you are welcome to delete your own, of course), but instead create a new issue and redirect over that way.

marcoscaceres:
Sure, will do.

rsolomakhin:
Hey there! Sorry, I'm late to the game. Our current implementation does not throw if you use non-`BasicCardRequest` for `"basic-card"` method name. We've had a lengthy discussion about this when we implemented that behavior. The resolution was documented in https://github.com/w3c/webpayments-methods-card/issues/20#issuecomment-268885348.

marcoscaceres:
Thanks @rsolomakhin

I also said there it should throw 😭 And Microsoft didn't respond. In light of new information, we should really get the Edge folks to let us know what they are doing here too. 

@patrickkettner, I'm really struggling to get an Edge version to run Web Platform tests on (it doesn't want to install with virtual box, and I'm scared of bricking a work machine with a insider build). Could you please please please help us? We have access to Suace Labs - so maybe we can run against Edge 15 somehow?

rsolomakhin:
I believe @adrianba gave a :+1: on my "may throw" proposal, which is hard to notice.

marcoscaceres:
(Sorry for all the typos, typing on my phone)

rsolomakhin:
@madmath has a Windows machine that runs Edge with PaymentRequest implementation. Can he help you in any way?

marcoscaceres:
Just want to know what they do with "basic-card" and if the perform the IDL conversion on the BasicCardRequest dictionary (i.e., does it throw).

madmath:
Here's what happens on https://rsolomakhin.github.io/pr/hooba-dooba/index.html in Edge. 

pr-hooba-dooba

marcoscaceres:
@madmath, so close - but I need the to know what happens if, in pr.js, you change `supportedInstruments` to the following:

```js
  var supportedInstruments = [{
    supportedMethods: [
      'basic-card',
    ],
    data: {
      supportedNetworks: ['hooba-dooba'],
      supportedTypes: ["this should throw", "credit"],
    }
  }];
```

And/or...

```js
  var supportedInstruments = [{
    supportedMethods: [
      'basic-card',
    ],
    data: {
      supportedNetworks: {"this-should-throw": ['hooba-dooba']},
      supportedTypes: ["credit"],
    }
  }];
```

And 

```JS
  var supportedInstruments = [{
    supportedMethods: [
      'basic-card',
    ],
    data: {
      supportedNetworks: "this will throw",
      supportedTypes: "so should this",
    },
  }];
```

rsolomakhin:
I've added these test cases to https://rsolomakhin.github.io/pr/hooba-dooba/. Please re-test, @madmath.

madmath:
All of the new examples open the PR dialog and give no indication of error, even in the console.

patrickkettner:
For all versions, I am getting

image
image

patrickkettner:
to be clearer, it throws after you close the wallet window, before that the window just stays there empty

marcoscaceres:
@patrickkettner, thank you so so so so much! The `unknown enum values in supported networks` is *exactly* what I was looking for. 

@domenic, seems the IDL conversion of `. data`  is implemented in Edge. I will independently verify this (I just got an ok from Mozilla to just go buy a Windows laptop to test on). 

@patrickkettner, about the timing of the error - that's some funky race condition which, in theory, should not happen. Expect more of those tricky race conditions with incoming tests. The sheet showing up blank is a bit of a worry, as that should never happen 😇

marcoscaceres:
@domenic Please see https://www.w3.org/2017/06/08-wpwg-minutes#item02 ... my reading is that WG is ok to merge this. 

WP Tests forthcoming.

ianbjacobs:
Please wait until the results of the 15 June teleconf before merging. @adrianhopebailie was not on the call and @mountainhippo and I would like to chat with him before finalizing the decision.

adrianhopebailie:
I have reviewed the minutes of the 8 June call and don't wish to block what appears to be consensus from the editors on a path forward.

I am disappointed that the editors appear to have effectively ignored the input that @dlongley, myself and others have provided for over 2 years on the bigger issue in favor of getting a decision on what to them seems likely a very minor issue.

We now have a design that makes it very difficult for decentralized payments systems to define a payment method that can be used by the Payment Request API but very easy for the incumbent monopoly (EMVCo) to maintain the status quo.

I honestly think that the browser experts pushing this work forward don't appreciate or understand the impact or magnitude of what they are doing to an industry they know very little about, but when my comments are deleted to avoid further discussion then I think I have exhausted my options in trying to change that.

The Web platform is littered with design decisions that are looked back on with great regret, usually because they were made without due consideration of the side-effects, I hope this is not one of them.

index.html
```diff
@@ -693,9 +693,22 @@
...
-              by <var>data</var>, the payment handler matches.
+              </li>
+              <li>For each <var>handler</var> of the resulting payment
+              handlers:
+                <ol>
+                  <li>If required by the specification that defines the
+                  <var>handler</var>, <a data-cite=
+                  "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+                  <var>data</var> to an IDL value. If it results in a error,
+                  reject <var>acceptPromise</var> with that error an terminate
```
domenic: s/an/and

domenic (approved):
Thanks so much for taking the time to work through this in the larger group. I hope I wasn't too obstructionist about this :).

I only have a minor editorial concern, which is that maybe we should make the hooks more explicit, and in the process allow validation to go beyond just Web IDL validation (as the group implied in the minutes). With that in mind, my phrasing for this PR would be:

- For each _handler_ of the resulting payment handlers:
  - If required by the specification that defines _handler_, **validate** the string _data_. If it results in an error, ...
  - If payment method specific capabilities supplied by the payment handler **match** the string _data_, the payment handler matches.

Here the **bold** terms are definitions which we define in this document. Then specs like basic-card can define validate and match algorithms.

WDYT?

Happy to land this as-is first and do that as a follow-up, as well.

index.html
```diff
@@ -693,9 +693,22 @@
...
-              For each resulting payment handler, if payment method specific
-              capabilities supplied by the payment handler match those provided
-              by <var>data</var>, the payment handler matches.
+              </li>
+              <li>For each <var>handler</var> of the resulting payment
+              handlers:
+                <ol>
+                  <li>If required by the specification that defines the
+                  <var>handler</var>, <a data-cite=
+                  "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
```
domenic: I guess if we go with this architecture you need to JSON-parse before you convert, otherwise you're just converting it to an IDL DOMString, which is no good.

index.html
ianbjacobs: It feels to me like there is conflation of "Handler" and "Payment method".It seems to me the text should say "If required by the specification that defines the payment method, ..."

Ian

index.html
```diff
@@ -693,9 +693,22 @@
...
+                  <var>data</var> to an IDL value. If it results in a error,
```
bentian: s/a/an

adrianhopebailie:
@marcoscaceres I may be misunderstanding what you have done here but I think you have replaced an algorithm for _determining which handlers are suitable to process the payment request_ with an algorithm that _validates the payment method specific data_.

We've all agreed that the latter is required but I think it should be in addition to the former not replace it.

i.e. For all accepted methods in the request, validate the `data`, then for all known handlers establish if they can handle one of the accepted methods. The outcome is a list of potential handlers that determine what the browser can present to the user as options to process the payment request.

This does beg the question as to whether methods with invalid data should just be discarded with a warning?

marcoscaceres:
> This does beg the question as to whether methods with invalid data should just be discarded with a warning?

They could be. That's an implementation detail tho. Just interested in getting the `.data` checked at this point so I can write more tests.

adrianhopebailie:
> Just interested in getting the .data checked at this point so I can write more tests.

Sure but this PR does discard some aspects of the original algorithm that are still required so I don't think it can be merged as is.

index.html
```diff
@@ -688,14 +688,31 @@
...
+              "!ECMA-262-2015#sec-json.parse">JSON-parsing</a> the second
+              element in <var>paymentMethod</var> tuple.
               </li>
-              <li>Determine which <a>payment handlers</a> support any of the
-              <a>payment method identifiers</a> given <var>identifiers</var>.
-              For each resulting payment handler, if payment method specific
-              capabilities supplied by the payment handler match those provided
-              by <var>data</var>, the payment handler matches.
+              <li>Let <var>methods</var> be a list of <a>payment methods</a>
+              that support <var>identifiers</var>.
```
ianbjacobs: I don't think "payment methods support identifiers". The identifiers identify the payment methods. The payment handlers support the payment methods. 

I prefer the earlier text.

Ian

marcoscaceres: Will switch it back. 

@adrianhopebailie, please provide code counter examples. If you can't show actual code examples then we can't have a discussion.

adrianhopebailie:
@marcoscaceres are you saying that if you delete something then in order to earn the privilege of engaging in a discussion with you on why, I must first re-write what you removed?

marcoscaceres:
Blocked on https://github.com/w3c/browser-payment-api/pull/551

ianbjacobs:
Just a note that we discussed on 15 June at the WPWG call:
 https://www.w3.org/2017/06/15-wpwg-minutes#item01

@zkoch is going to help coordinate with @marcoscaceres a fix that addresses both topics this text intends to cover: looping through payment methods and validation.

zkoch (approved):
LGTM. I think we've now addressed both use cases: validation of data AND finding matching payment handlers based on data and identifier

marcoscaceres:
@zkoch and I reworked this text in light of #551. It's ready to merge once #551 goes in.

ianbjacobs:
Thanks for adding back payment handler matching!

index.html
```diff
@@ -685,17 +685,29 @@
...
+              element in <var>paymentMethod</var> tuple.
+              </li>
+              <li>If required by the specification that defines the
+              <var>identifer</var>, then <a data-cite=
+              "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+              <var>data</var> to an IDL value. Otherwise, <a data-cite=
+              "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a> to
+              <a data-cite="!WEBIDL#idl-object">object</a>.
+              </li>
+              <li>If conversion results in an error, reject
```
adrianhopebailie, line 702: Isn't the intent here to discard this tuple and continue (possibly logging an error in the process)?

i.e. Rejecting the promise and terminating the algorithm results in no matched payment handlers, even if there are multiple `(identifer, data)` tuples and only one has invalid data.

Is it possible to throw and not reject the promise so that the browser can log the error and continue the next iteration of the loop?

If so, I propose the following, or something similar:

```
<li>If conversion results in an error, discard
 the <var>paymentMethod</var> tuple, throw the error and process 
the next tuple.</li>
```

domenic: What does "discard" mean? That's not a well-defined operation, on tuples or anything else...

ianbjacobs: I understand Adrian to be distinguishing two possible behaviors in the face of invalid data:

- Abandon that particular payment method identifier/data pair or
- Abandon the entire payment request 

I favor the former behavior (and think AdrianHB does).

Ian

domenic: OK. In that case I don't understand the suggestion. Discarding the tuple is a meaningless operation, and throwing the error is in direct contradiction to processing the next tuple (throwing an error terminates the algorithm, like it does in all programming languages).

adrianhopebailie: @domenic The algorithm is a loop through a list of tuples. I am proposing that instead of terminating the algorithm when bad data is found in one tuple, we just discard the offending tuple that contains the bad data and continue with the next.

domenic: Again, I don't understand what "discard" means. Would your sentence mean the same thing if you said "we just continue with the next"? And what happened to "throw the error"?

adrianhopebailie: @domenic please, try and understand that the majority of this WG are not experts on browser internals. We're trying our best to contribute nonetheless. Have some patience and read the whole comment. 

I only made a code suggestion because the last time I tried to contribute anything @marcoscaceres refused to read it unless it had code in it.

If you re-read my original comment I asked the following question:

> Is it possible to throw and not reject the promise so that the browser can log the error and continue the next iteration of the loop?
> If so, I propose...

I.e. There is a parsing step happening inside a loop. Can the parsing step throw and can the logic inside the loop catch and log the error and then continue?

This is the experience I'd expect a developer to get:
 * They call the API with a list of 3 payment methods
 * They have 3 different payment handlers registered that can each handle one of these payment methods
 * One payment method has bad data
 * They are presented with only 2 payment handlers to choose to handle the payment
 * An error is logged to the browser console

I used the phrase "discard" because in the process of looping through this list of payment methods we are also establishing a list of payment handlers that match them. If a payment method has bad data then it is "discarded" from the set used for matching.

Sorry if that terminology is confusing, I hope you understand my intent now and can suggest a better text replacement?

adrianba (approved):

adrianba:
LGTM - we've been moving in the direction of increasingly helping developers find issues so rejecting seems correct.

marcoscaceres:
Will start on tests this week.

index.html
zkoch: To close the loop on this (apologies for this post merge):

We discussed this on yesterday's editor call. While there is certainly merit in ignoring the offending payment method and its related data blob, we've been moving in the direction of explicitly throwing errors for the purpose of developer clarity across this spec. It seems best to follow suit here.

We realize there are other options (e.g. console logging a warning), but as this is core to how PR works, an explicit throw was voted as the best option by the editors.

adrianhopebailie: Per @adrianba 's comment below, I understand the editors to have decided to take a fail fast approach to this API, therefor a payment request with bad data in one of the provided payment methods will fail entirely rather than simply logging an error.

For consistency, I recommend that the payment method identifier is also validated as part of this algorithm. It should either be a recognised short-string identifier or a valid URL form identifier (per the validity rules in the PMI spec).

Also, I wonder if this sentence in the PMI spec should be revised:
"User agents MUST NOT enforce validity or well-formedness of standardized payment method identifiers. "

PR #540, opened 2017-05-29
Cater for platform-specific limitations

marcoscaceres:
Through [newly added web platform tests](https://w3c-test.org/payment-request/), we discovered that it was possible to crash various browsers through unreasonably large inputs (e.g., unrealistically large currency values of sizes of  10<sup>102</sup>). Or [absurdly sized payment request `id`s values](https://bugs.chromium.org/p/chromium/issues/detail?id=725744#c16). 

This differs across platforms, because of various platform specific limits. E.g., string size limits are different on Android vs on desktop platforms etc.   

As such, naturally, user agents will need to guard against unreasonably-large inputs (hence this PR).  

Accompanying this change, we will change the web platform tests to limit inputs to a size of 1024 characters. That's large enough to accommodate must ridiculous values (again, e.g., a currency being 1024 characters long), but not risk hitting platform specific limits. 

What we might do, is include these stress tests nonetheless, but just have them in try/catch blocks, to ensure the browser doesn't crash - but make no assumption that browsers will be able to handle these large values. 

@foolip, @jgraham, @zcorpan, would appreciate your input. Does the change to the web platform tests sound reasonable? Is there precedence we should follow? 

N.B.: Firefox and Chrome have been patched or are being patched to fix the crashes. But more evil tests to come! 😈

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/limits.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/a3c35fb...140b18c.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2904,7 +2904,7 @@
         result that would be obtained by the specification's algorithms.
       </p>
       <p>
-        User agents impose implementation-specific limits on otherwise
+        User agents MAY impose implementation-specific limits on otherwise
         unconstrained inputs, e.g., to prevent denial of service attacks, to
         guard against running out of memory, or to work around
         platform-specific limitations. When an input exceeds
--- a/index.html
+++ b/index.html
@@ -2904,6 +2904,12 @@
         result that would be obtained by the specification's algorithms.
       </p>
       <p>
+        User agents MAY impose implementation-specific limits on otherwise
+        unconstrained inputs, e.g., to prevent denial of service attacks, to
+        guard against running out of memory, or to work around
+        platform-specific limitations.
+      </p>
+      <p>
         The working group will demonstrate implementation experience by
         creating a <a href=
         "https://github.com/w3c/web-platform-tests/tree/master/payment-request">
--- a/index.html
+++ b/index.html
@@ -2904,10 +2904,13 @@
         result that would be obtained by the specification's algorithms.
       </p>
       <p>
-        User agents MAY impose implementation-specific limits on otherwise
+        User agents impose implementation-specific limits on otherwise
         unconstrained inputs, e.g., to prevent denial of service attacks, to
         guard against running out of memory, or to work around
-        platform-specific limitations.
+        platform-specific limitations. When an input exceeds
+        implementation-specific limit, the user agent MUST throw a
+        <a>TypeError</a> optionally informing the developer of how a particular
+        input exceeded a implementation-specific limit.
       </p>
       <p>
         The working group will demonstrate implementation experience by
--- a/index.html
+++ b/index.html
@@ -2910,7 +2910,7 @@
         platform-specific limitations. When an input exceeds
         implementation-specific limit, the user agent MUST throw, or, in the
         context of a promise, reject with, a <a>TypeError</a> optionally
-        informing the developer of how a particular input exceeded a
+        informing the developer of how a particular input exceeded an
         implementation-specific limit.
       </p>
       <p>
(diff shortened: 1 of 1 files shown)


foolip:
I think that crash tests as part of web-platform-tests would make good sense, especially when multiple browsers have had the same bug. We already have a few tests with "crash" in the filename:
* html/editing/dnd/crashers/dialog-001.html
* webgl/conformance-1.0.3/conformance/...*crash*

Presumably, some crashes would not reproduce if testharness.js is included, so maybe a new test type will be needed eventually. However, maybe it's fine to write crash tests using testharness.js where it's possible. I'd suggest naming them *-crash.html for now, so that it's possible to find them if we later want to impose some kind of structure on this.

domenic:
Strongly agree we should keep the crashing tests. If we're not going to standardize the limits (which I think might be a good idea?), then I guess those tests would pass whether or not an exception is thrown/the promise is rejected. We should perhaps standardize the exception?

foolip:
Yeah, if an exception is thrown, then I'd expect a spec to say what kind.

marcoscaceres:
Looks like it's going to be a TypeError for all cases:

https://chromium-review.googlesource.com/c/516882/7/third_party/WebKit/Source/modules/payments/PaymentRequest.cpp#168

That's fine with me.

marcoscaceres:
@molly26dalton, could you kindly please check if this works with the Edge team?

index.html
```diff
@@ -2904,6 +2904,16 @@
         result that would be obtained by the specification's algorithms.
       </p>
       <p>
+        User agents impose implementation-specific limits on otherwise
```
zcorpan: Missing "may".

index.html
```diff
@@ -2904,6 +2904,16 @@
...
+        unconstrained inputs, e.g., to prevent denial of service attacks, to
+        guard against running out of memory, or to work around
+        platform-specific limitations. When an input exceeds
+        implementation-specific limit, the user agent MUST throw, or, in the
+        context of a promise, reject with, a <a>TypeError</a> optionally
+        informing the developer of how a particular input exceeded a
```
zcorpan: s/a/an/

index.html
marcoscaceres: I don't think we need MAY here. It's a statement of fact that implementations do this.

zcorpan: Well they are non-conforming unless you allow it with a "may". :-)

zcorpan (approved):

marcoscaceres:
I've created the first set of crash tests:
https://github.com/w3c/web-platform-tests/pull/6108

Still missing PaymentDetailsModifier tests. Probably made a couple of copy/pasta errors - but would appreciate some feedback. 

@rsolomakhin, I've reduced the length of inputs on the constructor tests to 1024 or less:
https://github.com/w3c/web-platform-tests/pull/6063

Can you pease take a look. Chrome seems happy tho 👍

rsolomakhin:
LGTM, thx @marcoscaceres

marcoscaceres:
I'm going to move forward on merging this, but would appreciate finding additional tests owners to help with web platform tests (again, huge thanks to @rsolomakhin for getting the testing started and the tests Google provided thus far - they have been extremely helpful!). 

On the Mozilla side, the more tests I add, the more issues we keep finding - so would really appreciate more help from other browser vendors and other interested parties. I'm not feeling comfortable anymore moving this to CR until we have a complete test suite. It's revealing contentious problems (e.g., #541 and #536).

PR #541, opened 2017-05-29
Use first matching PaymentDetailsModifier

### LINKED ISSUE #309: Behavior for multiple modifiers for the same payment method is not defined
Currently the specification defines a `PaymentDetailsModifier` sequence which can be used to apply modifications to the total and display items as appropriate for the currently selected payment method:

https://w3c.github.io/browser-payment-api/#dom-paymentdetailsmodifier

However, the specification does not define the user agent behavior in the case where this sequence contains multiple `PaymentDetailsModifier` objects for the same payment method. This needs to be defined, otherwise user agents risk implementing differing behaviors that could result in incorrect amounts being charged.

I would suggest since this is a sequence stating the first entry to match a payment method will be used as its modifier.

Alternatively, we could say supplying multiple modifiers for a single payment instrument is not allowed and throw a `TypeError`.

marcoscaceres:
closes #309

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/multiple_modifiers.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/a7e1bda...2972fbb.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1364,6 +1364,75 @@
         modify the <a>PaymentDetailsBase</a> based on a <a>payment method
         identifier</a>. It contains the following members:
       </p>
+      <p>
+        When presenting a payment request to an end-user, if there are multiple
+        <a>PaymentDetailsModifier</a> dictionaries in a sequence for the same
+        <a>payment method identifier</a>, then the user agent MUST use the
+        first entry to match a <a>payment method identifier</a> as its
+        modifier.
+      </p>
+      <div class="note">
+        <p>
+          In the following example, the developer has included multiple
+          <a>PaymentDetailsModifier</a> for the same payment method identifier
+          ("basic-card"). In such a case, the user agent chooses the first
+          matching modifier. However, the second modifier would be still be
+          used for by the payment handler for "https://example.com/bobpay".
+        </p>
+        <pre class="example js" title="Multiple modifiers">
+          const modifiers = [
+            // This one will be used a for "basic-card" payment, when paying
+            // with a "Example Card" credit card...
+            {
+              supportedMethods: ["basic-card"],
+              total: {
+                currency: "USD",
+                value: "80.00",
+              },
+              additionalDisplayItems: [
+                {
+                  label: "ExampleCard loyalty discount, 20% off!",
+                  amount: {
+                    currency: "USD",
+                    value: "-20.00",
+                  },
+                },
+              ],
+              data: {
+                supportedTypes: ["credit"],
+                supportedNetworks: ["example-card"],
+              },
+            },
+            // In the following, "basic-card" is ignored because only
+            // the first "basic-card" above will match. However,
+            // users of the "example.com/bobpay" payment handler
+            // wishing to pay with an "Example Card" credit card
+            // will receive the 30% discount.
+            {
+              supportedMethods: ["basic-card", "https://example.com/bobpay"],
+              total: {
+                currency: "USD",
+                value: "70.00",
+              },
+              additionalDisplayItems: [
+                {
+                  label: "ExampleCard loyalty discount, 30% off!",
+                  amount: {
+                    currency: "USD",
+                    value: "-30.00",
+                  },
+                },
+              ],
+              data: {
+                supportedTypes: ["credit"],
+                supportedNetworks: ["example-card"],
+              },
+            },
+          ];
+          const options = { modifiers };
+          const request = new PaymentRequest(methods, details, options);
+        </pre>
+      </div>
       <dl>
         <dt>
           <dfn>supportedMethods</dfn>


marcoscaceres:
Can someone sanity check what is going on in the spec with regards to modifiers? My reading is that that matching on `.data` is not specified. 

We need #536 and a matching algorithm somwhere. Or we need to define how matching is done in Basic Card... otherwise, we can't really express what the example shows (i.e., one can't say "if you pay with an "Amex" credit card, you need to pay 3% extra"). 

Right now, you can only say, "if you use basic-card, you have to pay extra" - which is unhelpful.

sideshowbarker:
> Can someone sanity check what is going on in the spec with regards to modifiers? My reading is that that matching on `.data` is not specified.

Agreed, it’s not specified (but needs to be)

marcoscaceres:
@nickjshearer, this attempts to address #309, which you filed. Can you take a look please, and let me know if it address the concerns you raised.

domenic:
So I think this addresses the problem, but not in the ideal way. It should be part of the show() algorithm, IMO.

Also, currently we only store the modifiers as serialized strings in the [[serializedModifierData]] slot. It's a bug in the current spec that nobody every references that slot. We should probably reference it! But now I'm confused as to why it needs to be JSON-serialized at all. Will it be sent across processes? Is each payment app going to process it differently? If so we should make that explicit, like we kind of do for [[serializedModifierData]].

marcoscaceres:
> So I think this addresses the problem, but not in the ideal way. It should be part of the show() algorithm, IMO.

Agree - this is incomplete. I thought here was good as a starting point, as `updateWith()` also relies on this - not just `.show()`. 

What I'm thinking is that this section would define the "use the first matching one!" rule because: 
 
 1. This data needs to go to another UI process.
 2. The end-user may consult different payment methods.
 3. User could asynchronously add a payment handler, which may then be activated in the UI (e.g., this is the first time they use Web Payments - so they get onboarded).

For point 2 above, the use case/user story is:

"I'm going to try to pay with my Amex, because double points, oh yeah! 
- user picks basic card in UI.

"Oh, crap! If I do that, I get charged 5% more. So, screw that! I'm going to switch to 'bobpay.com', as I don't have to pay any fee (and, wow! I get a free mug!)." 
- User switches to "bobpay.com" as payment handler.

"Oh, wait, that's only if I use my bobpay debit card. M'eh. that's fine."
- user selects "debit card", clicks "Pay". 

> Also, currently we only store the modifiers as serialized strings in the [[serializedModifierData]] slot. It's a bug in the current spec that nobody every references that slot. We should probably reference it! 

Yes, probably here. 

> But now I'm confused as to why it needs to be JSON-serialized at all. Will it be sent across processes? 

Yes. Consider what I described above, where I might "shop around" for the best deal based on payment method:

screenshot 2017-06-05 19 05 44

I might jump back in forth between "basic-card" and "PayPal" above to see who gives me the better rates - this could include currency rates, free stuff, discounts, etc... all encoded in [[serializedModifierData]]. 

> Is each payment app going to process it differently? If so we should make that explicit, like we kind of do for [[serializedModifierData]].
 
Yeah. Hence why I think #536 is so important. Also, we can't filter things out of [[serializedModifierData]] because of point 3 above (payment method becomes available to the user _after_ payment request).

marcoscaceres:
Sorry, I mean serialized PaymentDetailsModifier above. Anyway, hopefully you get the idea.

domenic:
OK, that all makes sense. In that case I'd like us to explicitly reference [[serializedModifierData]] and talk about the process of sending each payment app (?) the serialized data, to tie things together.

I'm OK not doing it in show()/updateWith() for now, and keeping it here, as long as we have an issue marker and get it fixed as part of the CR transition.

ianbjacobs:
Hi @marcoscaceres, 

Is any additional WG discussion useful for this issue?

Ian

marcoscaceres:
@ianbjacobs, don't think so - but will ping you if need be.

marcoscaceres:
Hmm... maybe we might want the last matching one here.... because I think there can be partial/incremental matches (e.g., "modify all credit cards", "modify all visa credit cards", and it can get a little insane).

zkoch:
@marcoscaceres not sure what's happening with this. are we blocked on review?

marcoscaceres:
No, I'm was playing around with our implementation to seeing if "first" made more sense than "last" (https://github.com/w3c/payment-request/pull/541#issuecomment-317625011) ... same issue I raised in the basic card spec about this kind of behavior needing to be defined. 

Will go with "first" as we discussed, as it avoids some of craziness.

marcoscaceres:
Changed my mind around this. We should drop "seen" PMIs during construction - that way we only end up with the first unique ones. That solves for both .show() and .canMakePayment().

marcoscaceres:
`.updateWith()` will also need this code path.

PR #545, opened 2017-06-08
Remove dispatch flag check on .updateWith()

### LINKED ISSUE #544: updateWith() "InvalidStateError" DOMException
@domenic, in `updateWith()` step 2-4, are: 

> 2. Let request be the value of event's target attribute.
> 3. If request is not a PaymentRequest object, then throw a TypeError.
> 4. If the dispatch flag is unset, then throw an "InvalidStateError" DOMException.

It seems impossible to ever throw in 4 because an event's `.target` cannot be set without dispatching the event through another object (thus the dispatch flag is set). 

If we want to catch 4, we would need to swap it with 3. So:

```JS
const ev = new PaymentRequestUpateEvent("test");
ev.updateWith(Promise.resolve({})); // would throw InvalidStateError, not TypeError.
```

However, that seems slightly pointless - unless there is precedence in the platform to do that. 

WDTY? Swap 3-4, or just delete 4?

marcoscaceres:
Closes #544.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/dispatch_flag_check.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/23cd48a...f085bd3.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2090,9 +2090,6 @@
             <li>If <var>request</var> is not a <a>PaymentRequest</a> object,
             then <a>throw</a> a <a>TypeError</a>.
             </li>
-            <li>If the <a>dispatch flag</a> is unset, then <a>throw</a> an
-            "<a>InvalidStateError</a>" <a>DOMException</a>.
-            </li>
             <li>If <var>event</var>.<a>[[\waitForUpdate]]</a> is true, then <a>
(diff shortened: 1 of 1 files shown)


domenic (approved):
LGTM. I assume you probably discovered this in the course of writing tests so there will be good tests for it.

PR #547, opened 2017-06-09
throw if updateWith() is not trusted

### LINKED ISSUE #546: Unstrusted events can update PaymentRequest
Unless I've missed something, it seems one can co-opt and untrusted event to do the work of a trusted event*:

```JS
const request = new PaymentRequest(defaultMethods, defaultDetails);
const untrustedEvent = new PaymentRequestUpdateEvent("test"); 
request.dispatchEvent(ev); // set .target
request.show();
request.onshippingaddresschange = () => {
  const update = {
    total: {
      label: "Untrusted",
      amount: {
        currency: "USD",
        value: "1111.00",
      },
    },
  };
  untrustedEvent.updateWith(Promise.resolve(update));
};
```
We might just need to say in `.updateWith()` that if the event is untrusted, throw a `SecurityError` or `NotAllowedError`.

*Chrome ignores this.

marcoscaceres:
* closes #546

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/update_ev_is_trusted.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/3b7f783...a6f85f1.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2090,12 +2090,6 @@
             <li>Let <var>request</var> be the value of <var>event</var>'s
             <a data-cite="DOM#dom-event-target">target</a> attribute.
             </li>
-            <li>If <var>request</var> is not a <a>PaymentRequest</a> object,
-            then <a>throw</a> a <a>TypeError</a>.
-            </li>
-            <li>If the <a>dispatch flag</a> is unset, then <a>throw</a> an
-            "<a>InvalidStateError</a>" <a>DOMException</a>.
-            </li>
             <li>If <var>event</var>.<a>[[\waitForUpdate]]</a> is true, then <a>
               throw</a> an "<a>InvalidStateError</a>" <a>DOMException</a>.
             </li>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Given the`.isTrusted` assurance from the UA, it follows that: iff the user agent dispatches the event, then the dispatch flag is always set, and the .target is always a `PaymentRequest` on which event was dispatched. 

Therefore, the checks of 8771656 are now redundant.

marcoscaceres:
Oh no... I just read that mouse events can hack around this :) Removing 8771656. Let's see if I can come up with some evil tests using trusted events.

marcoscaceres:
Sorry for the noise... but actually https://github.com/w3c/browser-payment-api/pull/547#issuecomment-307305938 still applies. I got confused :(

marcoscaceres:
Associated tests: https://github.com/w3c/web-platform-tests/pull/6162

domenic (approved):
Hmm, do you really think NotAllowedError is the right choice? It implies the user denied permission, whereas I don't think permission was ever asked about whether you can update the payment details.

I might go with InvalidStateError.

But, LGTM, I am OK either way. It is borderline.

PR #551, opened 2017-06-18
BREAKING CHANGE: change `supportedMethods` to DOMString

### LINKED ISSUE #549: Drop array support for supported methods?
Given what's happening in https://github.com/w3c/browser-payment-api/pull/536, should we consider just dropping support for an array of supportedMethods? And just move to a simple string?

It simplifies the story quite a bit, and the original idea for making it an array was so we could do "supportedMethods: ['visa', 'mc']"?

@domenic was mentioning technically this might not be that bad given how the webidl works...

marcoscaceres:
Closes #549.

Changes `supportedMethods` from a sequence to single DOMString.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/supportedMethods.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/0eb169f...2a43dba.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1354,9 +1354,9 @@
           <dfn>supportedMethods</dfn>
         </dt>
         <dd>
-          The <a>supportedMethods</a> is a <a>payment method identifier</a>.
-          The members of the <a>PaymentDetailsModifier</a> only apply if the
-          user select this <a>payment method</a>.
+          The <a>supportedMethods</a> member is a <a>payment method
+          identifier</a>. The members of the <a>PaymentDetailsModifier</a> only
+          apply if the user select this <a>payment method</a>.
         </dd>
         <dt>
           <dfn>total</dfn>
--- a/index.html
+++ b/index.html
@@ -302,13 +302,13 @@
           </p>
           <pre class="example" title="The 'methodData' argument">
             const methodData = [{
-              supportedMethods: ["basic-card"],
+              supportedMethods: "basic-card",
               data: {
                 supportedNetworks: ['visa, 'mastercard'],
                 supportedTypes: ['debit']
               }
             }, {
-              supportedMethods: ["https://example.com/bobpay"],
+              supportedMethods: "https://example.com/bobpay",
               data: {
                 merchantIdentifier: "XXXX",
                 bobPaySpecificField: true
@@ -382,12 +382,6 @@
               </li>
               <li>For each <var>paymentMethod</var> of <var>methodData</var>:
                 <ol>
-                  <li>If the length of the
-                  <var>paymentMethod</var>.<a>supportedMethods</a> sequence is
-                  zero, then <a>throw</a> a <a>TypeError</a>, optionally
-                  informing the developer that each <a>payment method</a> needs
-                  to include at least one <a>payment method identifier</a>.
-                  </li>
                   <li>If the <a data-lt="PaymentMethodData.data">data</a>
                   member of <var>paymentMethod</var> is missing, let
                   <var>serializedData</var> be null. Otherwise, let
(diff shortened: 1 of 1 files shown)


zcorpan:
@marcoscaceres is there a web-platform-tests PR for this change?

marcoscaceres:
Not yet... if it gets accepted, we have to basically change all the tests (just quick find/replace). Waiting on Microsoft to give us the OK.

zcorpan:
I'd particularly like to see a test to assert that the old API is *not* supported (could be in historical.https.html).

marcoscaceres:
Yep, will definitely add those too.

marcoscaceres:
Unrelated to this, but @zcorpan, we should think about have some way getting github to block a pull request unless it has a corresponding test in WPT somehow (unless it's non-normative, of course). Would like to chat more about that at TPAC with you, @plehegar, @sideshowbarker, etc.

zcorpan:
Could maybe have a travis CI check that checks the commit messages (looking for a wpt PR URL) and the PR's labels (blocking for, e.g., "do not merge yet", "waiting for tests").

domenic (approved):
LGTM! Will let you decide when to merge given concerns about everyone being on-board, and having tests ready.

index.html
```diff
@@ -1346,26 +1338,25 @@
...
       <dl>
         <dt>
           <dfn>supportedMethods</dfn>
         </dt>
         <dd>
-          The <a>supportedMethods</a> member contains a sequence of <a>payment
-          method identifiers</a>. The remaining members in the
-          <a>PaymentDetailsModifier</a> apply only if the user selects a
-          <a>payment method</a> included in this sequence.
+          The <a>supportedMethods</a> is a <a>payment method identifier</a>.
```
zcorpan: Missing word "member"

index.html
```diff
@@ -1346,26 +1338,25 @@
...
       <p>
         The <a>PaymentDetailsModifier</a> dictionary provides details that
-        modify the <a>PaymentDetailsBase</a> based on <a>payment method
-        identifier</a>. It contains the following fields:
+        modify the <a>PaymentDetailsBase</a> based on a <a>payment method
+        identifier</a>. It contains the following members:
       </p>
       <dl>
         <dt>
           <dfn>supportedMethods</dfn>
```
zcorpan, line 1354: Should this be renamed to `supportedMethod`, now that it only takes a single payment method identifier?

marcoscaceres: I don't know how much breakage we can tolerate here. We are happy for it to break on the Mozilla side, but it's really up to Microsoft, Google, and others relying on Blink as they've shipped.

zcorpan: OK. So `["basic-card"]` would still work as it's casted to a DOMString "basic-card". An array with more than one value would concatenate with a comma, and likely (but not impossibly, as URLs can have commas) end up with an unsupported payment identifier. Is this intentional? Changing the name would be more of a clean slate, but I suppose risks ending up having to support both.

marcoscaceres: > Is this intentional? Changing the name would be more of a clean slate, but I suppose risks ending up having to support both.

Yeah, it's intentional. It's probably as backwards compatible as we are going to get.

zcorpan: You could use a union `(DOMString or sequence<DOMString>)` and throw if the sequence doesn't have exactly 1 item. Then an array with multiple items can't "accidentally" (maybe intentionally, if the payment provider is a bit evil) end up being accepted as a supported payment identifier.

zkoch (approved):
LGTM

mollyd26:
LGTM

mollyd16:
LGTM

marcoscaceres:
Will write some tests for this over next few days.

index.html
mnoorenberghe: > Should this be renamed to supportedMethod, now that it only takes a single payment method identifier?

This already threw me off when just looking at the spec. The plural name with a string type is confusing… I was wondering how we're serializing a list into a string (which would be unusual given we could use an array)… Was an implementer not willing to change it?

zcorpan:
@marcos-travel status on https://github.com/w3c/payment-request/pull/551#issuecomment-309408367 ? AFAICT that aspect is not covered by https://github.com/w3c/web-platform-tests/pull/6512

marcoscaceres:
Thanks for the reminder 😲 I forgot to add it.

PR #559, opened 2017-07-10
Editorial changes

ianbjacobs:
Hi Editors,

Here are some small editorial suggestions.

I also have some questions upon rereading the spec:

 * The phrase "settle" a promise first appears in 16.2.1 updateWith().
   It is not used elsewhere in the document. Should it be changed to
   be consistent with usage elsewhere in the document?

 * I would have expected information about testing and CR exit
   criteria in the status section, not the section on conformance.
  After CR it will no longer be useful in the section on conformance.
  Would you consider moving this information to the status section?
 
Thanks!

Ian

***
[Preview](https://s3.amazonaws.com/pr-preview/ianbjacobs/browser-payment-api/editorial-20170710.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/browser-payment-api/50df6d6...ianbjacobs:eb46443.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -239,13 +239,15 @@
         site, selecting a "Power Up" in an interactive game, or paying at a
         kiosk in a parking structure). The <a>PaymentRequest</a> allows
         developers to exchange information with the <a>user agent</a> while the
-        user is providing input before approving or denying a payment request.
+        user is providing input (up to the point of user approval or denial of
+        the payment request).
       </p>
       <p data-link-for="PaymentRequest">
-        The <a>user agent</a> as a whole has a single <dfn>payment request is
-        showing</dfn> boolean, initially false. This is used to prevent
-        multiple <a>PaymentRequest</a>s from being shown, via their
-        <a>show()</a> method, at the same time.
+        Because the simultaneous display of multiple <a>PaymentRequest</a> user
+        interfaces may confuse the user, this specification limits the <a>user
+        agent</a> to displaying one at a time via the <a>show()</a> method.
+        This is ensured by a <dfn>payment request is showing</dfn> boolean,
+        global to the browser and initially false.
       </p>
       <p data-link-for="PaymentRequest">
         The <a>shippingAddress</a>, <a>shippingOption</a>, and
@@ -619,10 +621,10 @@
           </p>
           <p>
             It is not possible to show multiple <a>PaymentRequest</a>s at the
-            same time within one <a>user agent</a>. Calling <a>show()</a> if
-            another <a>PaymentRequest</a> is already showing, even due to some
-            other site, will return a promise rejected with an
-            "<a>AbortError</a>" <a>DOMException</a>.
+            same time within one <a>user agent</a>. If a <a>PaymentRequest</a>
+            is already showing, calling <a>show()</a> —from any Web site— will
+            return a promise rejected with an "<a>AbortError</a>"
+            <a>DOMException</a>.
           </p>
         </div>
         <p>
@@ -775,6 +777,10 @@
             For example, if the <a>user agent</a> has delegated responsibility
             for the request to another app. In this situation, <a>abort()</a>
             will reject the returned <a>Promise</a>.
+          </p>
+          <p>
+            See also the algorithm when the <a>user aborts the payment
+            request</a>.
           </p>
         </div>
         <p>
@@ -1091,6 +1097,11 @@
           will be <a>JSON-serialized</a>.
         </dd>
       </dl>
+      <p class="note">
+        The value of <code>supportedMethods</code> was changed from array to
+        string, but the name was left as a plural to maintain compatibility
+        with deployed software.
+      </p>
     </section>
     <section data-dfn-for="PaymentCurrencyAmount" data-link-for=
     "PaymentCurrencyAmount">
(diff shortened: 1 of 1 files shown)


adrianba:
> * The phrase "settle" a promise first appears in 16.2.1 updateWith().
It is not used elsewhere in the document. Should it be changed to
be consistent with usage elsewhere in the document?

This is the only place where a developer provides a promise that needs to be settled.

ianbjacobs:
@adrianba wrote:

"This is the only place where a developer provides a promise that needs to be settled."

Ok, then it's a term of art that I didn't know. I see @domenic has written about it:
 https://github.com/domenic/promises-unwrapping/blob/master/docs/states-and-fates.md

Sorry for the noise,

Ian

index.html
```diff
@@ -2557,7 +2573,8 @@
         <ol class="algorithm">
           <li>If the <var>request</var>.<a>[[\updating]]</a> is true, then
           terminate this algorithm and take no further action. Only one update
-          may take place at a time. This should never occur.
+          may take place at a time. The <a>user agent</a> user interface should
```
marcoscaceres: Please revert this, as it introduces a MAY. Also, when implementing, the original is more clear - as it's converted to an `assert()` in each implementation.

index.html
```diff
@@ -2601,7 +2618,8 @@
           attribute of <var>request</var> is null or if the <a data-lt=
           "PaymentRequest.shippingOption">shippingOption</a> attribute of <var>
             request</var> is null, then terminate this algorithm and take no
-            further action. This should never occur.
+            further action. The <a>user agent</a> user interface should ensure
```
marcoscaceres: Please also revert this one - as above.

index.html
```diff
@@ -239,13 +239,15 @@
...
+        user is providing input (up to the point of user approval or denial of
+        the payment request).
       </p>
       <p data-link-for="PaymentRequest">
-        The <a>user agent</a> as a whole has a single <dfn>payment request is
-        showing</dfn> boolean, initially false. This is used to prevent
-        multiple <a>PaymentRequest</a>s from being shown, via their
-        <a>show()</a> method, at the same time.
+        Because the simultaneous display of multiple <a>PaymentRequest</a> user
+        interfaces may confuse the user, this specification limits the <a>user
```
marcoscaceres: s/may/might (we try not avoid putting conformance requirements on users in this spec)

index.html
```diff
@@ -239,13 +239,15 @@
...
+        agent</a> to displaying one at a time via the <a>show()</a> method.
+        This is ensured by a <dfn>payment request is showing</dfn> boolean,
+        global to the browser and initially false.
```
marcoscaceres: Please remove the ", global to the browser and initially false." Avoids duplication.

index.html
```diff
@@ -776,6 +778,10 @@
             for the request to another app. In this situation, <a>abort()</a>
             will reject the returned <a>Promise</a>.
           </p>
+          <p>
```
marcoscaceres: s/`<p>`/`<p class="note">`

index.html
```diff
@@ -1091,6 +1097,11 @@
           will be <a>JSON-serialized</a>.
         </dd>
       </dl>
+      <p class="note">
+        The value of <code>supportedMethods</code> was changed from array to
+        string, but the name was left as a plural to maintain compatibility
+        with deployed software.
```
marcoscaceres: s/deployed software/existing content on the Web.

index.html
```diff
@@ -1410,6 +1421,11 @@
           will be <a>JSON-serialized</a>.
         </dd>
       </dl>
+      <p class="note">
```
marcoscaceres: This should only appear once in the document. I suggest dropping this one.

marcoscaceres (changes requested):

index.html
ianbjacobs: Hi Marcos,

Just a note that my proposal does not introduce the word MAY. The proposed sentence is: "The <a>user agent</a> user interface should ensure that this never occurs." The replaced sentence was "This should never occur."

Here's why I propose the change:

* Cnsistency with 5 instances of that sentence elsewhere in the document. (There are 2
instances of the replaced sentence in the document, so I went with the majority formulation.)
* I think that formulating this as a user agent responsibility is more useful than stating existentially
that something should not occur.

Ian

P.S. I will update the pull request to take into account your other suggestions.

index.html
ianbjacobs: See above clarification (so not yet reverting).

index.html
ianbjacobs: I removed the latter of the 2 instances (leaving the first one earlier in the document).

index.html
marcoscaceres: I concur.

marcoscaceres (approved):

index.html
marcoscaceres: Actually, maybe just drop "user interface".

index.html
marcoscaceres: as above.

marcoscaceres (changes requested):

ianbjacobs:
Hi Marcos, I made the final requested change (deleted "user interface").

Ian

marcoscaceres (approved):

PR #564, opened 2017-07-24
chore: rename repo to match shortname

marcoscaceres:
Need to rename the repo, then merge this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ branches:
     - gh-pages
 env:
   global:
-    - URL="https://w3c.github.io/browser-payment-api/echidna-manifest.txt"
+    - URL="https://w3c.github.io/payment-request/echidna-manifest.txt"
     - DECISION="https://www.w3.org/2016/04/14-wpwg-minutes#item02"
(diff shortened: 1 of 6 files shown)


domenic:
Do we really want to break all those links?? :(

marcoscaceres:
I'd rather break this now before we go to CR and this gets really popular (as more browsers ship). 

We have /TR/ as the canonical for the spec regardless.

sideshowbarker (approved):

ianbjacobs:
@domenic,

> Do we really want to break all those links?? :(

I agree that's a bummer. I found this by you, however:

   Redirecting GitHub Pages after a repository move
   https://gist.github.com/domenic/1f286d415559b56d725bee51a62c24a7

Is that something we should do for the 4 or so repositories we have renamed?

Ian

domenic:
Yeah, that could work, if you are willing to edit the w3c.github.io page.

ianbjacobs:
@adrianhopebailie did the pull request for four specs, which I merged. Seems to work. Thanks @domenic!

Ian

PR #567, opened 2017-07-31
validate PaymentCurrencyAmount.currency (closes #490)

### LINKED ISSUE #490: Should user agent validate currency?
In the section "5. PaymentCurrencyAmount dictionary" the spec says:

>currency: A string containing a currency identifier. The value of currency can be any string that is valid within the currency system indicated by currencySystem.

The last sentence there makes sense, but it is not clear exactly how the user agent should behave here. Should the UA validate the currency if it knows about the currency system and then throw TypeError if the currency isn't valid in that system? And what should be the behavior of the UA if there is an unknown currency system?

I think it would be good for implementers if the spec was clearer or provided some guidelines here.

marcoscaceres:
* Hooks into ECMAScript's Intl's `IsWellFormedCurrencyCode`.
* Internally changes currencies into canonical form (upper case).
* Groups checking and normalization algorithms for "is valid amount" and "is valid total" - reducing a fair bit of spec text. 
* Relocates "valid decimal monetary value" to `PaymentCurrencyAmount`.
* Cites [[ECMASCRIPT]], instead of obsolete [[ECMA-262-2015]].
* Hooks into [[Infra]] ASCII uppercasing algo, for canonical form.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/payment-request/currency_codes.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/payment-request/8f30b32...a67b9d6.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1318,12 +1318,11 @@
           <li>Optionally, a single U+002D (-), to indicate that the amount is
           negative.
           </li>
-          <li>One or more <a>code points</a> in the range U+0030 DIGIT ZERO (0)
-          to U+0039 DIGIT NINE (9).
+          <li>One or more <a>code points</a> in the range U+0030 (0) to U+0039
+          (9).
           </li>
-          <li>Optionally, a single U+002E FULL STOP (.) followed by one or more
-          <a>code points</a> in the range U+0030 DIGIT ZERO (0) to U+0039 DIGIT
-          NINE (9).
+          <li>Optionally, a single U+002E (.) followed by one or more <a>code
+          points</a> in the range U+0030 (0) to U+0039 (9).
           </li>
         </ol>
         <div class="note">
@@ -1374,9 +1373,9 @@
             <a>Check and canonicalize <var>amount</var></a>. Rethrow any
             exceptions.
           </li>
-          <li>If the first <a>code point</a> of <var>value</var> is U+002D
-          HYPHEN-MINUS, then throw a <a>TypeError</a> optionally informing the
-          developer that a total can't be a negative number.
+          <li>If the first <a>code point</a> of <var>value</var> is U+002D (-),
+          then throw a <a>TypeError</a> optionally informing the developer that
+          a total can't be a negative number.
           </li>
         </ol>
       </section>
--- a/index.html
+++ b/index.html
@@ -1346,7 +1346,7 @@
           currency is invalid.
           </li>
           <li>Set <var>amount</var>.<a>currency</a> to the result of
-          <a data-cite="!infra#ascii-uppercase">ASCII uppercasing</a>
+          <a data-cite="!INFRA#ascii-uppercase">ASCII uppercasing</a>
           <var>amount</var>.<a>currency</a>.
           </li>
         </ol>
--- a/index.html
+++ b/index.html
@@ -1252,7 +1252,7 @@
           <dd>
             A URL that indicates the currency system that the <a>currency</a>
             identifier belongs to. By default, the value is
-            <code>urn:iso:std:iso:4217</code> indicating that <a>currency</a>
+            "<code>urn:iso:std:iso:4217</code>" indicating that <a>currency</a>
             is defined by [[!ISO4217]] (for example, <code>USD</code> for US
             Dollars).
           </dd>
--- a/index.html
+++ b/index.html
@@ -537,14 +537,11 @@
           </li>
           <li data-link-for="PaymentDetailsInit">Process the total:
             <ol>
-              <li>Let <var>total</var> be
-              <var>details</var>.<a>total</a>.<a data-lt=
-              "PaymentItem.amount">amount</a>.
-              </li>
               <li data-tests=
               "payment-request-ctor-currency-code-checks.https.html">
-                <a>Check and canonicalize <var>total</var></a>. Rethrow any
-                exceptions.
+                <a>Check and canonicalize total</a>
+                <var>details</var>.<a>total</a>.<a data-lt=
+                "PaymentItem.amount">amount</a>. Rethrow any exceptions.
               </li>
             </ol>
           </li>
@@ -552,13 +549,11 @@
           present, then for each <var>item</var> in
           <var>details</var>.<a>displayItems</a>:
             <ol>
-              <li>Let <var>amount</var> be <var>item</var>.<a data-lt=
-              "PaymentItem.amount">amount</a>.
-              </li>
               <li data-tests=
               "payment-request-ctor-currency-code-checks.https.html">
-                <a>Check and canonicalize <var>amount</var></a>. Rethrow any
-                exceptions.
+                <a>Check and canonicalize amount</a>
+                <var>item</var>.<a data-lt="PaymentItem.amount">amount</a>.
+                Rethrow any exceptions.
               </li>
             </ol>
           </li>
@@ -580,13 +575,12 @@
                   </li>
                   <li>For each <var>option</var> in <var>options</var>:
                     <ol>
-                      <li>Let <var>amount</var> be
-                        <var>item</var>.<a data-lt="PaymentItem.amount">amount</a>.
-                      </li>
                       <li data-tests=
                       "payment-request-ctor-currency-code-checks.https.html">
-                        <a>Check and canonicalize <var>amount</var></a>.
-                        Rethrow any exceptions.
+                        <a>Check and canonicalize amount</a>
+                        <var>item</var>.<a data-lt=
+                        "PaymentItem.amount">amount</a>. Rethrow any
+                        exceptions.
                       </li>
                       <li>If <var>seenIDs</var> contains
                       <var>option</var>.<a>id</a>, then set <var>options</var>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Tests incoming also

index.html
```diff
@@ -565,30 +537,26 @@
...
+              <li>Let <var>total</var> be
               <var>details</var>.<a>total</a>.<a data-lt=
-              "PaymentItem.amount">amount</a>.<a data-lt=
-              "PaymentCurrencyAmount.value">value</a> is U+002D HYPHEN-MINUS,
-              then <a>throw</a> a <a>TypeError</a>, optionally informing the
-              developer that the total can't be negative.
+              "PaymentItem.amount">amount</a>.
+              </li>
+              <li>
+                <a>Check and canonicalize <var>total</var></a>. Rethrow any
```
domenic: I'm a little weirded out by how the variable name is both a variable name and part of the algorithm name. IMO "[Check and canonicalize the total] _total_" would be better.

index.html
```diff
@@ -565,30 +537,26 @@
...
+                exceptions.
```
domenic: In theory rethrowing exceptions is not necessary (https://infra.spec.whatwg.org/#algorithm-control-flow), but in practice we should stay consistent, and I believe the spec mostly includes this verbiage right now. So, just noting FYI.

index.html
domenic: This would also allow you to avoid the intermediate variables that have been inserted in various places.

index.html
```diff
@@ -866,8 +827,8 @@
               <var>paymentMethod</var> tuple.
               </li>
               <li>Let <var>data</var> be the result of <a data-cite=
-              "!ECMA-262-2015#sec-json.parse">JSON-parsing</a> the second
-              element in <var>paymentMethod</var> tuple.
+              "!ECMASCRIPT#sec-json.parse">JSON-parsing</a> the second element
+              in <var>paymentMethod</var> tuple.
```
domenic: While here: "**item** in **the** paymentMethod tuple"

index.html
```diff
@@ -1307,9 +1268,21 @@
...
+          </p>
+          <p>
+            When using [[!ISO4217]], all <a data-cite=
+            "!ecma-402#sec-iswellformedcurrencycode">well-formed</a> 3-letter
+            currency codes are allowed. Their canonical form is upper case.
+            However, the set of combinations of currency code for which
+            localized currency symbols are available is implementation
+            dependent. Where a localized currency symbol is not available, a
+            user agent SHOULD us the [[!ISO4217]] currency code SHOULD be used
+            for formatting.
```
domenic: Last sentence is messed up; should be either "a user agent should" or "should be used", not both.

index.html
```diff
@@ -1321,12 +1294,85 @@
...
+        </h3>
+        <p>
+          A <dfn data-cite="INFRA#javascript-string">JavaScript string</dfn> is
+          a <dfn>valid decimal monetary value</dfn> if it consists of the
+          following <dfn data-lt="code point" data-cite="INFRA#code-point">code
+          points</dfn> in the given order: [[!INFRA]]
+        </p>
+        <ol>
+          <li>Optionally, a single U+002D HYPHEN-MINUS (-), to indicate that
+          the amount is negative.
```
domenic: In Infra we've moved toward omitting the spelled out name, so you'd just do `U+002D (-)` these days. See https://infra.spec.whatwg.org/#code-points

index.html
```diff
@@ -1321,12 +1294,85 @@
...
+        <p>
+          To <dfn data-lt="check and canonicalize amount">check and
+          canonicalize <var>amount</var> <a>PaymentCurrencyAmount</a></dfn>,
+          run the following steps:
+        </p>
+        <ol data-link-for="PaymentCurrencyAmount">
+          <li>If <a>currencySystem</a> is not
+          <code>urn:iso:std:iso:4217</code>, terminate this algorithm.
+          </li>
+          <li>let <var>isValidCurrency</var> be the result of calling
```
domenic: Uppercase "let"

index.html
```diff
@@ -1321,12 +1294,85 @@
...
+          <a data-cite=
+          "!ecma-402#sec-iswellformedcurrencycode">IsWellFormedCurrencyCode</a>
+          abstract operation with <a>currency</a> as the argument.
+          </li>
+          <li>If <var>isValidCurrency</var> returns false, then throw a
```
domenic: s/returns/is/

index.html
```diff
@@ -1321,12 +1294,85 @@
...
+          definition.
+          <pre class="hljs">^-?[0-9]+(\.[0-9]+)?$</pre>
+        </div>
+        <p>
+          To <dfn data-lt="check and canonicalize amount">check and
+          canonicalize <var>amount</var> <a>PaymentCurrencyAmount</a></dfn>,
+          run the following steps:
+        </p>
+        <ol data-link-for="PaymentCurrencyAmount">
+          <li>If <a>currencySystem</a> is not
```
domenic: It should be _amount_.currencySystem. Same for `currency` and "_amount_'s `value`" below.

index.html
```diff
@@ -1321,12 +1294,85 @@
...
+          <code>urn:iso:std:iso:4217</code>, terminate this algorithm.
```
domenic: This is a string, so needs quotes around it.

index.html
```diff
@@ -1321,12 +1294,85 @@
...
+          informing the developer that the currency is invalid.
+          </li>
+          <li>Let <var>value</var> be <var>amount</var>'s <a>value</a>.
+          </li>
+          <li>If <var>value</var> is not a <a>valid decimal monetary value</a>,
+          throw a <a>TypeError</a>, optionally informing the developer that the
+          currency is invalid.
+          </li>
+          <li>Canonicalize <a>currency</a> by converting it to upper case.
+          </li>
```
domenic: This is written a little ambiguously. I'd prefer "Set _amount_.currency to the result of [ASCII uppercasing](https://infra.spec.whatwg.org/#ascii-uppercase) _amount_.currency". (Or is ASCII uppercasing incorrect here? In which case we should pick a well-defined uppercasing algorithm.)

index.html
```diff
@@ -1321,12 +1294,85 @@
...
+        <ol data-link-for="PaymentCurrencyAmount">
+          <li>If <a>currencySystem</a> is not
+          <code>urn:iso:std:iso:4217</code>, terminate this algorithm.
+          </li>
+          <li>
+            <a>Check and canonicalize <var>amount</var></a>. Rethrow any
+            exceptions.
+          </li>
+          <li>If the first <a>code point</a> of <var>value</var> is U+002D
+          HYPHEN-MINUS, then throw a <a>TypeError</a> optionally informing the
```
domenic: RangeError seems better? But maybe not worth the churn at this stage.

domenic (changes requested):
Great simplification and consolidation. Some issues with the algorithm prose though.

marcoscaceres:
Tests: https://github.com/w3c/web-platform-tests/pull/6670

index.html
marcoscaceres: Yeah, let's leave this one... it be a pain to change it now.

index.html
marcoscaceres: Woops... this unhelpful anyway. I did some research, and what we actually want here is "SHOULD use U+00A4 (¤)" - which is the the generic currency sign: https://en.wikipedia.org/wiki/Currency_sign_(typography)

index.html
marcoscaceres: ASCII uppercasing is correct.

marcoscaceres:
Just noting the tests only cover constructor. Need to make a bunch of manual tests for `updateWith()`.

index.html
marcoscaceres: Fixing this now...

index.html
marcoscaceres: I fixed these globally also to match infra.

index.html
marcoscaceres: Yeah, the rethrow is a bit redundant - we need some kind of spec/platform convention whereby if algorithm name starts with the word "X", it potentially throw.

Maybe we already have that for these kinds of checker/validator algorithms?

index.html
```diff
@@ -1321,12 +1288,87 @@
...
+          currency is invalid.
+          </li>
+          <li>Set <var>amount</var>.<a>currency</a> to the result of
+          <a data-cite="!INFRA#ascii-uppercase">ASCII uppercasing</a>
+          <var>amount</var>.<a>currency</a>.
+          </li>
+        </ol>
+        <p>
+          To <dfn data-lt="check and canonicalize total">check and canonicalize
+          <var>total</var> <a>PaymentCurrencyAmount</a></dfn>, run the
```
domenic: This should be `<dfn>check and canonicalize total</dfn> given a <a>PaymentCurrencyAmount</a> <var>total</var>`, I think.

(Same for the "amount" version)

index.html
```diff
@@ -1321,12 +1288,87 @@
...
+          <a data-cite=
+          "!ecma-402#sec-iswellformedcurrencycode">IsWellFormedCurrencyCode</a>
+          abstract operation with <var>amount</var>.<a>currency</a> as the
+          argument.
+          </li>
+          <li>If <var>isValidCurrency</var> is false, then throw a
+          <a>RangeError</a> exception and terminate this algorithm, optionally
+          informing the developer that the currency is invalid.
+          </li>
+          <li>Let <var>value</var> be <var>amount</var>.<a>value</a>.
```
domenic: Unsure this intermediate `value` variable is helpful.

index.html
```diff
@@ -1321,12 +1288,87 @@
...
+          </li>
+        </ol>
+        <p>
+          To <dfn data-lt="check and canonicalize total">check and canonicalize
+          <var>total</var> <a>PaymentCurrencyAmount</a></dfn>, run the
+          following steps:
+        </p>
+        <ol data-link-for="PaymentCurrencyAmount">
+          <li>If <var>total</var>.<a>currencySystem</a> is not
+          <code>urn:iso:std:iso:4217</code>, terminate this algorithm.
```
domenic: Missing quotes here

index.html
```diff
@@ -3002,13 +3031,15 @@
         </dt>
```
domenic, line 3029: Above this line, it says "ECMA-262 6th Edition, The ECMAScript 2015 Language Specification", but probably it should say "ECMAScript" or "The ECMAScript Language Specification" or similar.

domenic (approved):
LGTM with nits

index.html
marcoscaceres: Went with ECMAScript.

marcoscaceres:
Huge thanks for the review, @domenic! Much improved.

PR #574, opened 2017-08-03
Editorial: change around payment handler language (closes #532)

### LINKED ISSUE #532: Matching algo doesn't handle absence of filters
The following text does not handle the case where the filters provided by `data` are empty and the capabilities supplied by the payment handler are not. This should result in a match (i.e. No handlers are filtered out)

The WG sought to alleviate confusion around this topic by using the term "capabilities" to indicate the properties of the payment handler considered in this algorithm and the term "filters" to indicate the parts of the payment method specific data considered in this algorithm.

This has been excluded from the current text:

```html
<li>Determine which <a>payment handlers</a> support any of the
  <a>payment method identifiers</a> given <var>identifiers</var>.
  For each resulting payment handler, if payment method specific
  capabilities supplied by the payment handler match those provided
  by <var>data</var>, the payment handler matches.
</li>
```

Proposed amendment:

```html
<li>Determine which <a>payment handlers</a> support any of the
  <a>payment method identifiers</a> given <var>identifiers</var>.
  For each resulting payment handler, if <var>data</var> contains
  filters that exclude the payment handler on the basis of the 
  capabilities it has supplied, then the payment handler
  does not match, otherwise the payment handler matches.
</li>
```

ianbjacobs:
Based on AHB's proposal:
 https://github.com/w3c/payment-request/issues/532#issuecomment-318745090

closes 532

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/payment-request/handler_tweak_ahb.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/payment-request/c2783a1...c12958a.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -841,9 +841,9 @@
               </li>
               <li>Determine which <a>payment handlers</a> support
               <var>identifier</var>. For each resulting payment handler, if
-              payment method specific capabilities supplied by the payment
-              handler match those provided by <var>data</var>, the payment
-              handler matches.
+              <var>data</var> contains filters that exclude the payment
+              handler, then the payment handler does not match, otherwise the
+              payment handler matches.
               </li>
             </ol>
           </li>


index.html
```diff
@@ -841,9 +841,9 @@
               </li>
               <li>Determine which <a>payment handlers</a> support
               <var>identifier</var>. For each resulting payment handler, if
-              payment method specific capabilities supplied by the payment
-              handler match those provided by <var>data</var>, the payment
-              handler matches.
+              <var>data</var> contains filters that exclude the payment
```
marcoscaceres, line 844: I still find the word "filter" extremely problematic. The `data` attribute contains only "members" - and those members are acted upon based on the specification that defines those member (or no action is taken, in the case the PMI is a URL-based PMI).

marcoscaceres (changes requested):
I'm inclined to reject this change as currently written, sorry.

index.html
ianbjacobs: I think Adrian's concern is the default behavior. Here's a version that endeavors
to capture (what I believe is) the intent, but sticks closer to the language currently
in the spec:

<li>Determine which <a>payment handlers</a> support <var>identifier</var>. 
For each resulting payment handler, if payment method specific capabilities 
supplied by the payment handler do not match those provided by 
<var>data</var>, the payment handler does not match, otherwise the 
payment handler matches.</li>

(*deleted mistaken text here*)

marcoscaceres: To be clear: I can't accept a PR with the word "filter".

marcoscaceres: Looking at this text (including the one in the spec), it actually doesn't make any sense at all... It's a case of "take these identifiers and their corresponding data, and let god sort it out." That's all we can really say here. Right now it's super hand-wavy on purpose.

ianbjacobs: @adrianhopebailie, when you have time, please weigh in.

adrianhopebailie: The current wording is not just "hand-wavy" it's incorrect. The intent when this was designed was that a handler matches by default UNLESS something provided in the `data` part of the request excludes it.

The current wording suggests that a payment handler ONLY matches if the appropriate `data` is provided.

Example: If the handler supports `basic-card` and supported networks of `visa` and `mastercard` the current text suggests that the payment request MUST list `visa` or `mastercard` under `supportedNetworks` in the `data`  of the request or it will not match.

The intent is that a payment request with a method of `basic-card` and an empty `data` should match. i.e Empty implies, "no filter applied". 

We've used the terminology "filters" for requests and "capabilities" for handlers for some time so not sure I understand this, but whatever...

> I can't accept a PR with the word "filter".

ianbjacobs: @adrianhopebailie, does this version work for you: "Determine which payment handlers support identifier. For each resulting payment handler, if payment method specific capabilities supplied by the payment handler do not match those provided by data, the payment handler does not match, otherwise the payment handler matches."

adrianhopebailie: @ianbjacobs - I'm still not clear what "capabilities provided by data" means. It doesn't make sense to me.

The spec should be talking about what is in the `data` as "filtering" the available payment handlers (even if not called "filters" explicitly). How about:

"Determine which payment handlers support identifier. Filter the set of matching payment handlers further by comparing the capabilities of each payment handler with the content of `data`. For each payment handler; if `data` implies that payment method specific capabilities are required and these are not provided, the payment handler does not match, otherwise the payment handler matches."

Bare in mind that we are exploring a `canMakePayment()` lambda on handlers so the assumption is that this would be called here for web-based handlers BUT we don't (can't) talk about that explicitly and create a dependency on Payment Handler so we need to have something that implies that (or similar behavior) but is generic enough to not introduce the dependency and also works for other types of handler like the built in handler for basic card and platform native handlers too.

dlongley: Would the word "restrictions" or "constraints" help resolve the issue with the word "filters"?

"`data` may constrain which payment handlers can be used by specifying required capabilities. Only those payment handlers that have indicated their support for the capabilities specified in `data` match. If `data` provides no such constraints then any payment handler that supports the payment method identifier matches."

marcoscaceres: What is being stated in the spec, and the new proposed text doesn't match implementations. If the merchant says:

> I support "basic-card", but only type "debit" of the "mastercard" network. 

consider: 

1. Firstly, `.canMakePayment()` returns "false" in Chrome and Edge. But it returns "true" in Firefox because, theoretically, the user "can make the payment" by later adding a Mastercard debit card (it's also privacy preserving in the Firefox case, because I might not want you to know if I have a card of this type). 
1. Now for `.show()`, the payment sheet will still be shown to the user because basic card still gives the user the possibility to add a card manually. So the whole thing above about matching here, sets up false expectations for developers, and the user agent just ignores it at this point: the payment sheet is still shown and `.data` *is not taken into consideration*.  

Please try it yourself in Chrome or Edge (and the behavior will be the same in Firefox irrespective of what the spec says - so let's make the spec match reality!):

```JS
const pr = new PaymentRequest(
  [
    {
      supportedMethods: ["basic-card"],
      data: {
        supportedTypes: ["debit"],
        supportedNetworks: ["mastercard"]
      }
    },
  ],
  {
    total: {
      label: "The cake is a lie",
      amount: { currency: "USD", value: "1.2" },
    },
  }
)
pr.canMakePayment().then(r => {
  console.log("can make payment?", r);
   return pr.show(); // ALL IMPLEMENTATIONS ALLOW THIS!
}).then(console.log("done", pr)).catch(err => console.log(err))
``` 

Thus, it's up to the "basic-card" or other PMI specs to define how things work (or not!). The `.data`'s `supportedNetwork` and `supportedTypes` are merely a hint (e.g., "visa" can be greyed out as a icon in the UI). In Edge, it doesn't seem to prevent the end-user from returning a Visa credit card number at all. 

In Chrome, the user is prevented from returning a card that is not "mastercard" and "debit". But this is a quality of implementation concern. Chrome *STILL SHOWS THE PAYMENT SHEET* (no filtering or matching happened, even tho `.canMakePayment()` returned false) ... and the merchant still needs to validate the resulting `details`.

My suggestion is we drop the text around `.data` entirely, and then define in each PMI spec how it's supposed to work (because, in the case of "basic-card" it's still waaaaAAAaaaay underspecified).

ianbjacobs: While it is important that the spec align with implementations, I think it's important to set high level expectations about .data filtering in PR API. Please do not remove high level messaging about capability matching from PR API.

marcoscaceres: All specs will fail this in the test suite then and this spec will never go to REC 🤷🏽‍♂️.

marcoscaceres: *err, all implementations I mean.

marcoscaceres: (and yes, you can count on me to add this test to force everyone to fail 😼)

marcoscaceres:
Superseded by https://github.com/w3c/payment-request/pull/578 to:

* match reality.  
* make this testable (without forcing everyone to fail).
* prevent confusing developers and implementers by including things in the spec that won't ever be implemented.

adrianhopebailie:
I like the direction @marcoscaceres counter-proposal is taking in #578 (although I don't think it is 100% ready yet) so closing this proposal.

(Thanks @ianbjacobs for posting on my behalf)

PR #578, opened 2017-08-08
remove matching .data text from .show()

### LINKED ISSUE #532: Matching algo doesn't handle absence of filters

marcoscaceres:
closes #532, #574

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/payment-request/redundant.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/payment-request/343fe47...95c1148.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -818,6 +818,8 @@
           <li>Return <var>acceptPromise</var> and perform the remaining steps
           <a>in parallel</a>.
           </li>
+          <li>Let <var>handlers</var> be an empty unordered set.
+          </li>
           <li>For each <var>paymentMethod</var> tuple in
           <var>request</var>.<a>[[\serializedMethodData]]</a>:
             <ol>
@@ -825,8 +827,8 @@
               <var>paymentMethod</var> tuple.
               </li>
               <li>Let <var>data</var> be the result of <a data-cite=
-              "!ECMASCRIPT#sec-json.parse">JSON-parsing</a> the second item in
-              the <var>paymentMethod</var> tuple.
+              "!ECMASCRIPT#sec-json.parse">JSON-parsing</a> the second element
+              in the <var>paymentMethod</var> tuple.
               </li>
               <li>If required by the specification that defines the
               <var>identifer</var>, then <a data-cite=
@@ -839,33 +841,36 @@
               <var>acceptPromise</var> with that error and terminate this
               algorithm.
               </li>
-              <li>Determine which <a>payment handlers</a> support
-              <var>identifier</var>. For each resulting payment handler, if
-              payment method specific capabilities supplied by the payment
-              handler match those provided by <var>data</var>, the payment
-              handler matches.
+              <li>If the user agent has a registered <a>payment handler</a>
+              that supports <var>identifier</var>, then check if the payment
+              handler is authorized and capable of handling the payment request
+              with <var>data</var>. If the check returns in the affirmative,
+              then add payment handler to <a>handlers</a>.
               </li>
             </ol>
           </li>
-          <li>If this consultation produced no supported method of paying, then
-          reject <var>acceptPromise</var> with a "<a>NotSupportedError</a>" <a>
-            DOMException</a>, and set the <a>user agent</a>'s <a>payment
-            request is showing</a> boolean to false.
+          <li>If <var>handlers</var> is empty, then reject
+          <var>acceptPromise</var> with a "<a>NotSupportedError</a>"
+          <a>DOMException</a>, and set the <a>user agent</a>'s <a>payment
+          request is showing</a> boolean to false.
           </li>
           <li>
             <p>
-              Otherwise, show a user interface to allow the user to interact
-              with the payment request process, using those <a>payment
-              handlers</a> and <a>payment methods</a> which the above step
-              identified as feasible. The user agent SHOULD prioritize the
-              preference of the user when presenting payment methods and
-              applications.
+              Otherwise, present a user interface to allow the user to interact
+              with the <var>handlers</var>. The user agent SHOULD prioritize
+              the preference of the user when presenting payment methods.
             </p>
             <p>
-              The <a>payment handler</a> should be sent the appropriate data
-              from <var>request</var> in order to guide the user through the
-              payment process. This includes the various attributes and
-              internal slots of <var>request</var>.
+              For the <a>payment handler</a> selected by the end-user, the user
+              agent MUST pass the <a data-cite=
+              "!WEBIDL#dfn-convert-ecmascript-to-idl-value">converted</a>
+              second element in the <var>paymentMethod</var> tuple. Optionally,
+              the user agent SHOULD send the appropriate data from
+              <var>request</var> to the user-selected <a>payment handler</a> in
+              order to guide the user through the payment process. This
+              includes the various attributes and internal slots of
+              <var>request</var> (some MAY be excluded for privacy reasons
+              where appropriate).
             </p>
             <p>
               The <var>acceptPromise</var> will later be resolved or rejected


ianbjacobs (commented):
Hi @marcoscaceres,

You observed that payment method specs determine "how things work." It is also the case that Payment Method Manifest will play a role in determining which payment handlers match.

Could we say, therefore:

"Determine which <a>payment handlers</a> 
support <var>identifier</var> (for example,
by looking at payment method capabilities)."

marcoscaceres:
> "Determine which payment handlers support identifier (for example, by looking at payment method capabilities)."

No. Because: At this point in the `.show()` algorithm you can only look at the PMI. Then later, when the payment sheet is actually shown, and the method is selected by the end-user, the rules for the payment method put restrictions on the payment handler's UI. 

Hopefully this screenshot from Chrome clarifies things:

screenshot_2017-08-08_14_32_32

So: 

1. PMI is "basic-card" - so the browser's Payment Handler for "basic-card" is shown.
2. Only "debit cards". 
3. Only "mastercard".

Thus, it's impossible for the user to "Pay" without meeting the condition - they can only abort.  

We can amend the end of `.show()` to talk about this. I'm currently fixing the Basic Card spec to describe these things better.

marcoscaceres:
`<rant>`
(note also that it's super terrible that `.canMakePayment()` returns "false", because it becomes a completely unreliable signal/proxy as to wether the user can make an ACTUAL payment.... which they could very well do by inputing the credit card details - this is going to make `.canMakePayment()` totally unreliable for bootstrapping users onto new payment systems, and will confuse the hell out of merchants - and it would be doubly bad if we blessed the `.show()` check, which would mean that a payment sheet would never show up for new users... but 🤷🏽‍♂️.) 
`</rant>`

marcoscaceres:
Updated the spec to more closely resemble what actually needs to happen here.

index.html
```diff
@@ -839,33 +841,33 @@
               <var>acceptPromise</var> with that error and terminate this
               algorithm.
               </li>
-              <li>Determine which <a>payment handlers</a> support
-              <var>identifier</var>. For each resulting payment handler, if
-              payment method specific capabilities supplied by the payment
-              handler match those provided by <var>data</var>, the payment
-              handler matches.
+              <li>Add each <a>payment handler</a> that supports
```
adrianhopebailie: This is much clearer wording, nice!

BUT, it does ignore the case where the browser is able to fail fast because the handler allows it to. 

The argument that the payment sheet is still shown is correct BUT that is a side-effect of how the `basic-card` payment handler is implemented in those browsers.
i.e. It claims that it supports all networks and card types so it always matches. 

Those implementations have decided that "support" does not mean "has a card loaded" but rather "is able to load a card if needed". If we want to be specific about what "support" means then I agree that should be defined in the payment method spec for that payment method. But, we still need a hook here to tell implementors that this is where they do that check.

A future payment method or a different implementation of a `basic-card` handler may have a much more definitive response when determining if the handler supports the request based on what's in the data.

(And I agree that the output of this should be consistent between `show()` and `canMakePayment()`).

Could this read:

```html
<li>For each <a>payment handler</a> that supports <var>identifier</var>:
    <ol>
        <li>If <var>data</var> indicates the <a>payment handler</a> 
        will not be able to handle the payment, continue.</li>
        <li>Else, add <a>payment handler</a> into <var>handlers</var>.</li>
    </ol>
</li>
```

adrianhopebailie:
> At this point in the .show() algorithm you can only look at the PMI.

I disagree with this. What is actually happening here is that the `basic-card` payment handler in Chrome claims support for all networks and types irrespective of what cards are loaded.

Another implementation of the handler could choose to fail fast here and indicate that it can't handle the request.

dlongley:
@adrianhopebailie,

> Another implementation of the handler could choose to fail fast here and indicate that it can't handle the request.

How `basic-card` behaves with respect to "capabilities" is also likely to be quite different from other payment methods. `basic-card` is essentially a data collection mechanism, whereas other payment methods may produce more legitimate reasons for payment handlers to be unable to provide certain capabilities. The "capability requirements" will not simply be used to restrict the payment handler UI, but rather, a payment handler will be unable to actually process the request. Allowing the user to select such a payment handler will result in an unpleasant user experience.

It would be nice to provide a generic mechanism for payment methods to use that does not require the browsers to do additional implementation for every new payment method. However, I suppose the argument could be made that the appropriate mechanism for expressing even those granular distinctions is a payment method identifier -- or put another way -- that *is inevitably* how it will happen if we don't have another mechanism in the spec.

So I would anticipate new payment method specs that have a need for this sort of thing to essentially define multiple payment method identifiers. This will produce bloat in payment request data, but I'm not sure how common this sort of thing will happen.

index.html
ianbjacobs: I agree with @adrianba's formulation from some time ago:

 1) Build a set of handlers based on PMI matching
 2) Then potentially remove that set based on some other factors

I can think of at least three factors a user agent might taken into account
to remove a payment handler:

 * Payment method specific capability matching
 * Payment method manifest statements
 * Security considerations (e.g., known malicious code or origin)
 
Marcos' statement is broad enough to accommodate these factors:
   "Add each <a>payment handler</a> that supports <var>identifier</var> into <var>handlers</var>."

But I would like to be more explicit about the factors that inform the user agent's decision.
For example:

<pre>
  <li>For each <a>payment handler</a> that supports <var>identifier</var>:
    <ol>
        <li>If the <a>payment handler</a> is not able or authorized to handle
              the payment request, continue. Note: Payment method-specific
             <var>data</var> and information provided through other mechanisms
             inform the user agent's determination.</li>
        <li>Else, add <a>payment handler</a> into <var>handlers</var>.</li>
    </ol>
  </li>
</pre>

marcoscaceres: It would be disingenuous the mention `.data` here, because the steps before this one already rejected bad data explicitly with an error. Let's please avoid making statements about how things work without solid proof that they work that way. 

Hypothetical "well, another user agent might maybe oneday kinda sorta do it differently" strawpersons are extremely unhelpful. Let's please stick to documenting reality.

marcoscaceres: > BUT, it does ignore the case where the browser is able to fail fast because the handler allows it to.

There is no browser today that fails fast. If a browser in the future does decide to fail fast, we should amend the spec then.  

Again, I feel very strongly that we should only document how things work today for the purpose of CR: We can spin up a "V2" very easily in a few weeks, the day _AFTER_ CR, if we want to add any extended behavior. If we try to capture what future browser engines will do now, we will likely get it wrong - so let's do it right when the time comes (in a "V2" please!). 

> The argument that the payment sheet is still shown is correct BUT that is a side-effect of how the basic-card payment handler is implemented in those browsers. i.e. It claims that it supports all networks and card types so it always matches.

Actually, we are both wrong (I accidentally misled you, sorry!): When Chrome doesn't know the network, it restricts to the networks it knows about (Visa, MasterCard, etc.) - while allowing any type (I was surprised by this too! 😱). 

So, if a merchant requests "this-does-not-exists" and/or "voyager", they will likely get back a "visa" card (or something else Chrome supports and the end-user has on hand). If the user doesn't have any other card on hand, they have to abort the payment request. 

I've specified the above here:
https://github.com/w3c/payment-method-basic-card/pull/39/

However, my assertion above still holds. The `.data` doesn't play a role here in today's implementations. It may in the future, but again, please please please 🎶PLLeeaAAAASssseeEE🎶, let's not add aspirational things right before CR. We should be looking at reducing the spec right now, not growing its scope.

marcoscaceres: We discussed this on today's Editor's call. @zkoch gave me a "reality check": when Chrome does work in concert with third-party payment methods (native apps in this case), they do "fail fast" at this point. Given that what I sought was confirmation that another use agent was in fact doing doing the filtering, then I concur and will add the text that @ianbjacobs suggested.

ianbjacobs: Many thanks to the editors for working this out!

Ian

marcoscaceres:
@adrianhopebailie @ianbjacobs, I've taken your suggested text, but reworded to be a `if(true) handlers.push(handler)` check, instead of an `if false, then if true`. Please review, and approve is satisfied with the text. I opted not to include the note, as I think those kinds of things are best defined in the specs of individual payment handlers.

ianbjacobs:
Hi @marcoscaceres, I'm ok with the new text. Thank you.

I would still prefer an informative note about potential sources of information, but can live with going into CR without it.

Ian

adrianhopebailie:
@marcoscaceres LGTM, thanks

marcoscaceres:
@dontcallmedom sorry to bother you, but the IPR bot has been failing for our PRs the last week or so? Not sure who to ping about it.

dontcallmedom:
@marcoscaceres I'm the right person to ask; looking at it quicky, I think that's because the repo name has changed; will fix shortly

dontcallmedom:
@marcoscaceres I think I've fixed it but somehow I'm not managing to get the latest pull request to be validated for now; if new PRs don't get IPR checked, ping me again

marcoscaceres:
@dontcallmedom thanks for fixing it. Will do.

PR #583, opened 2017-08-15
editorial: add example of .toJSON()+fetch POST

### LINKED ISSUE #576: Need short example of serializing response
Be nice to have a short example of `response.toJSON()` and `POST`ing it with `fetch()` to a server. That would "tell the complete story" of how to use the API. 

We can slot this into the existing example for payment response.

marcoscaceres:
* closes #576

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/payment-request/toJSON_example2.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/payment-request/52563ad...4e34712.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -430,6 +430,38 @@
           };
         </pre>
       </section>
+      <section data-link-for="PaymentResponse">
+        <h3>
+          POSTing payment response back to a server
+        </h3>
+        <p>
+          It's expected that data in a <a>PaymentResponse</a> will be POSTed
+          back to a server for processing. To make this as easy as possible,
+          <a>PaymentResponse</a> provides a <a>toJSON()</a> method that
+          serializes the object directly into JSON. This makes it trivial to
+          POST the resulting JSON back to a server using the <a data-cite=
+          "fetch">Fetch API</a>:
+        </p>
+        <pre class="example">
+          async function doPaymentRequest() {
+            const payRequest = new PaymentRequest(methodData, details, options);
+            const payResponse = await payRequest.show();
+            let result = "unknown";
+            try {
+              const httpResponse = await fetch("/process-payment", {
+                method: "POST",
+                headers: new Headers({ "Content-Type": "application/json" }),
+                body: payResponse.toJSON(),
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -430,6 +430,38 @@
...
+        </p>
+        <pre class="example">
+          async function doPaymentRequest() {
+            const payRequest = new PaymentRequest(methodData, details, options);
+            const payResponse = await payRequest.show();
+            let result = "unknown";
+            try {
+              const httpResponse = await fetch("/process-payment", {
+                method: "POST",
+                headers: new Headers({ "Content-Type": "application/json" }),
```
domenic: `new Headers` is not necessary; you can pass the object literal directly.

index.html
```diff
@@ -430,6 +430,38 @@
...
+                body: payResponse.toJSON(),
+              });
+              result = httpResponse.ok ? "success" : "fail";
+            } catch (err) {
+              console.error(err);
```
domenic, line 458: Should this set result = "fail"?

domenic (approved):
LGTM with potential improvements

index.html
marcoscaceres: Probably... but wanted to show the "unknown" state somehow.

domenic: Maybe do a quick parse of response.json() ? I dunno. I have a hard time understanding "unknown" in real-world systems...

marcoscaceres: yeah, me too :( Ok, will just add a comment about "unknown" that it's an option that people can use if they need it... however, we should try to come up with something real or we should mark it "at risk".

Chrome currently does nothing with "unknown" (behaves the same as "success", in that no error is shown to the user). I don't know if we (Firefox) will do anything with "unknown" at this point.

marcoscaceres: Filed https://github.com/w3c/payment-request/pull/583

marcoscaceres: Removed mention of "unknown", as I have a feeling it's it won't survive CR.

PR #590, opened 2017-08-22
Don't block UI if .updateWith() not called (closes #589)

### LINKED ISSUE #589: .updateWith() should only be callable if event was canceled
The way the spec is currently written, the payment sheet of a browser can lock up waiting on `ev.updateWith()` to be called. 

What actually needs to happen is: 

  1. That the developer needs to cancel the event somehow and then call `ev.updateWith()` - maybe .updateWith() cancels the event automatically? (I like this, because it means existing apps can continue to work without devs needing to make any changes to their existing code! And browsers need only make a small change.)
  1. the browser needs to check if the event was cancelled. If yes, wait for `.updateWith()`. If no, reuse the existing data (or present whatever data the user has changed - like a new address or shipping option). Payment sheets already do this when changing credit cards, email, phone number, for instance (where no event is fired at the page). 

So, Proposal: 

* Calling `updateWith()` sets the event's cancel flag. 
* In the sheet:
   * If the event was not cancelled, then just update reusing the existing data + whatever new data the user provide/changed.
* Otherwise, wait for `.updateWith()`'s promise to settle. 

Pre-empting discussion: checking if an event listener/handler is set on the payment request would [violate the DOM spec](https://dom.spec.whatwg.org/#observing-event-listeners) (because it would be a side-effect). That would be very bad.

marcoscaceres:
***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/payment-request/cancel_flag.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/payment-request/70a4769...40579b4.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2122,7 +2122,8 @@
         <h2>
           <dfn>methodName</dfn> attribute
         </h2>
-        <p data-tests="payment-response/methodName-attribute-manual.https.html">
+        <p data-tests=
+        "payment-response/methodName-attribute-manual.https.html">
           The <a>payment method identifier</a> for the <a>payment method</a>
           that the user selected to fulfill the transaction.
         </p>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Sorry, screwed up the commit history rebasing this, sending again.

PR #591, opened 2017-08-22
Dont block UI if .updateWith() not called (closes #589)

### LINKED ISSUE #589: .updateWith() should only be callable if event was canceled

marcoscaceres:
Tests:
w3c/web-platform-tests#7288

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/payment-request/cancel_flag.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/payment-request/2cefb58...0e8e780.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2406,13 +2406,52 @@
           <h2>
             <dfn data-lt="updateWith(detailsPromise)">updateWith()</dfn> method
           </h2>
-          <p class="note">
-            If a developer wants to update the payment request, then they need
-            to call <a>updateWith()</a> and provide a
-            <a>PaymentDetailsUpdate</a> dictionary, or a promise for one,
-            containing changed values that the <a>user agent</a> presents to
-            the user.
-          </p>
+          <aside class="note">
+            <p>
+              If a developer wants to update the payment request, then they
+              need to call <a>updateWith()</a> and provide a
+              <a>PaymentDetailsUpdate</a> dictionary, or a promise for one,
+              containing changed values that the <a>user agent</a> presents to
+              the user.
+            </p>
+            <p>
+              To prevent the user interface from blocking (and to reflect
+              changes made by the end-user through the UI), developers need to
+              immediately call <a>updateWith()</a>. If they don't,
+              <a>[[\didUpdate]]</a> becomes true and <a>updateWith()</a>
+              will throw.
+            </p>
+            <pre class="example" title="how to use `updateWith()` correctly.">
+              // ❌ Bad - this won't work!
+              request.onshippingaddresschange = async ev =&gt; {
+                // Next tick, so [[didUpdate]] is now true.
+                const details = await Promise.resolve({ ...details, ...updatedProps });
+                // 💥 Too late, will throw an "InvalidStateError"
+                ev.updateWith(details);
+              };
+
+              // ✅ Good - UI will wait
+              request.onshippingaddresschange = async ev =&gt; {
+                //  This is fine, of course.
+                const promiseForNewDetails = getNewDetails(oldDetails);
+                ev.updateWith(promiseForNewDetails);
+              };
+            </pre>
+            <p>
+              Additionally, <a>[[\didUpdate]]</a> prevents reuse of
+              <a>PaymentRequestUpdateEvent</a>.
+            </p>
+            <pre class="example" title="can only call `upadteWith()` once">
+              request.onshippingaddresschange = async ev =&gt; {
+                const p = Promise.resolve({ ...details });
+                ev.updateWith(p);
+                await p;
+                // 💥 [[didUpdate]] is true, this would throw an "InvalidStateError"
+                const q = Promise.resolve({ ...details });
+                ev.updateWith(q);
+              };
+            </pre>
+          </aside>
           <p>
             The <a>updateWith(detailsPromise)</a> method MUST act as follows:
           </p>
--- a/index.html
+++ b/index.html
@@ -2717,7 +2717,7 @@
               <ol>
                 <li>Set <var>event</var>.<a>[[\waitForUpdate]]</a> to false.
                 </li>
-                <li>Set <var>event</var>.<a>[[\didUpdate]]</a> to to true.
+                <li>Set <var>event</var>.<a>[[\didUpdate]]</a> to true.
                 </li>
                 <li>Set <var>request</var>.<a>[[\updating]]</a> to false.
                 </li>
--- a/index.html
+++ b/index.html
@@ -2424,7 +2424,7 @@
             <pre class="example" title="how to use `updateWith()` correctly.">
               // ❌ Bad - this won't work!
               request.onshippingaddresschange = async ev =&gt; {
-                // Next tick, so [[didUpdate]] is now true.
+                // `await` goes to next tick, so [[didUpdate]] is now true.
                 const details = await Promise.resolve({ ...details, ...updatedProps });
                 // 💥 Too late, throws "InvalidStateError".
                 ev.updateWith(details);
--- a/index.html
+++ b/index.html
@@ -420,8 +420,8 @@
         <pre class="example" title="Registering event handlers">
           const request = new PaymentRequest(methodData, details, options);
           // Async update to details
-          request.onshippingaddresschange = ev =&gt;{
-            ev.updateWith(checkShipping(request))
+          request.onshippingaddresschange = ev =&gt; {
+            ev.updateWith(checkShipping(request));
           };
           // Sync update to the total
           request.onshippingoptionchange = ev =&gt; {
@@ -443,7 +443,9 @@
               const { shippingOptions, total } = await calculateShipping(json);
               Object.assign(newDetails, { shippingOptions, total });
             } catch (err) {
-              Object.assign(newDetails, { error: `Sorry! we can't ship to your address.` });
+              Object.assign(newDetails, {
+                error: `Sorry! we can't ship to your address.`,
+              });
             }
             return newDetails;
           }
--- a/index.html
+++ b/index.html
@@ -2956,8 +2956,8 @@
           "!DOM#concept-event-create">creating an event</a> using
           <a>PaymentRequestUpdateEvent</a>.
           </li>
-          <li>Initialize <var>event</var>'s <a data-cite="!DOM#dom-event-type">
-            type</a> attribute to <var>name</var>.
+          <li>Initialize <var>event</var>'s <code><a data-cite=
+          "!DOM#dom-event-type">type</a></code> attribute to <var>name</var>.
           </li>
           <li>
             <a data-cite="!DOM#concept-event-dispatch">Dispatch</a>
--- a/index.html
+++ b/index.html
@@ -2426,7 +2426,7 @@
               request.onshippingaddresschange = async ev =&gt; {
                 // Next tick, so [[didUpdate]] is now true.
                 const details = await Promise.resolve({ ...details, ...updatedProps });
-                // 💥 Too late, will throw an "InvalidStateError"
+                // 💥 Too late, throws "InvalidStateError"
                 ev.updateWith(details);
               };
 
@@ -2442,14 +2442,21 @@
               <a>PaymentRequestUpdateEvent</a>.
             </p>
             <pre class="example" title="can only call `upadteWith()` once">
-              request.onshippingaddresschange = async ev =&gt; {
+               // ❌ Bad - calling updateWith() doesn't work!
+              request.addEventListener("shippingaddresschange", async ev =&gt; {
+                ev.updateWith(details); // this is ok.
+                // 💥 [[waitForUpdate]] is true, throws an "InvalidStateError"!
+                ev.updateWith(otherDetails);
+              });
+
+               // ❌ Bad - this won't work either!
+              request.addEventListener("shippingaddresschange", async ev =&gt; {
                 const p = Promise.resolve({ ...details });
                 ev.updateWith(p);
                 await p;
-                // 💥 [[didUpdate]] is true, this would throw an "InvalidStateError"
-                const q = Promise.resolve({ ...details });
-                ev.updateWith(q);
-              };
+                // 💥 [[didUpdate]] is now true, throws "InvalidStateError"!
+                ev.updateWith({ ...newDetails });
+              });
             </pre>
           </aside>
           <p>
--- a/index.html
+++ b/index.html
@@ -2914,9 +2914,8 @@
           fired. Wait for <var>event</var>.<a>updateWith()</a> method's
           <var>detailsPromise</var> to settle.
           </li>
-          <li>Otherwise, <var>event</var>.<a>[[\waitForUpdate]]</a> is false,
-          then set <var>event</var>.<a>[[\didUpdate]]</a> to true and update
-          the user interface based on any changed values.
+          <li>Otherwise, set <var>event</var>.<a>[[\didUpdate]]</a> to true and
+          update the user interface based on any changed values.
           </li>
         </ol>
       </section>
--- a/index.html
+++ b/index.html
@@ -2393,12 +2393,6 @@
           details of the payment request in response to a user interaction.
         </p>
         <p>
-          If a developer wants to update the payment request, then they need to
-          call <a>updateWith()</a> and provide a <a>PaymentDetailsUpdate</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2417,6 +2422,9 @@
             <li>If <var>event</var>'s <a>isTrusted</a> attribute is false, then
             then <a>throw</a> a "<a>InvalidStateError</a>" <a>DOMException</a>.
             </li>
+            <li>If <var>event</var>.<a>[[\didUpdate]]</a> is true, then
```
domenic: So this would happen if someone calls updateWith(p), p settles, and then someone calls updateWith(q), right?

Whereas the next line would happen if someone calls updateWith(p), and before p settles, someone calls updateWith(q)?

If I got that right, then I'd suggest adding quick examples or notes inside these steps re-stating the above.

index.html
```diff
@@ -2880,9 +2893,18 @@
           further action. The <a>user agent</a> user interface should ensure
           that this never occurs.
           </li>
-          <li>
-            <a>Fire an event</a> named <var>name</var> at <var>request</var>
-            using <a>PaymentRequestUpdateEvent</a>.
+          <li>Let <var>event</var> be the result of <a>fire an event</a> named
```
domenic: This doesn't work. Fire an event returns a boolean, not an event.

You should instead:

- Let _event_ be the result of [creating an event](https://dom.spec.whatwg.org/#concept-event-create) using Event.
- Initialize _event_'s type attribute to _name_
- [Dispatch](https://dom.spec.whatwg.org/#concept-event-dispatch) _event_ at _target_.

index.html
```diff
@@ -2880,9 +2893,18 @@
...
+          <var>name</var> at <var>request</var> using
+          <a>PaymentRequestUpdateEvent</a>.
+          </li>
+          <li data-link-for="PaymentRequestUpdateEvent">If
+          <var>event</var>.<a>[[\waitForUpdate]]</a> is true, disable any part
+          of the user interface that could cause another update event to be
+          fired. Wait for <var>event</var>.<a>updateWith()</a> method's
+          <var>detailsPromise</var> to settle.
```
domenic: ", then re-enable the relevant parts of user interface", I think.

Maybe this should be a series of nested steps under the "If ... is true".

domenic (changes requested):

index.html
marcoscaceres: > This doesn't work. Fire an event returns a boolean, not an event.

Argh, I should have remembered that. I've used that boolean before elsewhere.

index.html
marcoscaceres: `.updateWith()` controls re-enabling the user interface, in step 14. It might be redundant to re-state it here.

domenic: Ah ok, in that case, agreed.

index.html
marcoscaceres: > So this would happen if someone calls updateWith(p), p settles, and then someone calls updateWith(q), right?

I was thinking of this case: 

```JS
request.onaddresschange  = ev => { 
   Promise.resolve({...details})
     // next tick, so [[didUpdate]] is now true. Throws InvalidStateError.
     .then(ev.updateWith.bind(ev));
}
```

>Whereas the next line would happen if someone calls updateWith(p), and before p settles, someone calls updateWith(q)?

Correct. It also prevents someone holding a reference to event and calling `.updateWith(p)` after `p` has settled. The problem in the current spec is that [[\waitForUpdate]] and request.[[updating]] are both reset to `false` in step 14 of `updateWith()` - so theoretically, a developer could have just waited like this:  

```JS
request.onaddresschange  = async ev => {
   const p =  Promise.resolve({...details});
   ev.updateWith(p);
   await p;  // [[waitForUpdate]] and request.[[updating]] are both reset to `false`! 
   // [[didUpdate]] is true, this now throws
   const q = Promise.resolve({...details});
   ev.updateWith(q);
}
```

marcoscaceres: Fixed first case above...

marcoscaceres: Ok, added example above.

index.html
```diff
@@ -2400,13 +2405,59 @@
...
-          <p class="note">
-            If a developer wants to update the payment request, then they need
-            to call <a>updateWith()</a> and provide a
-            <a>PaymentDetailsUpdate</a> dictionary, or a promise for one,
-            containing changed values that the <a>user agent</a> presents to
-            the user.
-          </p>
+          <aside class="note">
+            <p>
+              If a developer wants to update the payment request, then they
```
domenic, line 2405: I'm just noticing that this repeats text from slightly above. Let's delete the above text, as it feels pretty out of place anyway.

index.html
```diff
@@ -2400,13 +2405,59 @@
...
+              // ❌ Bad - this won't work!
+              request.onshippingaddresschange = async ev =&gt; {
+                // `await` goes to next tick, so [[didUpdate]] is now true.
+                const details = await Promise.resolve({ ...details, ...updatedProps });
+                // 💥 Too late, throws "InvalidStateError".
+                ev.updateWith(details);
+              };
+
+              // ✅ Good - UI will wait.
+              request.onshippingaddresschange = async ev =&gt; {
```
domenic: No `async` here

index.html
```diff
@@ -2400,13 +2405,59 @@
...
+                //  This is fine, of course.
```
domenic: "of course" is a bit aggressive and implies the reader understands what's going on. You shouldn't really say "of course" while teaching someone something like this, IMO.

index.html
```diff
@@ -2400,13 +2405,59 @@
...
+              changes made by the end-user through the UI), developers need to
+              immediately call <a>updateWith()</a>. If they don't,
+              <a>[[\didUpdate]]</a> becomes true and <a>updateWith()</a>
+              will throw.
+            </p>
+            <pre class="example" title="how to use `updateWith()` correctly.">
+              // ❌ Bad - this won't work!
+              request.onshippingaddresschange = async ev =&gt; {
+                // `await` goes to next tick, so [[didUpdate]] is now true.
+                const details = await Promise.resolve({ ...details, ...updatedProps });
```
domenic: Let's do `await getNewDetails(oldDetails)` to mirror the below.

index.html
```diff
@@ -2400,13 +2405,59 @@
...
+              the user.
+            </p>
+            <p>
+              To prevent the user interface from blocking (and to reflect
+              changes made by the end-user through the UI), developers need to
+              immediately call <a>updateWith()</a>. If they don't,
+              <a>[[\didUpdate]]</a> becomes true and <a>updateWith()</a>
+              will throw.
+            </p>
+            <pre class="example" title="how to use `updateWith()` correctly.">
```
domenic: Backticks don't work in example titles. Respec bug? Or just remove it for now?

Same below.

index.html
```diff
@@ -2400,13 +2405,59 @@
...
+                //  This is fine, of course.
+                const promiseForNewDetails = getNewDetails(oldDetails);
+                ev.updateWith(promiseForNewDetails);
+              };
+            </pre>
+            <p>
+              Additionally, <a>[[\didUpdate]]</a> prevents reuse of
+              <a>PaymentRequestUpdateEvent</a>.
+            </p>
+            <pre class="example" title="can only call `upadteWith()` once">
```
domenic: In addition to the backticks, updateWith() is misspelled

index.html
```diff
@@ -2400,13 +2405,59 @@
...
+
+              // ✅ Good - UI will wait.
+              request.onshippingaddresschange = async ev =&gt; {
+                //  This is fine, of course.
+                const promiseForNewDetails = getNewDetails(oldDetails);
+                ev.updateWith(promiseForNewDetails);
+              };
+            </pre>
+            <p>
+              Additionally, <a>[[\didUpdate]]</a> prevents reuse of
```
domenic: [[didUpdate]] and [[waitForUpdate]] together prevent reuse

index.html
```diff
@@ -2400,13 +2405,59 @@
...
+              <a>PaymentRequestUpdateEvent</a>.
+            </p>
+            <pre class="example" title="can only call `upadteWith()` once">
+              // ❌ Bad - calling updateWith() doesn't work!
+              request.addEventListener("shippingaddresschange", async ev =&gt; {
```
domenic: no `async`

index.html
```diff
@@ -2400,13 +2405,59 @@
...
+              To prevent the user interface from blocking (and to reflect
+              changes made by the end-user through the UI), developers need to
+              immediately call <a>updateWith()</a>. If they don't,
+              <a>[[\didUpdate]]</a> becomes true and <a>updateWith()</a>
+              will throw.
+            </p>
+            <pre class="example" title="how to use `updateWith()` correctly.">
+              // ❌ Bad - this won't work!
+              request.onshippingaddresschange = async ev =&gt; {
+                // `await` goes to next tick, so [[didUpdate]] is now true.
```
domenic: To distinguish this from the last example, we should say something like "await goes to next tick, and updateWith() was not called before that, so [[didUpdate]] gets set to true".

Although that calls into question whether didUpdate is the right name, since nothing actually updated.

index.html
```diff
@@ -2400,13 +2405,59 @@
...
+                // 💥 [[waitForUpdate]] is true, throws an "InvalidStateError".
+                ev.updateWith(otherDetails);
+              });
+
+              // ❌ Bad - this won't work either!
+              request.addEventListener("shippingaddresschange", async ev =&gt; {
+                const p = Promise.resolve({ ...details });
+                ev.updateWith(p);
+                await p;
+                // 💥 [[didUpdate]] is now true, throws "InvalidStateError".
```
domenic: To distinguish this from the first example, we can say "the update has now completed, so [[didUpdate]] gets set to true; this throws "InvalidStateError".

index.html
```diff
@@ -2880,9 +2939,24 @@
           further action. The <a>user agent</a> user interface should ensure
           that this never occurs.
           </li>
+          <li>Let <var>event</var> be the result of <a data-cite=
+          "!DOM#concept-event-create">creating an event</a> using
+          <a>PaymentRequestUpdateEvent</a>.
+          </li>
+          <li>Initialize <var>event</var>'s type attribute to <var>name</var>.
```
domenic: Ideally link type, or at least `<code>`ify it.

index.html
```diff
@@ -2880,9 +2939,24 @@
...
-            <a>Fire an event</a> named <var>name</var> at <var>request</var>
-            using <a>PaymentRequestUpdateEvent</a>.
+            <a data-cite="!DOM#concept-event-dispatch">Dispatch</a>
+            <var>event</var> at <var>request</var>.
+          </li>
+          <li data-link-for="PaymentRequestUpdateEvent">If
+          <var>event</var>.<a>[[\waitForUpdate]]</a> is true, disable any part
+          of the user interface that could cause another update event to be
+          fired. Wait for <var>event</var>.<a>updateWith()</a> method's
+          <var>detailsPromise</var> to settle.
```
domenic: event.updateWith()'s detailsPromise is pretty terrible. Can we store it in an internal slot?

Actually, what is the point of this "wait" step? Nothing happens after it. I think it should just be removed.

zkoch:
Is there a TL;DR for what we're trying to accomplish here? It's not obvious to me. :)

marcoscaceres:
@zkoch, tldr:

 1. Don't block the payment sheet if the developer doesn't call `.updateWith()`. 
 1. Don't allow developers to keep recycling the event (e.g., keep a reference to a shipping change event, and then use it to update the the payment sheet when an option change event happens - see code below).  

===

Note to self, add this test: 
```js
var methods = [{ supportedMethods: ["basic-card"] }];
var details = {
  total: { label: "", amount: { currency: "USD", value: "1" } },
};
var shippingOptions = [
  {
    id: "standard",
    label: "🚛 Ground Shipping (2 days)",
    amount: { currency: "USD", value: "5.00" },
    selected: true,
  },
  {
    id: "drone",
    label: "🚀 Drone Express (2 hours)",
    amount: { currency: "USD", value: "25.00" },
  },
];
details.shippingOptions = shippingOptions;
var r = new PaymentRequest(methods, details, { requestShipping: true });
r.addEventListener(
  "shippingaddresschange",
  e => {
    ev = e;
    ev.updateWith(details);
    var listener = ev2 => {
      var newDetails = { ...details };
      newDetails.total.label = "FAIL" + Math.random();
      ev.updateWith(newDetails);
    };
    r.addEventListener("shippingoptionchange", listener);
    r.addEventListener("shippingaddresschange", listener);
  },
  { once: true }
);
r.show();
```

index.html
marcoscaceres: I know... "didUpdate" is not a great name... I wanted something that denoted "consumed" or "expired" or "this can't be used again". I'm open to suggestions.

dlongley: Could flip it and go with "canUpdate" or "updatePermitted" and turn it off after the opportunity to update is lost or has already been taken.

index.html
marcoscaceres: Agree. I wasn't sure if it would be clear. But if one follows the updateWith() path, it provides all the steps there.

index.html
marcoscaceres: I plan to add support for this in ReSpec, so will leave it for now and we will get this for free soon.

index.html
marcoscaceres: > Although that calls into question whether didUpdate is the right name, since nothing actually updated.

Rethinking this: it actually did update... but instead of using the developer's `details`, the sheet used the details that were provided by the payment sheet itself. So maybe `updatedUI` might be better.
 
@dlongley, I'd prefer not to flip it. I don't want to rewrite it :(

index.html
marcoscaceres: deleted the one above.

index.html
```diff
@@ -2880,9 +2935,24 @@
           further action. The <a>user agent</a> user interface should ensure
           that this never occurs.
           </li>
+          <li>Let <var>event</var> be the result of <a data-cite=
+          "!DOM#concept-event-create">creating an event</a> using
+          <a>PaymentRequestUpdateEvent</a>.
+          </li>
+          <li>Initialize <var>event</var>'s <a data-cite="!DOM#dom-event-type">
+            type</a> attribute to <var>name</var>.
```
domenic: Can this be code-ified *and* linked? :)

domenic (approved):
LGTM with one remaining nit; please don't wait on another round of review before merging :).

adrianba:
I think the current spec already blocks this. The waitForUpdate slot is essentially equivalent to the new slot introduced. To avoid re-use of events, we should remove step 14, part 1, which sets waitForUpdate back to false.

marcoscaceres:
> I think the current spec already blocks this. The waitForUpdate slot is essentially equivalent to the new slot introduced. To avoid re-use of events, we should remove step 14, part 1, which sets waitForUpdate back to false.

Nice!

marcoscaceres:
@adrianba, I removed the new slot as you suggested. Adding appropriate tests and see if this works. Will carry over domenic's R+ if it works ok.

marcoscaceres:
Rebased.

marcoscaceres:
Moved test requirement to WPT repo.
https://github.com/w3c/web-platform-tests/issues/7288

I've got this mostly tested locally. Tests will hopefully land in the next few days. 

I feel confident merging this now having tested it.

Issue #603, opened 2017-09-05
Creation and assignment of PaymentAddress is underspecified  


marcoscaceres:
The spec says:

> Set the shippingAddress attribute on request to the shipping address provided by the user.

This is woefully inadequate. 

It needs to construct and instances, perform canonicalization and various attribute values, set missing things to the empty string, etc.

PR #604, opened 2017-09-05
Fix PaymentAddress attr definition (closes #601)

### LINKED ISSUE #601: PaymentAddress attribute definition is wrong
Spec currently says: 

> This is the [CLDR] (Common Locale Data Repository) region code. For example, US, GB, CN, or JP.

This should actually be ISO 3166-1 alpha-2 codes. 

It should also state that the user agent needs to convert the code to upper case.

marcoscaceres:
***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/payment-request/country_codes.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/payment-request/3fb1879...1ddbf42.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1901,8 +1901,8 @@
           <dfn>country</dfn> attribute
         </h2>
         <p>
-          This is the [[!CLDR]] (Common Locale Data Repository) region code.
-          For example, US, GB, CN, or JP.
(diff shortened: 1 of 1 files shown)


domenic (approved):

marcoscaceres:
Will test in #603

Issue #606, opened 2017-09-05
PayerAddress.phone and PayerResponse.payerPhone


marcoscaceres:
We should explicitly set `PayerAddress.phone` to empty string when `requestPayerPhone` is false. 

Otherwise, it can leak through accidentally in #603.

rsolomakhin:
I don't see `PayerAddress` in the spec. Can you elaborate on this?

marcoscaceres:
Sorry, "PaymentAddress" I mean. Been looking at this spec for too long.

rsolomakhin:
Shipping address phone number may be required for the carrier to resolve delivery problems. If you're shipping something as a gift, then the shipping address phone number may not be the same as the payer phone. Therefore, I believe that `requestPayerPhone` should not affect `PaymentAddress.phone`.

marcoscaceres:
I see. I'd still like to somehow make it clear the a user's phone number doesn't accidentally get leaked tho.

zkoch:
I agree with Rouslan. I'm not sure I see the issue here. A phone number is part of a shipping address. The user will have to select this for it to get passed back to the merchant. This, to me, is WAI.

marcoscaceres:
Agree. They are separate things that sometimes overlap.

PR #607, opened 2017-09-05
Construct and set attrs on PaymentAddress

### LINKED ISSUE #603: Creation and assignment of PaymentAddress is underspecified

marcoscaceres:
* closes #603 #606

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/payment-request/paymentaddress.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/payment-request/45ef36e...4309348.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1901,8 +1901,8 @@
           <dfn>country</dfn> attribute
         </h2>
         <p>
-          When getting, returns the <a>PaymentAddress</a>'s <a>[[\country]]</a>
-          internal slot.
+          When getting, returns the value of the <a>PaymentAddress</a>'s
+          <a>[[\country]]</a> internal slot.
         </p>
       </section>
       <section>
@@ -1910,7 +1910,7 @@
           <dfn>addressLine</dfn> attribute
         </h2>
         <p>
-          When getting, returns the <a>PaymentAddress</a>'s
+          When getting, returns the value of the <a>PaymentAddress</a>'s
           <a>[[\addressLine]]</a> internal slot.
         </p>
       </section>
@@ -1919,8 +1919,8 @@
           <dfn>region</dfn> attribute
         </h2>
         <p>
-          When getting, returns the <a>PaymentAddress</a>'s <a>[[\region]]</a>
-          internal slot.
+          When getting, returns the value of the <a>PaymentAddress</a>'s
+          <a>[[\region]]</a> internal slot.
         </p>
       </section>
       <section>
@@ -1928,8 +1928,8 @@
           <dfn>city</dfn> attribute
         </h2>
         <p>
-          When getting, returns the <a>PaymentAddress</a>'s <a>[[\city]]</a>
-          internal slot.
+          When getting, returns the value of the <a>PaymentAddress</a>'s
+          <a>[[\city]]</a> internal slot.
         </p>
       </section>
       <section>
@@ -1937,7 +1937,7 @@
           <dfn>dependentLocality</dfn> attribute
         </h2>
         <p>
-          When getting, returns the <a>PaymentAddress</a>'s
+          When getting, returns the value of the <a>PaymentAddress</a>'s
           <a>[[\dependentLocality]]</a> internal slot.
         </p>
       </section>
@@ -1946,7 +1946,7 @@
           <dfn>postalCode</dfn> attribute
         </h2>
         <p>
-          When getting, returns the <a>PaymentAddress</a>'s
+          When getting, returns the value of the <a>PaymentAddress</a>'s
           <a>[[\postalCode]]</a> internal slot.
         </p>
       </section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2829,10 +2829,66 @@
...
+              none was provided, set it to the empty <a data-cite=
+              "!WEBIDL#dfn-frozen-array-type">frozen array</a>.
+                <aside class="note">
+                  How to split an address line is locale dependent and beyond
+                  the scope of this specification.
+                </aside>
+              </li>
+              <li>Set the <a>country</a> attribute of <var>address</var> to the
+              end user's country as an upper case [[!ISO3166]] alpha-2 code or
+              to the empty string if none was provided.
```
domenic: Nit: comma after "code" reads nicer.

index.html
```diff
@@ -2829,10 +2829,66 @@
           <li>
             <a>Queue a task</a> on the <a>user interaction task source</a> to
             run the following steps:
-            <ol>
+            <ol data-link-for="PaymentAddress">
+              <li>Let <var>address</var> be a new instance of
+              <a>PaymentAddress</a>.
+              </li>
+              <li>Set the <a>addressLine</a> attribute of <var>address</var> to
```
domenic: You cannot set attributes that are readonly. There need to be internal slots for all of these, and then the attributes need to have normative text that their getters return the value of the internal slot.

This only slipped by until now because this whole thing was underspecified. Thanks for catching it :)

index.html
```diff
@@ -2829,10 +2829,66 @@
...
+              "PaymentOptions.requestPayerPhone">requestPayerPhone</a> is
+              false, set the <a>phone</a> attribute of <var>address</var> to
+              the empty string. Otherwise, set the <a>phone</a> attribute of
+              <var>address</var> to a user provided phone number, optionally
+              formatted to adhere to [[!E.164]], or to the empty string if none
+              was provided.
+              </li>
+              <li>Set the <a>languageCode</a> attribute of <var>address</var>
+              to a <a data-cite="!BCP47#section-4.5">canonicalized language
+              tag</a>, or to the empty sting if none was provided.
```
domenic: "sting" -> "string"

Do we really expect a language to be "provided"? We might want to expand on this somehow.

domenic (changes requested):

index.html
marcoscaceres: So, this opened this interesting can of worms: https://github.com/w3c/payment-request/issues/608 ... it might be simple to solve, but I need to look in detail at how Chrome implements this via the Unicode CLDR library. 

This area is new to me though, so will need to do a bit of research.

index.html
marcoscaceres: Good point. I'll add the slots.

index.html
domenic: Yeah... maybe mark it with a warning class or something linking to that issue in the meantime?

index.html
```diff
@@ -2829,10 +2945,70 @@
...
+              <var>address</var> to the empty string. Otherwise, set the
+              <a>[[\phone]]</a> internal slot of <var>address</var> to a user
+              provided phone number, optionally formatted to adhere to
+              [[!E.164]], or to the empty string if none was provided.
+              </li>
+              <li>Set the <a>[[\languageCode]]</a> internal slot of
+              <var>address</var> to a <a data-cite=
+              "!BCP47#section-4.5">canonicalized language tag</a>, or to the
+              empty sting if none was provided.
+                <div class="issue" data-number="608"></div>
```
marcoscaceres: I'll make the issue description better before merging.

index.html
```diff
@@ -1901,100 +1901,216 @@
...
+        <h2>
+          Internal slots
+        </h2>
+        <table>
+          <tr>
+            <th>
+              Internal slot
+            </th>
+            <th>
+              Description
```
domenic: I like putting "Description (non-normative)" to be extra clear.

index.html
```diff
@@ -1901,100 +1901,216 @@
           <dfn>country</dfn> attribute
         </h2>
         <p>
```
domenic, line 1903: Above this we have "When called, run" for toJSON. To stay consistent in the tenses used, we should change that to "When called, runs".

index.html
```diff
@@ -2829,10 +2945,70 @@
...
+              provided phone number, optionally formatted to adhere to
+              [[!E.164]], or to the empty string if none was provided.
+              </li>
+              <li>Set the <a>[[\languageCode]]</a> internal slot of
+              <var>address</var> to a <a data-cite=
+              "!BCP47#section-4.5">canonicalized language tag</a>, or to the
+              empty sting if none was provided.
+                <div class="issue" data-number="608"></div>
+              </li>
+              <li>Set the <a>[[\city]]</a> internal slot of <var>address</var>
```
domenic: We've used _address_.[[city]] before; I think it's best to be consistent.

domenic (approved):
LGTM with editorial suggestions. Great stuff.

index.html
marcoscaceres: fixed both .toJSON() instances

PR #610, opened 2017-09-05
Only select shippingOption if asked to (closes #609)

### LINKED ISSUE #609: If requestShipping is false, pr.shippingOption should be null
During testing, a weird case came up. 

If you have:

```JS 
const options = {
   requestShipping: false
};

const someShippingOption = {
   id: "never-selected",
   selected: true,
  // other props 
};

const details = {
   //... things
   shippingOptions: [someShippingOption]
}

const pr = new PaymentRequest(methods, details, options); 
console.assert(pr.shippingOption === "never-selected");
```

That seems bad. If no shipping was requested by the merchant, then the selected shipping option should actually be `null`, no? Even if the merchant screwed up and added shipping options that can't ever actually be selected by the end user (there is no UI to make the selection).

marcoscaceres:
Test: https://github.com/w3c/web-platform-tests/pull/7239

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/payment-request/null_shipping_option.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/payment-request/58c6867...231a79a.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -623,7 +623,9 @@
           </li>
           <li>Let <var>selectedShippingOption</var> be null.
           </li>
-          <li>Process shipping options:
+          <li>If the <a data-lt=
+          "PaymentOptions.requestShipping">requestShipping</a> member of <var>
+            options</var> is present and set to true, process shipping options:
             <ol>
               <li>Let <var>options</var> be an empty <code><a data-cite=
               "!WEBIDL#idl-sequence">sequence</a></code>&lt;<a>PaymentShippingOption</a>&gt;.


domenic (approved):
This was kind of tricky to trace all the implications of, but I've convinced myself this is in fact the right fix. Nice work.

marcoscaceres:
[Test added and reviewed](https://github.com/w3c/web-platform-tests/pull/7239). Merging.

marcoscaceres:
Actually, will add one more constructor test.

PR #631, opened 2017-09-21
UAs should not alter value string (addresses #624)

### LINKED ISSUE #624: Are integer values for PaymentCurrencyAmount.value left as integers?
Hi @marcoscaceres, @rsolomakhin, @adrianba,

I was asked today whether integer values (PaymentCurrencyAmount.value) are preserved as integers, or whether ".00" is appended by implementations. 

I don't see anything in the algorithm [1] about this type of conversion. I think the desired behavior is "if you get an integer, leave it as an integer."

Is that how implementations behave today? 
Should the spec say anything different?

Thanks!
Ian

[1] https://w3c.github.io/payment-request/#validity-checkers

ianbjacobs:
Closes #624

* UAs should not alter values provided as input.
* UAs may display values according to user's local settings

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/payment-request/no_touch_currency_vals.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/payment-request/7e4a2ca...8de1e7f.html)


index.html
```diff
@@ -1391,6 +1391,9 @@
         </dt>
         <dd>
           A <a>valid decimal monetary value</a> containing a monetary amount.
+          User agents should not alter values provided as input; for example,
```
marcoscaceres: As I mentioned in the bug, it's a "statement of fact" that UAs can present the value differently - but this doesn't belong here. It belongs in the "currency member" description. 

I suggest we add the following, underneath  "A user agent SHOULD use U+00A4 (¤) for formatting.": 

```
Users agents MAY format the display of the <a>currency</a> member 
to adhere to OS conventions (e.g., for localization purposes).
```

marcoscaceres (changes requested):

index.html
marcoscaceres: Ok, but this should be removed now :) We definitely don't want this SHOULD here.

ianbjacobs: Hi @marcoscaceres,

Do you mean the same sentence belongs somewhere else? That would be fine. But I hope we get to keep it or something close.

Ian

marcoscaceres: No, I mean it should be removed entirely. The algorithms already handle all this stuff... saying "...SHOULD NOT..." here would lead to confusion.

marcoscaceres: To be clear, if we were to leave the "SHOULD NOT", it opens the door to the possibility that an implementer could change the value... that would be very very bad and would lead to all sorts of interop chaos. 

Instead, implementers should follow what the algorithms says - which make it clear, if you follow them, that the value never gets changed.

ianbjacobs: I think it would be helpful to include an explicit note to that effect. I will propose something in an updated PR.

index.html
```diff
@@ -1478,6 +1476,13 @@
           a total can't be a negative number.
           </li>
         </ol>
+        <div class="note" title="No alteration of values">
+          <p>
+            One feature of this algorithm is that user agents do not alter the
```
marcoscaceres: How about: 

```HTML
The algorithm does not alter or canonicalize the <var>amount</var>.<a>value</a>.  
For example, a user agent will not change "55" into "55.00". 
Payment handlers need to be prepared to deal with such values. 
```

marcoscaceres: Or maybe something about being able to deal with variants (i.e., "55", "55.0", and "55.0000000000" are all be equivalent, even if presented as `US$55.00`).

index.html
```diff
@@ -1474,6 +1476,13 @@
           a total can't be a negative number.
           </li>
         </ol>
+        <div class="note" title="No alteration of values">
```
marcoscaceres: nit: s/`div`/`aside`

marcoscaceres (approved):
Enduring my nit picking is why they pay you the big bucks, @ianbjacobs 🤑

I think this still needs a tidy 😉

marcoscaceres:
Ah no, just checked... no tidy needed.

PR #656, opened 2017-11-24
Add localization hint for payment sheet

### LINKED ISSUE #650: Allow locale to be specified by the developer.
It would be helpful for merchants to be able to specify the locale that the user agent displays the payment sheet in.

Especially as we're discussing custom error messages, which are merchant-supplied, it is really awkward if the message the merchant supplies doesn't match the language (and amount / currency display) in the sheet. 

An additional data point we have from Stripe Checkout is that localization only helps conversion if the merchant page and the payment modal are displayed in the same locale.

marcoscaceres:
* Adds "lang" member to PaymentRequestInit dictionary
* Adds processing model for `lang`
* Optionally allows UA to localize the payment sheet

closes #650

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/656.html" title="Last updated on Jan 29, 2018, 6:21 AM GMT (8eaebb0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/656/c1fa933...8eaebb0.html" title="Last updated on Jan 29, 2018, 6:21 AM GMT (8eaebb0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -751,7 +751,23 @@
               </li>
             </ol>
           </li>
+          <li>Let <var>uiLang</var> be null.
+          </li>
+          <li data-link-for="PaymentDetailsInit">If
+          <var>details</var>["<a>lang</a>"] is present:
+            <ol>
+              <li>If <a data-cite=
+              "ecma-402#sec-isstructurallyvalidlanguagetag">IsStructurallyValidLanguageTag</a>(<var>details</var>["<a>lang</a>"])
+              is false, throw a <a>RangeError</a>.
+              </li>
+              <li>Set <var>uiLang</var> to <a data-cite=
+              "ecma-402#sec-canonicalizelanguagetag">CanonicalizeLanguageTag</a>(<var>details</var>["<a>lang</a>"]).
+              </li>
+            </ol>
+          </li>
           <li>Let <var>request</var> be a new <a>PaymentRequest</a>.
+          </li>
+          <li>Set <var>request</var>.<a>[[\uiLang]]</a> to <var>uiLang</var>.
           </li>
           <li>Set <var>request</var>.<a>[[\options]]</a> to <var>options</var>.
           </li>
@@ -937,6 +953,15 @@
               Otherwise, present a user interface to allow the user to interact
               with the <var>handlers</var>. The user agent SHOULD prioritize
               the preference of the user when presenting payment methods.
+            </p>
+            <p>
+              Optionally, if <var>request</var>.<a>[[\uiLang]]</a> is not null,
+              localize the user interface to match, as closely as possible, the
+              language of <a>[[\uiLang]]</a>. Otherwise, if
+              <var>request</var>.<a>[[\uiLang]]</a> is null, it is RECOMMENDED
+              that the language of the user interface match the <a data-cite=
+              "!HTML#language">language</a> of <a data-cite=
+              "html#the-body-element-2">the body element</a>.
             </p>
             <p>
               For the <a>payment handler</a> selected by the end-user, the user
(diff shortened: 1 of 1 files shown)


jarrenwe26:
help im being contorld by hackers and our young family

index.html
```diff
@@ -935,8 +951,11 @@ <h2>
           <li>
             <p>
               Otherwise, present a user interface to allow the user to interact
-              with the <var>handlers</var>. The user agent SHOULD prioritize
-              the preference of the user when presenting payment methods.
+              with the <var>handlers</var>. Optionally, if
+              <var>request</var>.<a>[[\uiLang]]</a> is not null, localize the
+              user interface to match, as closely as possible, the language of
+              <a>[[\uiLang]]</a>. The user agent SHOULD prioritize the
```
domenic: I think you should paragraph break before the last sentence as now it's somewhat disconnected from the rest.

domenic: Should we state more about what to do if [[uiLang]] is null? For example using the [language](https://html.spec.whatwg.org/#language) of [the body element](https://html.spec.whatwg.org/#the-body-element-2)?

domenic (approved):
LGTM with a nit and a suggestion for expansion. Although, just using the language of the body element might suffice for #650 without any new APIs...

index.html
marcoscaceres: Yeah, that might be good to recommend using the document's language.

marcoscaceres:
> Although, just using the language of the body element might suffice for #650 without any new APIs...

Having both the API hint and falling back to body's lang seems like it would cover all cases. I don't have a strong opinion, and could definitely live with just "use the document's lang". @michelle, can you think of cases where it would be challenging to just use the document's language? 

Having said that, implementation support for localizing the UI will remain a challenge for implementers, as we don't generally ship support for multiple UI languages. So we will have to see how much demand there is for this.

index.html
```diff
@@ -938,6 +954,15 @@ <h2>
               with the <var>handlers</var>. The user agent SHOULD prioritize
               the preference of the user when presenting payment methods.
             </p>
+            <p>
+              Optionally, if <var>request</var>.<a>[[\uiLang]]</a> is not null,
+              localize the user interface to match, as closely as possible, the
+              language of <a>[[\uiLang]]</a>. Otherwise, if
+              <var>request</var>.<a>[[\uiLang]]</a> is null, is it RECOMMEDED
```
stpeter: s/is it/it is/

index.html
```diff
@@ -1584,7 +1619,17 @@ <h2>
           <a>PaymentDetailsBase</a> dictionary, the following members are part
           of the <a>PaymentDetailsInit</a> dictionary:
         </p>
-        <dl>
+        <dl data-sort="">
+          <dt>
+            <dfn>lang</dfn> member
+          </dt>
+          <dd>
+            A [[!BCP47]] language tag representing the preferred language of
```
stpeter: s/preferred/merchant's preferred/ (what the merchant prefers might not be what the buyer prefers)

stpeter:
The `Accept-Language` HTTP header might also be communicated from the user's browser to the merchant website. See https://www.w3.org/International/questions/qa-when-lang-neg regarding use of this header for language negotiation. For example, the hint that the merchant sends could be constrained to one of the languages that is in the `Accept-Language` set. See also https://www.w3.org/International/wiki/Locale-based_forms for related discussion regarding number and date formats (which could be important for payment totals and ship dates).

marcoscaceres:
It might be worth looking at what effect, if any, `Content-Language` has on "the body element" in HTML.

marcoscaceres:
@domenic, changed as you suggested (no API). 

@michelle, can you live with this for now and we see how we go?

domenic (approved):

PR #666, opened 2018-01-17
Feat: adds PaymentItemType enum and PaymentItem.type

### LINKED ISSUE #163: Add type to PaymentItem
Before I spend the time to put out a PR on this, I want to get some initial feedback. I think there are many kinds of PaymentItems that could be interesting/impactful to the payment app. I still like calling out the total as a separate item, but I would like to propose we go a step further and give all PaymentItems a type. 

The PaymentItemType enum would provide a way to assign specific meaning/semantics for common use cases, e.g. taxes, discounts, etc. We would control this enum and make additions to it as needed. The enum would have a fallback value that means the PaymentItem is purely for display and has no significant meaning beyond the merchant's purposes.

What do you think? With moderate approval I will code it up.

marcoscaceres:
* Tests: https://github.com/w3c/web-platform-tests/pull/9099
* dev docs: 

Adds optional "type" member `PaymentItem` and introduces `PaymentItemType` enum. 

I'm thinking we should introduce "discount" or similar when we add coupon support.  

@domenic, I'm wondering if there would be any advantage to having a default value for the `type` member (e.g., "other")? Thought I'd ask just in case.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/666.html" title="Last updated on Feb 21, 2018, 11:31 PM GMT (a5dbf13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/666/678a110...a5dbf13.html" title="Last updated on Feb 21, 2018, 11:31 PM GMT (a5dbf13)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1885,7 +1885,7 @@
         <dt>
           <dfn>type</dfn> member
         </dt>
-        <dd>
+        <dd data-tests="PaymentItem/type_member.https.html">
           A <a>PaymentItemType</a> enum value or null. A user agent MAY use the
           value of <a>type</a> to assist in the presentation of
           <a>PaymentItem</a> by, for example, visually grouping types together
--- a/index.html
+++ b/index.html
@@ -1886,11 +1886,12 @@
           <dfn>type</dfn> member
         </dt>
         <dd data-tests="PaymentItem/type_member.https.html">
-          A <a>PaymentItemType</a> enum value. A user agent MAY use the value
-          of <a>type</a> to assist in the presentation of <a>PaymentItem</a>
-          by, for example, visually grouping types together or other otherwise
-          distinguishing them from other types (or from items that have no
-          associated type).
+          A <a>PaymentItemType</a> enum value, which a developer can use to
+          explicitly indicate that this member is of a particular type. A user
+          agent MAY use the value of <a>type</a> to assist in the presentation
+          of <a>PaymentItem</a> by, for example, visually grouping types
+          together or other otherwise distinguishing them from other types (or
+          from items that have no associated type).
         </dd>
       </dl>
     </section>
@@ -1913,8 +1914,8 @@
         </dt>
         <dd>
           Indicates that the corresponding <a>PaymentItem</a> represents a form
-          of tax (e.g., sales tax, goods and services tax, value added tax,
-          etc.).
+          of taxation. Examples include sales tax, goods and services tax,
+          value added tax, an so on.
         </dd>
       </dl>
     </section>
--- a/index.html
+++ b/index.html
@@ -1849,7 +1849,7 @@
           required DOMString label;
           required PaymentCurrencyAmount amount;
           boolean pending = false;
-          PaymentItemType type;
+          PaymentItemType? type;
         };
       </pre>
       <p>
@@ -1886,10 +1886,11 @@
           <dfn>type</dfn> member
         </dt>
         <dd>
-          A <a>PaymentItemType</a> enum value. A the value of <a>type</a> to
-          assist in the presentation of <a>PaymentItem</a> by, for example,
-          visually grouping types together or other otherwise distinguishing
-          them from other types (or from items that have no associated type).
+          A <a>PaymentItemType</a> enum value or null. A user agent MAY use the
+          value of <a>type</a> to assist in the presentation of
+          <a>PaymentItem</a> by, for example, visually grouping types together
+          or other otherwise distinguishing them from other types (or from
+          items that have no associated type).
         </dd>
       </dl>
     </section>
--- a/index.html
+++ b/index.html
@@ -124,6 +124,11 @@
             <a href="https://github.com/w3c/payment-request/issues/617">issue
             617</a>).
           </li>
+          <li data-link-for="PaymentItem">
+            <a>PaymentItem</a>'s <a>type</a> member and the
+            <a>PaymentItemType</a> enum (see <a href=
+            "https://github.com/w3c/payment-request/issues/163">issue 163</a>).
+          </li>
         </ul>
       </section>
     </section>
@@ -1849,6 +1854,7 @@
           required DOMString label;
           required PaymentCurrencyAmount amount;
           boolean pending = false;
+          // Note: type member is "at risk" of being removed!
           PaymentItemType type;
         };
       </pre>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1885,6 +1887,39 @@ <h2>
           shipping option. <a>User agents</a> MAY indicate pending fields in
           the user interface for the payment request.
         </dd>
+        <dt>
+          <dfn>type</dfn> member
+        </dt>
+        <dd>
+          A <a>PaymentItemType</a> enum value. A the value of <a>type</a> to
```
mnoorenberghe: I think you have a mistake at "A the value of"

mnoorenberghe (commented):
From an implementation perspective, this will nicely solve our use case. 👍

mnoorenberghe (approved):

index.html
```diff
@@ -1885,6 +1887,40 @@ <h2>
...
+        </dd>
+      </dl>
+    </section>
+    <section data-dfn-for="PaymentItemType">
+      <h2>
+        <dfn>PaymentItemType</dfn> enum
+      </h2>
+      <pre class="idl">
+        enum PaymentItemType {
+          "tax"
```
adrianhopebailie: Instead of `other`, could we have `lineItem` as the default and include in the enum?

index.html
```diff
@@ -1885,6 +1887,40 @@ <h2>
...
+        };
+      </pre>
+      <p>
+        The <a>PaymentItemType</a> serves to categorize a <a>PaymentItem</a>
+        into particular types.
```
adrianhopebailie: Should we have some indication of why this feature exists? i.e. Browsers MAY use this to influence their UI

stpeter: +1 - it's always good to say why. :-) I'm not sure if we have user experience data to back it up, but grouping line items into categories (say, national and local taxes) could help users understand what they're being charged.

index.html
marcoscaceres: I'm not sure we need that default at all.

index.html
marcoscaceres: @adrianhopebailie, @stpeter, [above](https://github.com/w3c/payment-request/pull/666/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR1894), I state that the user agent MAY use this to, "for example, visually grouping types together or otherwise distinguishing them from other types (or from items that have no associated type)." 

Is that sufficient?

stpeter: Yes, that seems good enough for now.

marcoscaceres:
PR for tests sent: https://github.com/w3c/web-platform-tests/pull/9099

marcoscaceres:
To make this work with #669, will probably need to make `.type` nullable.

index.html
```diff
@@ -1853,6 +1854,7 @@ <h2>
           required DOMString label;
           required PaymentCurrencyAmount amount;
           boolean pending = false;
+          PaymentItemType? type;
```
domenic: No need for this to be null. Just have them not supply it if they don't want to set it (or, set it to undefined).

index.html
```diff
@@ -1885,6 +1887,40 @@ <h2>
           shipping option. <a>User agents</a> MAY indicate pending fields in
           the user interface for the payment request.
         </dd>
+        <dt>
+          <dfn>type</dfn> member
+        </dt>
+        <dd data-tests="PaymentItem/type_member.https.html">
+          A <a>PaymentItemType</a> enum value or null. A user agent MAY use the
```
domenic: So here I would say "If supplied, a PaymentItemType enum value"

index.html
```diff
@@ -1885,6 +1887,40 @@ <h2>
           shipping option. <a>User agents</a> MAY indicate pending fields in
           the user interface for the payment request.
         </dd>
+        <dt>
+          <dfn>type</dfn> member
```
domenic: Since there is no type for the "normal" payment items, I wonder if this should be named "specialType" or something? It's probably fine as just `type` though.

index.html
```diff
@@ -1885,6 +1887,40 @@ <h2>
...
+        into particular types.
+      </p>
+      <dl>
+        <dt>
+          "<dfn>tax</dfn>"
+        </dt>
+        <dd>
+          Indicates that the corresponding <a>PaymentItem</a> represents a form
+          of tax (e.g., sales tax, goods and services tax, value added tax,
+          etc.).
```
domenic: Would it be useful to say something like "(The developer can use the `label` member to be more specific.)"? I guess in that case I'd move the examples outside of the parentheses so there aren't too many parentheses.

domenic (changes requested):
Nit on IDL style, where we shouldn't use nullable dictionary members generally, instead just allowing them to be omitted.

Happy to leave naming concerns in others' hands, although I did weigh in a bit myself.

domenic:
> @domenic, I'm wondering if there would be any advantage to having a default value for the type member (e.g., "other")? Thought I'd ask just in case.

Since these are never reflected back to the web developer, I don't see any concrete advantage to a default. I guess it would allow people to be explicit and write `type: isTaxable(x) ? "tax" : "other"` instead of `type: isTaxable(x) ? "tax" : undefined`. But even them coming up with the right name seems hard, so I wouldn't bother.

index.html
marcoscaceres: I'd like to just keep it as `type`.

index.html
marcoscaceres: Yeah, but I think the developer bit would probably be better in `type` member above, so will add it there. I'll remove the parens.

domenic (approved):

marcoscaceres:
@zkoch, @adrianba, @aestes, need your ok on this as implementers. 

We are [planning to support this in Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1432079).

zkoch:
@rsolomakhin thoughts? seems good to me and easy to support.

rsolomakhin:
@zkoch: :+1: looks good to me as well.

adrianba:
We previously had a type of display items and decided to remove it from v1 because it was hard to get agreement on exactly how it should work. I don't support rushing this in at this point. It hasn't been a feature we've seen high priority requests for.

marcoscaceres:
We (Firefox) are willing to implement it, so maybe we can add it as a "Feature at Risk" to see if any merchants will use it.

stpeter:
+1 @marcoscaceres LGTM

stpeter (approved):

marcoscaceres:
@ianbjacobs, could you please send a request to republish the spec.

ianbjacobs:
@marcoscaceres, request sent.
Ian

PR #668, opened 2018-01-17
wip: allow show() to take optional promise

### LINKED ISSUE #645: Add way to update `total` and `displayItems` after `.show()` but before user interaction
This is related to but different from #639.

Some users have reported to us that they may not know the total or line items at the time of attempting to open the payment sheet with `show()` -- there may be asynchronous server-based logic yet to be resolved.

Apple Pay provides an unrelated API that users can repurpose to work around this: if you listen for [onpaymentmethodselected](https://developer.apple.com/documentation/applepayjs/applepaysession/1778013-onpaymentmethodselected), you can get a loading state when the sheet opens and then asynchronously update the total and return control to the sheet. We have at least one user that has been relying on this behavior with Apple Pay.

Is it worth adding a way to signal that the browser should show a spinner or equivalent and wait to allow end user interaction until an update? The alternative is for Payment Request implementors to block the end-user action (not great) or show a manually-created loading modal (probably also not great).

marcoscaceres:
Closes #645.

Changes `show()` to take an optional promise that resolves with a `PaymentDetailsUpdate`:

```
Promise<PaymentResponse> show(optional Promise<PaymentDetailsUpdate> detailsPromise);
```

We can then reuse the machinery to perform an update, but splitting it out into its own algorithm

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/668.html" title="Last updated on Jan 22, 2018, 4:47 AM GMT (de92b30)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/668/fe1a3ca...de92b30.html" title="Last updated on Jan 22, 2018, 4:47 AM GMT (de92b30)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -940,11 +940,22 @@
               <li>Set <var>request</var>.<a>[[\updating]]</a> to true.
               </li>
               <li>
-                <a>In parallel</a>, run the <a>update a
+                <a>In parallel</a>, disable the user interface user interface
+                that will allow the user to interact with the
+                <var>handlers</var> and run the <a>update a
                 <code>PaymentRequest</code>'s details algorithm</a> with
                 <var>detailsPromise</var> and <var>request</var>.
               </li>
-              <li>Wait for <var>detailsPromise</var> to settle.
+              <li>Wait for the <var>detailsPromise</var> to settle.
+                <p class="note">
+                  Based on how the <var>detailsPromise</var> settles, the
+                  <a>update a <code>PaymentRequest</code>'s details
+                  algorithm</a> determines how the payment UI behaves. That is,
+                  <a>upon rejection</a> of the <var>detailsPromise</var>, the
+                  payment request aborts. Otherwise, <a>upon fulfillment</a>
+                  <var>detailsPromise</var>, the user agent re-enables the
+                  payment request UI and the payment flow can continue.
+                </p>
               </li>
             </ol>
           </li>
--- a/index.html
+++ b/index.html
@@ -818,7 +818,8 @@
           </p>
         </div>
         <p data-tests="payment-request-show-method.https.html">
-          The <a>show()</a> method MUST act as follows:
+          The <code><a data-lt="show()">show(detailsPromise)</a></code> method
+          MUST act as follows:
         </p>
         <ol class="algorithm">
           <li>Let <var>request</var> be the <a>PaymentRequest</a> object on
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2701,317 +2717,11 @@ <h2>
...
-            <li>Abort the current user interaction and close down any remaining
-            user interface.
-            </li>
-            <li>Set <var>request</var>.<a>[[\state]]</a> to "<a>closed</a>".
-            </li>
-            <li>Reject the promise <var>request</var>.<a>[[\acceptPromise]]</a>
-            with <var>exception</var>.
-            </li>
-            <li>Abort the algorithm.
+            <li>Return from the method and, <a>in parallel</a>, run the
```
marcoscaceres: The stuff deleted here is just moved and given its own definition below ("update a <code>PaymentRequest</code>'s details algorithm").

index.html
```diff
@@ -932,11 +932,38 @@ <h2>
           <a>DOMException</a>, and set the <a>user agent</a>'s <a>payment
           request is showing</a> boolean to false.
           </li>
+          <li>Otherwise, present a user interface that will allow the user to
```
domenic, line 936: The intro text which says "The show() method must act as follows" should be updated to say "the show(detailsPromise) method". Cf. updateWith()'s definition elsewhere.

index.html
```diff
@@ -932,11 +932,38 @@ <h2>
...
+          interact with the <var>handlers</var>.
+          </li>
+          <li>If <var>detailsPromise</var> was passed, then:
+            <ol>
+              <li>Set <var>request</var>.<a>[[\updating]]</a> to true.
+              </li>
+              <li>
+                <a>In parallel</a>, disable the user interface user interface
```
domenic: This should not be done in parallel, as you can't update the UI from off the main thread (i.e. in parallel with the main thread). I guess it's my bad for implying the updating should be done in parallel; it shouldn't. (See later comment.)

index.html
```diff
@@ -932,11 +932,38 @@ <h2>
...
+                </p>
+              </li>
+            </ol>
+          </li>
           <li>
             <p>
-              Otherwise, present a user interface to allow the user to interact
-              with the <var>handlers</var>. The user agent SHOULD prioritize
-              the preference of the user when presenting payment methods.
+              Present <var>handlers</var> to the user. The user agent SHOULD
```
domenic: I'm confused. Above you displayed a user interface that allowed the user to interact with _handlers_. Now you are presenting _handlers_ to the user. What is the difference between these two? Why do both?

I think what I would do is add a step in the "If _detailsPromise_ was passed, then:" saying "Re-enable the user interface that allows the users to interact with _handlers_." And, I would move all the paragraphs in this last step up to the first "present a user interface" step. Am I on the right track?

index.html
```diff
@@ -2701,317 +2728,11 @@ <h2>
...
-            <li>Abort the current user interaction and close down any remaining
-            user interface.
-            </li>
-            <li>Set <var>request</var>.<a>[[\state]]</a> to "<a>closed</a>".
-            </li>
-            <li>Reject the promise <var>request</var>.<a>[[\acceptPromise]]</a>
-            with <var>exception</var>.
-            </li>
-            <li>Abort the algorithm.
+            <li>Return from the method and, <a>in parallel</a>, run the
```
domenic: This shouldn't be done in parallel, it turns out; the algorithm just waits on promises, which should be done on the main thread (like it is done in JS).

domenic (changes requested):
A few issues but overall seems to work!

index.html
marcoscaceres: You are correct, I screwed this up.

index.html
marcoscaceres: Updated to match method signature.

marcoscaceres:
@domenic, hopefully I understood correctly and you wanted me to drop the "in parallels", right? You are correct that the state machine should handle the signaling to the UI thread while allowing all this to run on the main thread (and just waiting for the `detailsPromise` to settle). 

Giving it another read... to see if it makes sense.

marcoscaceres:
Going to split this into two PR. One editorial one, for the update algo, and then the next for the API change.

Issue #649, opened 2017-11-07
updateWith(data) should merge with request.data


marcoscaceres:
Given partial PaymentDetailsUpdate, should we object merge with whatever the sheet is showing? 

## Problem
Given: 

```JS
const details = {...}; // total, shippingOptions,  displayItems, etc. 
const request = new PaymentRequest(methods, details, options);
// later
request.addEventListener("shippingaddresschange", ev =>{
   ev.updateWith(new Promise(async resolve => {
       const total =  await calculateNewShippingCost(request.shippingAddress); 
       resolve({total}); // Partial update, just the total
   });
});
```

So, `details` initially populates the payment sheet with a `total`, `displayItems`, `shippingOptions` etc. 

Today, `ev.updateWith({total})` being called causes `displayItems`, `shippingOptions`, etc. to all get trashed - and only the total updates. Obviously, that's not great. 

## Proposal 
The proposal is that when `updateWith()` is called, we do a property spread over the original object, so old properties values are retained, and new properties are added (or nulled/emptied out).

Basically, this would be the browser code:

````JS
// `newData` what is sent in when via ev.updateWith() resolves 
sheet.addEventListener("updatepaymentsheet", ({newData}) => {
   // merge the existing data with the new data, so creates a partial update  
   const updatedData = {...sheet.getCurrentData(), ...newData};
  sheet.render(updatedData);
});
````      

cc @aestes, @rsolomakhin, @michelle

ianbjacobs:
Can you describe a use case to help explain the issue? (I think I am missing context.)

marcoscaceres:
@ianbjacobs, described the problem more fully above. This change is backwards compatible, because right now developers have to send in everything. 

However, we would need everyone to agree to make this change in their implementations... or it's gonna be much sadness.

domenic:
So, I don't think this is actually a problem with the algorithm as written. For example, if you don't pass `displayItems`, what happens is that "If the displayItems member of details is present" is false in step 5 and in step 6, so _request_.[[details]].displayItems is never touched. The same goes for all the others.

Am I misreading?

marcoscaceres:
Hmm, maybe it’s just incorrectly implemented in Chrome (and possibly Edge). Right now, Chrome definitely requires a full ‘details’ object or things vanish.

domenic:
Sounds like a job for web platform tests ^_^

marcoscaceres:
[...]

PR #669, opened 2018-01-19
Spread resolved `data` on updateWith()

marcoscaceres:
This builds on https://github.com/w3c/payment-request/pull/668 to allow incremental data updates as proposed in #649.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/669.html" title="Last updated on Jan 19, 2018, 5:19 AM GMT (fc5ea16)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/669/fe1a3ca...fc5ea16.html" title="Last updated on Jan 19, 2018, 5:19 AM GMT (fc5ea16)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -940,11 +940,22 @@
               <li>Set <var>request</var>.<a>[[\updating]]</a> to true.
               </li>
               <li>
-                <a>In parallel</a>, run the <a>update a
+                <a>In parallel</a>, disable the user interface user interface
+                that will allow the user to interact with the
+                <var>handlers</var> and run the <a>update a
                 <code>PaymentRequest</code>'s details algorithm</a> with
                 <var>detailsPromise</var> and <var>request</var>.
               </li>
-              <li>Wait for <var>detailsPromise</var> to settle.
+              <li>Wait for the <var>detailsPromise</var> to settle.
+                <p class="note">
+                  Based on how the <var>detailsPromise</var> settles, the
+                  <a>update a <code>PaymentRequest</code>'s details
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@domenic, the interesting stuff only happens in https://github.com/w3c/payment-request/pull/669/commits/fc5ea166c04eec1f642fcdc2b60c3070973065ab. The rest is just #668.

domenic (changes requested):
So it's better to stay at the Infra level than to drop down to the ECMAScript level. Otherwise what happens here will be different if someone monkeys with Object.prototype, for example. You'll need to define some kind of ordered map merging operation, or let us do that for you in Infra.

BUT, as I pointed out in https://github.com/w3c/payment-request/issues/649#issuecomment-359108225, I don't think this patch is necessary at all. Let's continue discussing there.

marcoscaceres:
Confirmed, not needed. Sorry about the noise. Someone needs to fire the person writing test tho :)

marcoscaceres:
@domenic, I'm still kinda wondering if we should add "property spread" to infra (even if we didn't need it here). 

Consider, after validation the `values` object, it would be nice to just spread them over `request.[[details]]` instead of adding/replacing the properties one by one.

domenic:
Yeah, I definitely think some kind of "merge-with-overwrite" for ordered maps would be reasonable. I'd prefer to wait until we have a consumer, but we could open a tracking issue perhaps.

Issue #629, opened 2017-09-21
"Exposing available payment methods" is confusing given the existence of canMakePayment()


domenic:
A Chrome developer ended up confused by this section because it says

> The fact that a successful match to a payment method causes a user interface to be displayed mitigates the disclosure risk.

However this is not true for "the payment request API" in general, only for paymentRequest.show(). In particular canMakePayment() can be called without UI.

This section should be rewritten to be specific what methods it's talking about, and talk about canMakePayment()'s step 3 mitigations additionally.

---

As a separate problem, the "may" requirements in this section are very bad, and should be moved to the show() method.

---

I can try to work on this "soon", but it's hard to guarantee availability for this week or next, so since it seems things are heading toward some sort of spec freeze, maybe someone else can help out here.

marcoscaceres:
I can have a go at fixing it.

lknik:
I support this request :)

marcoscaceres:
@lknik, be careful, or we will rope you in to help and review stuff.

ianbjacobs:
From a republication of CR perspective,this sounds editorial to me.

marcoscaceres:
> As a separate problem, the "may" requirements in this section are very bad, and should be moved to the show() method.

This will be a MUST https://github.com/w3c/payment-request/pull/655

Sending a PR that removes this section entirely, because it's bad (also as per #475).

marcoscaceres:
Sent PR for this https://github.com/w3c/payment-request/pull/673

domenic:
Fixed by https://github.com/w3c/payment-request/pull/673

PR #655, opened 2017-11-24
show() must be triggered by user activation

marcoscaceres:
closes #651

The following tasks have been completed:

 * [X] [web platform tests](https://github.com/w3c/web-platform-tests/pull/9990)
 * [x] [MDN Docs added](https://developer.mozilla.org/en-US/docs/Web/API/PaymentRequest/show) 

Implementation commitment:

 * [x] Safari (already in WebKit)
 * [x] [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=817807)
 * [x] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1442078)
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/655.html" title="Last updated on Mar 14, 2018, 12:43 AM GMT (a9cc8ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/655/463d25c...a9cc8ba.html" title="Last updated on Mar 14, 2018, 12:43 AM GMT (a9cc8ba)">Diff</a>

Issue #641, opened 2017-10-09
Extract actionable items from @lknik's review


marcoscaceres:
There are a few actionable items from @lknik's review that we should fix: 
https://blog.lukaszolejnik.com/privacy-of-web-request-api/

We should review the article and extract actionable things that we can fix (specially around iframes, which @lknik shows are pretty easy to bypass).

I believe this is related: 
https://github.com/w3c/payment-request/issues/629

ianbjacobs:
Thank you @lknik and @marcoscaceres. One idea occurred to me upon reading the piece:
extend rate limiting to cover both the "parent origin" and any parent/child (included script)
pairs. I don't know if that's what implementations do today. 

A script could be included by multiple parents which is why I thought of rate limiting the pairs (as part of rate limiting the parent).

I've added this topic to our FTF agenda.

Ian

lknik:
Yeah, thought of a similar solution (quota on effective top-level context/domain). Then again, what about CORS or other hacks?

marcoscaceres:
closing, as priv/sec review are complete.

PR #673, opened 2018-01-23
Editorial fixes

### LINKED ISSUE #440: Authors are not conformance classes
The spec says:

``` HTML
Authors SHOULD NOT set <code>selected</code> to true on more than one item. 
```

This needs to be converted to a note. 

See also #437.

### LINKED ISSUE #475: Clarity required in section 21.2 regarding privacy
Section 21.2 states "There may be legitimate scenarios for calling repeatedly (for example, to control the order of payment method selection)".

Does this imply that this should be supported?

The following wording "The fact that a successful match to a payment method causes a user interface to be displayed mitigates the disclosure risk" does not allow for the use of CanMakePayments() which has no such mitigation.

I recommend this wording is reviewed.

marcoscaceres:
* closes #440
* closes #475, #629

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/673.html" title="Last updated on Jan 24, 2018, 1:52 AM GMT (2283a7d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/673/3ad5aeb...2283a7d.html" title="Last updated on Jan 24, 2018, 1:52 AM GMT (2283a7d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -840,8 +840,7 @@
               rejected with a "<a>SecurityError</a>" <a>DOMException</a>. For
               example, the <a>user agent</a> may require the call to be
               <a>triggered by user activation</a>, or may limit the rate at
-              which a page can call <a>show()</a>, as described in the <a href=
-              "#privacy">privacy considerations</a> section.
+              which a page can call <a>show()</a>.
             </p>
             <p class="note">
               During the Candidate Recommendation phase, implementations are
@@ -3350,22 +3349,6 @@
           "PaymentDetailsModifier.additionalDisplayItems">additionalDisplayItems</a>
           member with a third-party <a>payment handler</a> without user
           consent.
-        </p>
-      </section>
-      <section>
-        <h2>
-          Exposing available payment methods
-        </h2>
-        <p>
-          Developers might try to call the payment request API repeatedly with
-          only one payment method identifier to try to determine what payment
-          methods a <a>user agent</a> has installed. There are legitimate
-          scenarios for calling repeatedly (for example, to control the flow of
-          payment method selection). The fact that a successful match to a
-          payment method causes a user interface to be displayed mitigates the
-          disclosure risk. Implementations MAY require a user action to
-          initiate a payment request or they MAY rate limit the calls to the
-          API to prevent too many repeated calls.
         </p>
       </section>
     </section>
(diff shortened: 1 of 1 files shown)


ianbjacobs (commented):
Hi @marcoscaceres,

I think this is ok given that other parts of the spec address these points. However, note the text earlier in the document:

 " For example, the user agent may require the call to be triggered by user activation, or may limit the rate at which a page can call show(), as described in the privacy considerations section."

The "as described..." needs to be deleted.

(I looked for other references to the section but only see the one.)

Ian

marcoscaceres:
@ianbjacobs, thanks. Fixed.

marcoscaceres:
Noting that this is also addressed in https://github.com/w3c/payment-request/pull/655 

... should get that merged... but it means we are going to break 90% of the tests, so it's work and so lazy 😴

index.html
```diff
@@ -3354,22 +3351,6 @@ <h2>
           consent.
         </p>
       </section>
-      <section>
-        <h2>
-          Exposing available payment methods
-        </h2>
-        <p>
-          Developers might try to call the payment request API repeatedly with
```
domenic, line 3362: Personally I find consolidating privacy considerations into a single section to be a good idea. The difference would be that you'd have a pointer to the parts of the algorithms that help preserve privacy, instead of adding MAY stuff there. So I would change this to include all the same helpful overview of the problem, then point to the mitigations we've put in place in show() and canMakePayment().

I guess that's what was already discussed in https://github.com/w3c/payment-request/issues/475#issuecomment-320657383, and I didn't say anything then, so let me sign off on this since it's not a big deal either way and I don't want to relitigate something decided last August.

index.html
```diff
@@ -1530,9 +1529,7 @@ <h2>
               <a data-lt="PaymentRequest.shippingOption">shippingOption</a>
               will be set to the <a data-lt="PaymentShippingOption.id">id</a>
               of this option without running the <a>shipping option changed
-              algorithm</a>. Authors SHOULD NOT set <a data-lt=
```
domenic, line 1533: As noted in https://github.com/w3c/payment-request/issues/440 I don't understand this removal; it seems to be removing good information.

domenic (approved):
LGTM to proceed although I guess I have some quibbles if you want to discuss.

marcoscaceres:
I'd prefer we move forward and rewrite the Priv/Security section properly once we land #655, and we can reflect more fully on what we've learned from the CR process. Additionally, we need to document the mitigations put in place as a result of #641.

marcoscaceres:
Will pick this up again in https://github.com/w3c/payment-request/issues/675

PR #682, opened 2018-02-12
Editorial: align show() and canMakePayment()

ianbjacobs:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/682.html" title="Last updated on Apr 6, 2018, 6:14 AM GMT (dd951a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/682/678a110...dd951a5.html" title="Last updated on Apr 6, 2018, 6:14 AM GMT (dd951a5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3371,9 +3371,7 @@
           For security reasons, a user agent can limit matching (in
           <a>show()</a> and <a>canMakePayment()</a>) to <a>payment handlers</a>
           from the same <a data-cite="rfc6454#section-3.2">origin</a> as a URL
-          <a>payment method identifier</a>. User agents can also use
-          information provided by a <a>payment method</a> owner to match
-          <a>payment handlers</a> from other origins.
+          <a>payment method identifier</a>.
         </p>
       </section>
     </section>
--- a/index.html
+++ b/index.html
@@ -3354,14 +3354,18 @@
           <a>payment methods</a> support this.
         </p>
       </section>
-      <section>
+      <section class="informative">
         <h2>
-          Payment Handler Matching
+          How user agents match payment handlers
         </h2>
         <p>
-          A user agent's payment handler matching algorithm takes into account
-          the requirements of this specification as well as matching defined in
-          payment methods and other specifications.
+          As part of <a>show()</a>, the user agent typically displays a list of
+          matching <a>payment handlers</a> that satisfy the <a>payment
+          methods</a> accepted by the merchant and other conditions. Matching
+          can take into account information provided by the merchant,
+          information provided by the <a>payment method</a> owner, the
+          <a>payment handlers</a> registered by the user, user preferences, and
+          other considerations.
         </p>
         <p data-link-for="PaymentRequest">
           For security reasons, a user agent can limit matching (in
--- a/index.html
+++ b/index.html
@@ -1078,14 +1078,32 @@
           <li>Return <var>promise</var>, and perform the remaining steps <a>in
           parallel</a>.
           </li>
-          <li>For each <var>methodData</var> in
+          <li>For each <var>paymentMethod</var> tuple in
           <var>request</var>.<a>[[\serializedMethodData]]</a>:
             <ol>
-              <li>If <var>methodData</var>.<a data-lt=
-              "PaymentMethodData.supportedMethods">supportedMethods</a> is a
-              supported <a>payment method identifier</a> (including its
-              <a>payment method</a> specific capabilities), resolve
-              <var>promise</var> with true, and abort this algorithm.
+              <li>Let <var>identifier</var> be the first element in the
+              <var>paymentMethod</var> tuple.
+              </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3340,6 +3358,11 @@ <h2>
         <h2>
           Payment Handler Matching
         </h2>
+        <p>
+          A user agent's payment handler matching algorithm takes into account
```
marcoscaceres: I feel like "the requirements of this specification" needs to link to something - as the specification has a lot of requirements and it's not clear exactly which we are talking about here. 

Also, there is no defined "payment handler matching algorithm", so maybe this text should not be here at all.

marcoscaceres (changes requested):
The algorithm seems fine, but the "payment handler matching" still seems super hand-wavy. I'd be more inclined to remove that text entirely.

index.html
ianbjacobs: @marcoscaceres,

Here is some additional detail. If you want to understand why some apps show up in the list and some don't, you need to be aware of:
 * Payment method identifier matching defined in PR API.
 * canMakePayment() matching, which is payment method specific 
 * payment method manifest matching
 * potentially other matching to come (e.g., if we add a layer for 3DS flows where the merchant can say "I only accept payments from apps that do 3DS flows given the jurisdiction in which I operate.")

I'd like to raise awareness of this. I welcome your suggestions of how and where that might be done.

Ian

marcoscaceres: We can do it here, but as a note (i.e., class = "informative" and in `<div class=note>`)... like:

### How user agents match payment handlers
A merchant can accept payment by supporting many different payment handlers (e.g, "basic-card" and "`https://example.com/pay`"). As such, when a merchant initiates a request for payment, a user agent will typically display a "payment sheet" and give the user the option to select from the matching payments handlers available. The matching is done on: 
  
  * The payment method identifier: e.g., the user agent supports "basic-card", 
     but not "https://example.com/pay". So "basic-card" would be the only 
     option available to the user. 

  * User preference - a user might say they exclusively want to use "https://example.com/pay", and not be bothered with "basic-card". So the use completes the payment with the "https://example.com/pay" payment method. 

And so on... 

I think it would be overstepping to mention payment method manifest or 3D at this point (obviously). But if those two could be generalized, then we should include them above maybe... depends if it would imply support for something none supports yet.

index.html
```diff
@@ -3354,22 +3354,24 @@ <h2>
...
+      <section class="informative">
         <h2>
-          Payment Handler Matching
+          How user agents match payment handlers
         </h2>
         <p>
-          A user agent's payment handler matching algorithm takes into account
-          the requirements of this specification as well as matching defined in
-          payment methods and other specifications.
+          As part of <a>show()</a>, the user agent typically displays a list of
```
marcoscaceres: nit: add `data-link-for="PaymentRequest"` to the `p` :)

index.html
```diff
@@ -3354,22 +3354,24 @@ <h2>
...
+          matching <a>payment handlers</a> that satisfy the <a>payment
```
marcoscaceres: It's still quite unclear what "matching" is - but maybe 🤷🏽‍♂️

ianbjacobs:
Hi @marcoscaceres,
I have made some edits to the text based on your comment.

Chrome supports payment method manifest, so I think it is appropriate to mention the concept of matching affected by information coming from the payment method owner. Whether we choose to name the Payment Method Manifest spec (a FPWD) is a separate question. (I think we should, FWIW.)

marcoscaceres:
> Chrome supports payment method manifest, so I think it is appropriate to mention the concept of matching affected by information coming from the payment method owner. 

You know I love Chrome, but I think it's unfair to mention it, because we don't know if any other browser vendors is going to support it. Mozilla currently has no plans to support it, and I don't think anyone else has said anything about supporting it (though it may not be possible to support payment handlers without it... but I've no idea, I've not looked that much into it).

ianbjacobs:
@marcoscaceres wrote 

 "Mozilla currently has no plans to support it, and I don't think anyone else has said anything about supporting it (though it may not be possible to support payment handlers without it... but I've no idea, I've not looked that much into it)."

I think you and I should discuss it the manifest spec since I think it's an important complement to Payment Handler API.

marcoscaceres:
> I think you and I should discuss it the manifest spec since I think it's an important complement to Payment Handler API.

Always happy to chat, but we are not quite ready to discuss supporting more stuff until we have a solid implementation of Payment Request and Basic Card. Too much work, too little time :(

ianbjacobs:
@marcoscaceres,

I hear the prioritization and understand the resource crunch. 

Nonetheless, I think it is ok to mention this in an informative note at this time (where we have FPWDs, like Payment Handler and Payment Method Manifest)

Ian

marcoscaceres (approved):

PR #690, opened 2018-03-01
Add regionCode attribute

### LINKED ISSUE #65: ContactAddress 'region' inconsistencies
I'm having a number of issues integrating our checkout with the payment request API when it comes to shipping options / shipping address forms. These centre around the address 'region', e.g. 'CA' in the US. 

**Project info / requirement**
To give some background, I have a requirement to have shipping options that are available to certain regions within a country. This could be for legal reasons, e.g. in the US, only being able to ship certain goods to certain states. I must therefore, verify user inputted address region values against our shipping options when determining which shipping options to pass to the payment request interface.

**Lack of 'region' input**
Depending on the country however, sometimes the region is mandatory (US, Australia), optional (Ireland), missing (Qatar). Can this field be mandatory for all countries? (i.e. we would need 'region' to be mandatory in all cases so we can verify our shipping options against the user's country and state)

**Inconsistency of input type and value**
For both address forms that could be used, i.e. billing and shipping, the value retrieved from the user is not in a consistent format, in terms of input type and response value. On web browsers (as far as I've tested), the region is a dropdown. However, depending on the country, the value returned is not consistent, e.g. it's a 2 letter code such as 'CA' for the US, state name such as 'Delhi' for India, and 'Co. Dublin' for Ireland. 

On mobile browsers (and possibly some web, but generally mobile), the 'region' is a free form input, so the customer could enter anything essentially, e.g. 'CA', 'California', 'Californnn'. In this case, we have no direct solution for mapping this value.

**Overall solution proposal**
- 'Region' is mandatory.
- 'Region' returns the state as a 2 letter code (ISO 3166-2), from a pre-defined set of values.

### LINKED ISSUE #663: ContactAddress 'region' inconsistencies
I'm having a number of issues integrating our checkout with the payment request API when it comes to shipping options / shipping address forms. These centre around the address 'region', e.g. 'CA' in the US. 

**Project info / requirement**
To give some background, I have a requirement to have shipping options that are available to certain regions within a country. This could be for legal reasons, e.g. in the US, only being able to ship certain goods to certain states. I must therefore, verify user inputted address region values against our shipping options when determining which shipping options to pass to the payment request interface.

**Lack of 'region' input**
Depending on the country however, sometimes the region is mandatory (US, Australia), optional (Ireland), missing (Qatar). Can this field be mandatory for all countries? (i.e. we would need 'region' to be mandatory in all cases so we can verify our shipping options against the user's country and state)

**Inconsistency of input type and value**
For both address forms that could be used, i.e. billing and shipping, the value retrieved from the user is not in a consistent format, in terms of input type and response value. On web browsers (as far as I've tested), the region is a dropdown. However, depending on the country, the value returned is not consistent, e.g. it's a 2 letter code such as 'CA' for the US, state name such as 'Delhi' for India, and 'Co. Dublin' for Ireland. 

On mobile browsers (and possibly some web, but generally mobile), the 'region' is a free form input, so the customer could enter anything essentially, e.g. 'CA', 'California', 'Californnn'. In this case, we have no direct solution for mapping this value.

**Overall solution proposal**
- 'Region' is mandatory.
- 'Region' returns the state as a 2 letter code (ISO 3166-2), from a pre-defined set of values.

marcoscaceres:
closes w3c/contact-picker#65 

 * [x] [web platform tests](https://github.com/w3c/web-platform-tests/pull/11021)
 * [x] [MDN Docs added ](https://developer.mozilla.org/en-US/docs/Web/API/PaymentAddress/regionCode) 

Implementation commitment:

 * [ ] Safari (link to issue)
 * [x] [Chrome](https://crbug.com/817960) 
 * [x] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1441752)
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/690.html" title="Last updated on May 16, 2018, 6:20 AM GMT (699d329)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/690/b2c32fe...699d329.html" title="Last updated on May 16, 2018, 6:20 AM GMT (699d329)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2204,21 +2204,24 @@
             the empty string:
               <ol>
                 <li>Let <var>regionCode</var> be the result of <a>strip leading
-                and trailing ASCII whitespace</a> and <a data-cite=
-                "!INFRA#ascii-uppercase">ASCII uppercasing</a>
-                <var>details</var>["<a>regionCode</a>"].
+                and trailing ASCII whitespace</a> from
+                <var>details</var>["<a>regionCode</a>"] and then
+                  <a data-cite="!INFRA#ascii-uppercase">ASCII uppercasing</a>
+                  the result.
                 </li>
                 <li>
                   <p>
-                    If <var>regionCode</var> is not a valid subdivision code as
-                    per [[!ISO3166-2]]'s section 5.2 Structure of country
-                    subdivision code elements (details below), throw a
+                    If <var>regionCode</var> is not a valid <a>country
+                    subdivision code element</a> as per [[!ISO3166-2]]'s
+                    section 5.2 "Structure of country subdivision code
+                    elements" (non-normative details below), throw a
                     <a>RangeError</a> exception.
                   </p>
                   <div class="note" title=
                   "Structure of country subdivision code elements">
                     <p>
-                      The structure of country subdivision code elements is
+                      <strong>Do not implement from this note.</strong> The
+                      structure of a <a>country subdivision code element</a> is
                       formally defined in [[!ISO3166-2]] (section 5.2).
                       Although the structure is not expected to change at the
                       time of writing, implementers are expected to track
@@ -2226,8 +2229,8 @@
                     </p>
                     <p>
                       As [[!ISO3166-2]] is not freely available to the general
-                      public, the structure of a country subdivision code
-                      elements at the time of publication is as follow:
+                      public, the structure of a <a>country subdivision code
+                      element</a> at the time of publication is as follows:
                     </p>
                     <ul>
                       <li>Two <a>code points</a> that match an [[!ISO3166-1]]
@@ -2236,15 +2239,33 @@
                       <li>A single U+002D (-) <a>code point</a>.
                       </li>
                       <li>One, two, or three <a data-cite=
-                      "infra#ascii-alphanumeric">ASCII alphanumeric</a> code
-                      points, in any order. This is referred to as the
-                      subdivision name.
+                      "INFRA#ascii-alphanumeric">ASCII alphanumeric</a> code
+                      points, in any order.
                       </li>
                     </ul>
                   </div>
                 </li>
                 <li>Set <var>address</var>.<a>[[\regionCode]]</a> to
                 <var>regionCode</var>.
+                </li>
+                <li>Let <var>region</var> be the corresponding [[!ISO3166-2]]
+                <a>country subdivision name</a> for <var>regionCode</var>.
+                Where [[!ISO3166-2]] defines multiple <a>country subdivision
+                names</a> for a <var>regionCode</var>, it is RECOMMENDED the
+                user agent select one by matching on:
+                  <ol>
+                    <li>The <a data-cite="!HTML#language">language</a> of
+                    <a data-cite="HTML#the-body-element-2">the body
+                    element</a>.
+                    </li>
+                    <li>The user's preferred languages.
+                    </li>
+                    <li>Any other criteria the user agent deems suitable.
+                    </li>
+                  </ol>
+                </li>
+                <li>Set <var>address</var>.<a>[[\region]]</a> to
+                <var>region</var>.
                 </li>
               </ol>
             </li>
@@ -2371,9 +2392,9 @@
           </h2>
           <p data-link-for="">
             Represents the <a>region</a> of the address as an [[!ISO3166-2]]
-            country subdivision code element. When getting, returns the value
-            of the <a>PaymentAddress</a>'s <a>[[\regionCode]]</a> internal
-            slot.
+            <a>country subdivision code element</a>. When getting, returns the
+            value of the <a>PaymentAddress</a>'s <a>[[\regionCode]]</a>
+            internal slot.
           </p>
         </section>
         <section>
@@ -2494,9 +2515,9 @@
                 <dfn>[[\region]]</dfn>
               </td>
               <td>
-                A <a>region</a> as a country subdivision name or the empty
-                string, such as "Victoria", representing the state of Victoria
-                in Australia.
+                A <a>region</a> as a <a>country subdivision name</a> or the
+                empty string, such as "Victoria", representing the state of
+                Victoria in Australia.
               </td>
             </tr>
             <tr>
@@ -2504,9 +2525,9 @@
                 <dfn>[[\regionCode]]</dfn>
               </td>
               <td>
-                A <a>region</a> represented as a [[!ISO3166-2]] country
-                subdivision code element or the empty string, stored in its
-                canonical uppercase form. For example, "<code>PT-11</code>"
+                A <a>region</a> represented as a [[!ISO3166-2]] <a>country
+                subdivision code element</a> stored in its canonical uppercase
+                form, or the empty string. For example, "<code>PT-11</code>"
                 represents the Lisbon district of Portugal.
               </td>
             </tr>
@@ -2626,8 +2647,8 @@
             <dfn>regionCode</dfn> member
           </dt>
           <dd>
-            An <a>region</a>, represented as an [[!ISO3166-2]] country
-            subdivision code element.
+            An <a>region</a>, represented as a <a>country subdivision code
+            element</a>.
           </dd>
           <dt>
             <dfn>city</dfn> member
@@ -2769,8 +2790,9 @@
               region, or to the empty string if none was provided.
               </li>
               <li>If <var>details</var>["<a>region</a>"] has as a corresponding
-              [[!ISO3166-2]] country subdivision code element, set
-              <var>details</var>["<a>regionCode</a>"] to that subdivision code.
+              <a>country subdivision code element</a>, set
+              <var>details</var>["<a>regionCode</a>"] to that <a>country
+              subdivision code element</a>.
               </li>
             </ol>
           </li>
@@ -4009,6 +4031,14 @@
         This specification relies on several other underlying specifications.
       </p>
       <dl data-sort="">
+        <dt>
+          ISO 3366-2
+        </dt>
+        <dd>
+          <dfn data-lt="country subdivision names">Country subdivision
+          name</dfn> and <dfn>country subdivision code element</dfn> are
+          defined in [[!ISO3166-2]].
+        </dd>
         <dt>
           Infra
         </dt>
--- a/index.html
+++ b/index.html
@@ -2400,9 +2400,9 @@
             <dfn>regionCode</dfn> attribute
           </h2>
           <p data-link-for="">
-            Represents the <a>region</a> of the address, but as the code
-            element of an [[!ISO3166-2]] country subdivision name (e.g., "CA"
-            for California, USA). When getting, returns the value of the
+            Represents the <a>region</a> of the address as a code element of an
+            [[!ISO3166-2]] country subdivision name (e.g., "CA" for California,
+            USA). When getting, returns the value of the
             <a>PaymentAddress</a>'s <a>[[\regionCode]]</a> internal slot.
           </p>
         </section>
--- a/index.html
+++ b/index.html
@@ -2185,7 +2185,8 @@
             <li>If <var>details</var> was not passed, return
             <var>address</var>.
             </li>
-            <li>If <var>details</var>["<a>country</a>"] is present:
+            <li>If <var>details</var>["<a>country</a>"] is present and not the
+            empty string:
               <ol>
                 <li>Set <var>country</var> the result of <a>strip leading and
                 trailing ASCII whitespace</a> from
@@ -2203,11 +2204,15 @@
             <li>If <var>details</var>["<a>regionCode</a>"] is present and not
             the empty string:
               <ol>
-                <li>Let <var>regionCode</var> be the result of <a>strip leading
+                <li>Let <var>regionPart</var> be the result of <a>strip leading
                 and trailing ASCII whitespace</a> from
                 <var>details</var>["<a>regionCode</a>"] and then
                   <a data-cite="!INFRA#ascii-uppercase">ASCII uppercasing</a>
                   the result.
+                </li>
+                <li>Let <var>regionCode</var> be the concatenation of
+                <var>address</var>.<a>[[\country]]</a>, a single U+002D (-) <a>
+                  code point</a>, <var>regionPart</var>.
                 </li>
                 <li>
                   <p>
@@ -2246,26 +2251,30 @@
                   </div>
                 </li>
                 <li>Set <var>address</var>.<a>[[\regionCode]]</a> to
-                <var>regionCode</var>.
+                <var>regionPart</var>.
                 </li>
-                <li>Let <var>region</var> be the corresponding <a>country
-                subdivision name</a> for <var>regionCode</var>. Where
-                [[!ISO3166-2]] defines multiple <a>country subdivision
-                names</a> for a <var>regionCode</var>, it is RECOMMENDED the
-                user agent select one by matching on:
+                <li>If <var>details</var>["<a>region</a>"] is not present:
                   <ol>
-                    <li>The <a data-cite="!HTML#language">language</a> of
-                    <a data-cite="HTML#the-body-element-2">the body
-                    element</a>.
+                    <li>Let <var>region</var> be the corresponding <a>country
+                    subdivision name</a> for <var>regionCode</var>. Where
+                    [[!ISO3166-2]] defines multiple <a>country subdivision
+                    names</a> for a <var>regionCode</var>, it is RECOMMENDED
+                    the user agent select one by matching on:
+                      <ol>
+                        <li>The <a data-cite="!HTML#language">language</a> of
+                        <a data-cite="HTML#the-body-element-2">the body
+                        element</a>.
+                        </li>
+                        <li>The user's preferred languages.
+                        </li>
+                        <li>Any other criteria the user agent deems suitable.
+                        </li>
+                      </ol>
                     </li>
-                    <li>The user's preferred languages.
-                    </li>
-                    <li>Any other criteria the user agent deems suitable.
+                    <li>
+                      <var>details</var>["<a>region</a>"] to <var>region</var>.
                     </li>
                   </ol>
-                </li>
-                <li>Set <var>address</var>.<a>[[\region]]</a> to
-                <var>region</var>.
                 </li>
               </ol>
             </li>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Added test and indicated commitment to support this in Gecko. 

Anyone else going to support this (`PaymentAddress.regionCode`): @aestes, @adrianba, @zkoch, @adrianba? If so, please give a tick above. 

If "maybe later", also useful, in case this needs to be also marked at risk.

marcoscaceres:
Added [MDN Page](https://developer.mozilla.org/en-US/docs/Web/API/PaymentAddress/regionCode) also now part of the [PaymentAddress](https://developer.mozilla.org/en-US/docs/Web/API/PaymentAddress) definition.

rsolomakhin:
@sebsg: I saw you working on bringing ISO region codes to libaddressinput and Chrome. Do you have a tracking bug for exposing it in PaymentRequest that we can link here?

sebsg:
I created https://bugs.chromium.org/p/chromium/issues/detail?id=817960 to track the work

marcoscaceres:
@rsolomakhin, when you can, could you please let me know the bug number for Chrome for adding the attribute?

rsolomakhin:
@marcoscaceres: We will use https://crbug.com/817960 for adding the attribute as well.

index.html
```diff
@@ -2138,6 +2139,22 @@ <h2>
...
+            <li>If <var>details</var>["<a>regionCode</a>"] is present and not
+            the empty string:
+              <ol>
+                <li>Let <var>regionCode</var> be the result of <a>strip leading
+                and trailing ASCII whitespace</a> and <a data-cite=
+                "!INFRA#ascii-uppercase">ASCII uppercasing</a>
+                <var>details</var>["<a>regionCode</a>"].
+                </li>
+                <li>If <var>regionCode</var> is not a valid [[!ISO3166-2]]
+                subdivision code, throw a <a>RangeError</a> exception.
```
domenic: Does this reference give an algorithm for determining validity? I would check, but, I assume it's behind a paywall...

index.html
```diff
@@ -2138,6 +2139,22 @@ <h2>
                 </li>
               </ol>
             </li>
+            <li>If <var>details</var>["<a>regionCode</a>"] is present and not
+            the empty string:
+              <ol>
+                <li>Let <var>regionCode</var> be the result of <a>strip leading
+                and trailing ASCII whitespace</a> and <a data-cite=
```
domenic: missing "from" after "stripping leading and trailing ASCII whitespace"

index.html
```diff
@@ -2138,6 +2139,22 @@ <h2>
...
+                <var>details</var>["<a>regionCode</a>"].
+                </li>
+                <li>If <var>regionCode</var> is not a valid [[!ISO3166-2]]
+                subdivision code, throw a <a>RangeError</a> exception.
+                </li>
+                <li>Set <var>address</var>.<a>[[\regionCode]]</a> to
+                <var>regionCode</var>.
+                </li>
+              </ol>
+            </li>
```
domenic: Should we consider deriving region from regionCode, so the caller only has to pass one? I'm thinking ahead to the constructor; do we want to prevent `new PaymentAddress({ ..., regionCode: "PT-11", region: "Tasmania" })`?

I guess for the cases where there is no corresponding regionCode, we'd still want to pass region. But if regionCode is not the empty string, deriving region from it might be good...

index.html
```diff
@@ -2378,6 +2405,17 @@ <h2>
...
               </td>
             </tr>
+            <tr>
+              <td>
+                <dfn>[[\regionCode]]</dfn>
+              </td>
+              <td>
+                A <a>region</a> represented as a [[!ISO3166-2]] subdivision
+                code or the empty string, stored in its canonical uppercase
+                form. For example, "<code>PT-11</code>" represents the Lisbon
```
domenic: "or the empty string" at the end of the sentence, please. There is no canonical uppercase form for the empty string.

index.html
marcoscaceres: Behind a paywall (and I doubt it would have the algorithm... but you never know). I'll buy the spec and have a look.

marcoscaceres: ISO3166-2 defines the structure of subdivision codes, and could be usable (section 5.2)... but it's defined in terms that are not "web terms" (e.g., speaks about "characters" instead of code points). 

Options:

  1. Just send implementers to ISO3166-2, and say "valid as per Section 5.2 of ISO3166-2".
  1. Take ISO3166-2 format, and talk about it in terms of infra.
  
With 1, the interop risk is low... the format is pretty simple: 
 *  a country-code, a hyphen, 1-3 aphanum.

For 2, we could do:

The steps to validate a region code are as follows. The algorithm takes a string `regionCode` as input. The algorithm returns true if valid, false if invalid.

  1. If `regionCode` doesn't have exactly one U+002D (-), return false.
  1. Set `subCode` to the result of ASCII uppercase `regionCode`.   
  1. Split `subCode` on hyphen, and let `country` be the first item, and `subdivision` be the second item. 
  1. If `country` is not  ASCII upper alpha and length 2, return false.      
  1. If `country` is not a [[!ISO3166-1]] country code, return false.
  1. If `subdivision` is not ASCII alphanumeric and of length between 1 and 3, return false. 
  1. If  `subdivision` is not [[!ISO3166-2]] subdivision that corresponds to `country`, return false.
  1. Return true.

@domenic, happy to add 2.

index.html
marcoscaceres: > But if regionCode is not the empty string, deriving region from it might be good...

Agree. It's definitely doable. ISO3166-2 provides lookup tables, for example:

```
AD-07* Andorra la Vella
AD-02* Canillo
AD-03* Encamp
AD-08* Escaldes-Engordany
AD-04* La Massana
AD-05* Ordino
AD-06* Sant Julià de Lòria
```

Where **column 2** (defined thing) is "...the country subdivision names in the administrative language of the country concerned, where relevant with diacritic signs..." (as Unicode). Note that "a country’s administrative language is a written language used by the administration of the country at the national level". So, the result won't be in English a lot of the time - but that's fine, IMO. This is not for display purposes. 

So, something like:

The steps to derive a region from a country subdivision code element are as follows. The string takes a DOMString <var>subCode</var> as input and returns a DOMString.  
 
 1. If validate a region code <var>subCode</var> return false, throw `RangeError` exception.
 1. ASCII uppercase <var>subCode</var>, and let <var>normalizeCode</var> be the result.
 1. Find <var>normalizeCode</var> in <cite>Section 8, "List of country subdivision names and their code elements"</cite> of ISO3166-2 and return the value of "column 2" for the matching country subdivision code element. 

Sound good?

domenic: That sounds great! Probably we should get implementer buy-ins though; maybe as a separate PR? Because it'd mean having to ship that table with the browser...

index.html
domenic: Hmm. I guess 1 is good. I was going to say 2 was nice because then you don't have to pay for ISO3166-2 to implement this spec, but it turns out you would still have to do so for your step 7, so let's just stick with 1.

index.html
marcoscaceres: Agree... ☝️@rsolomakhin, @mnoorenberghe, something to start thinking about. Would appreciate your input.

mnoorenberghe: We are already using libaddressinput (and I believe Chromium does too) so it would be great if we could also use that here instead of having two versions of similar data. Does ISO3166-2 only have one name per region? What about when there are multiple official languages? Consider CA-QC: is it "[Quebec](https://chromium-i18n.appspot.com/ssl-address/data/CA)" or "[Québec](https://chromium-i18n.appspot.com/ssl-address/data/CA--fr)" in ISO3166-2. libaddressinput provides both but without a clear way to know which one would match ISO3166-2.

marcoscaceres: So, just to troll us, the spec is in English and French side-by-side 👍. Presumedly, we can just grab the "en" column 2: 

screenshot 2018-03-16 16 13 35

mnoorenberghe: > Presumedly, we can just grab the "en" column 2

I'm not quite sure how this all fits together but won't this affect what gets returned in the `region` property and therefore may be displayed back to the user on the payment confirmation page the merchant renders? If so, it would seem like we should use the language best for the user.

marcoscaceres: We can probably perform some kind of lookup, based on the document language. But then it gets into preference order for when it doesn't match the user's language (and there is no "en").

marcoscaceres: Seems crazy town, but: 

   1. Try to match document language. 
   1. Try to match user language (like Accept-Language)? 
   1. Try to match "en".
   1. Um... just give me whatever you got.

rsolomakhin: We use libaddressinput in Chromium and @sebsg is helping us to add the ISO codes to that.

rsolomakhin: > But if regionCode is not the empty string, deriving region from it might be good.

The `regionCode` field takes care of interop, so we don't necessarily have to agree on how the `region` field is derived from that. Having said that, the rule of thumb we have been using in Chrome is to use the browser user interface language when deciding between Quebec and Québec. If the browser user interface language does not match any of the languages supported by the country, then Chrome falls back to the first language specified in libaddressinput data for that country. For example, if the browser user interface language is Italian, then Canadian addresses will be in English. Although the default language works out to be English for Canada, that's not something that we should hardcode for all countries in the world, IMHO.

index.html
```diff
@@ -2138,6 +2139,75 @@ <h2>
...
+                      "INFRA#ascii-alphanumeric">ASCII alphanumeric</a> code
+                      points, in any order.
+                      </li>
+                    </ul>
+                  </div>
+                </li>
+                <li>Set <var>address</var>.<a>[[\regionCode]]</a> to
+                <var>regionCode</var>.
+                </li>
+                <li>Let <var>region</var> be the corresponding [[!ISO3166-2]]
```
marcoscaceres: @rsolomakhin, I've tried to summarize what you said about how Chrome does the ordering, while keeping the order in which things are matched optional. However, I added matching on the document's body's language as first, to match what we say in `.show()`. The "Any other criteria the user agent deems suitable" if for the fallback you described, whereby you pick the first available.  

WDYT? Would this work?

index.html
```diff
@@ -2138,6 +2139,75 @@ <h2>
...
+                      "INFRA#ascii-alphanumeric">ASCII alphanumeric</a> code
+                      points, in any order.
+                      </li>
+                    </ul>
+                  </div>
+                </li>
+                <li>Set <var>address</var>.<a>[[\regionCode]]</a> to
+                <var>regionCode</var>.
+                </li>
+                <li>Let <var>region</var> be the corresponding <a>country
```
marcoscaceres: (Adding comment again, so we can discuss here)

@rsolomakhin, I've tried to summarize what you said about how Chrome does the ordering, while keeping the order in which things are matched optional. However, I added matching on the document's body's language as first, to match what we say in `.show()`. The "Any other criteria the user agent deems suitable" if for the fallback you described, whereby you pick the first available.  

WDYT? Would this work?

index.html
```diff
@@ -2138,6 +2139,75 @@ <h2>
...
+                    <a data-cite="HTML#the-body-element-2">the body
+                    element</a>.
+                    </li>
+                    <li>The user's preferred languages.
+                    </li>
+                    <li>Any other criteria the user agent deems suitable.
+                    </li>
+                  </ol>
+                </li>
+                <li>Set <var>address</var>.<a>[[\region]]</a> to
```
marcoscaceres: @domenic, stop reading this! get back to vacation! 🍹

But when you are back... by design, `address.region` is overridable by the developer.  So, eventually:

```JS
const dict = {
   country: "PT",
   regionCode: "PT-11",
}
const address = new PaymentAddress(dict);
address.region; // "Lisboa", per ISO ISO3166-2

// And... then.. 
const dict = {
   country: "PT",
   regionCode: "PT-11",
   region: "Lisbon",
}
const address = new PaymentAddress(dict);
address.region; // Lisbon, per developer override.
```

Hmm.... now I'm wondering if we should allow some members to be nullable on the `AddressInit` dictionary:

```JS
const dict = {
   country: "PT",
   regionCode: "PT-11",
   region: null, // or undefined, means "derive it for me" 
}
``` 

While:
```JS
const dict = {
   country: "PT",
   regionCode: "PT-11",
   region: "", // trash it ... I'll handle it
}
```

index.html
rsolomakhin: This works.

marcoscaceres: @mnoorenberghe, wdty?

index.html
marcoscaceres: @domenic, if you are back from vacation, https://github.com/w3c/payment-request/pull/690#discussion_r176312874

index.html
```diff
@@ -2623,9 +2723,17 @@ <h2>
...
           </li>
-          <li>If "region" is not in <var>redactList</var>, set
-          <var>details</var>["<a>region</a>"] to the user-provided region, or
-          to the empty string if none was provided.
+          <li>If "region" is not in <var>redactList</var>:
+            <ol>
+              <li>Set <var>details</var>["<a>region</a>"] to the user-provided
+              region, or to the empty string if none was provided.
+              </li>
+              <li>If <var>details</var>["<a>region</a>"] has as a corresponding
```
mnoorenberghe: typo: "has as a"

index.html
mnoorenberghe: I'm not sure what the specific question is… do you want me to comment on the locale ordering or the spec change in general? For the locale ordering, as long as the UA isn't required to show the PR dialog in a language that the UA doesn't have installed then I'm fine with it.

mnoorenberghe: On a separate note, I wonder whether we really need to derive `region` from the `regionCode`… it would be simpler to also return the ISO3166-2 code as the `region`.

One concern I have with the spec change is that I've never seen ISO3166-2 used with the country code prefix in the wild. As a web developer I would expect `regionCode` to be "CA" for California, not "US-CA" if I hadn't read documentation on the API and only saw the webidl FWIW. I'm not sure if there are exceptions to cutting off the first 3 characters in order to get the region code specific to the country. If there is then this will likely cause issues for merchants when mapping the PR's address to existing systems.

marcoscaceres:
Can definitely just return everything after “-“, as we have the country. A 3166-2 code is always country-subregion. However, the codes are easier to look up as complete codes so my personal preference would be to not chop them up. Happy to hear other opinions.

index.html
domenic: I think `undefined` ("not present" in dictionary-speak) is the right signifier. So basically only execute this step if `details["region"]` does not exist. Otherwise set address.[[region]] to details["region"].

Whether we want to allow details["region"] to be nullable or not, as some kind of explicit "don't derive this for me, but also make it not a real region value"---I'm not sure.

index.html
marcoscaceres: > As a web developer I would expect regionCode to be "CA" for California, not "US-CA"

With the information I have at the moment, I'm leaning towards just returning the ISO3166-2 format. It might save us from having to deal with special cases where there is ambiguity, but I also don't know if there are any. I'll try to do a bit more reading.  

> I'm not sure if there are exceptions to cutting off the first 3 characters in order to get the region code specific to the country.

I don't think there are, from my reading of the ISO3166-2 spec. But I'll take a closer look as it's been a few weeks.

marcoscaceres: Coming back to this and re-reading the ISO spec, I can't determine that there would be significant issues with chopping off country and dash. I'm still reluctant to do it tho. I'd rather just stick to the full code "just in case" and just so we are not creating new conventions. @mnoorenberghe, is that ok?

stpeter: Although I'm always hesitant to parse other people's strings, ISO 3166-2 is defined to use ISO 3166-1 alpha-2 code elements as the first two characters of the string, followed by the hyphen character, followed by the region itself. So I _think_ we'd be safe in truncated the first three chars.

marcoscaceres: I guess it does also remove a source of ambiguity/redundancy, as `.country` will now need to be there to derive `.region`. Do we still attempt to do validation on a synthetic 3166-2 code? I would guess no: we attempt to populate `region` iff `county` and `regionCode` are both there and they both together form a known 3366-2 code. Can work with that.

marcoscaceres:
@mnoorenberghe ok, just subdivision part is returned now. Check it! 

Rules are:

  * if `regionCode` is present, valid country is required (or throws). 
  * If `region` is missing, and `regionCode` is present and valid, browser gives you `region`.  

Need to redo the tests and MDN docs, but would appreciate review in meantime.

marcoscaceres:
@domenic, this spec text should hopefully now be good to go 🤞

index.html
```diff
@@ -2199,6 +2201,83 @@ <h2>
                 </li>
               </ol>
             </li>
+            <li>If <var>details</var>["<a>regionCode</a>"] is present and not
+            the empty string:
+              <ol>
+                <li>Let <var>regionPart</var> be the result of <a>strip leading
```
domenic: The variable naming here is super-confusing. To a web developer, a regionCode is something like "CA", right? But in this algorithm, `details["regionCode"]` is "CA", `regionPart` is "CA", and `regionCode` is "USA-CA".

I think the variable on this line should be _regionCode_, and on line 2213 should be _putativeCountrySubdivisionCodeElement_ or similar.

index.html
```diff
@@ -2199,6 +2201,83 @@ <h2>
...
+                and trailing ASCII whitespace</a> from
+                <var>details</var>["<a>regionCode</a>"] and then
+                  <a data-cite="!INFRA#ascii-uppercase">ASCII uppercasing</a>
+                  the result.
+                </li>
+                <li>Let <var>regionCode</var> be the concatenation of
+                <var>address</var>.<a>[[\country]]</a>, a single U+002D (-) <a>
+                  code point</a>, <var>regionPart</var>.
```
domenic: Missing "and" after last comma.

domenic (approved):
Nits with variable naming, but overall makes sense...

marcoscaceres:
Added new tests, updated MDN docs.

marcoscaceres:
@ianbjacobs, could you kindly please republish the specification on TR? Thanks in advance!

marcoscaceres:
@domenic thanks again for your time and the great feedback! ✨

ianbjacobs:
@marcoscaceres, aiming for 17 May publication. Thanks all!

Update: 22 May due to moratorium.

PR #695, opened 2018-03-22
add paymentmethodchange event

### LINKED ISSUE #662: Add equivalent of applepaypaymentmethodchanged
WebKit's Payment Request implementation recently [switched](https://trac.webkit.org/changeset/225849/webkit) from the non-standard `applepaypaymentmethodchanged` event to using `PaymentDetailsModifier`s for handling payment instrument changes in Apple Pay. This allows merchants to add additional display items based on the type of card selected by the user.

Right now, Apple Pay supports the following card types: `credit`, `debit`, `prepaid`, and `store`. For the first three types, the only information WebKit provided in `applepaypaymentmethodchanged` was the card type itself, so it's an easy migration to `PaymentDetailsModifier` for these types, e.g.:

```JS
{
    total: {
        label: 'Total',
        amount: {
            currency: 'USD',
            value: '10.00',
        },
    },
    displayItems: [{
        label: 'Item',
        amount: {
            currency: 'USD',
            value: '10.00',
        },
    }],
    modifiers: [{
        supportedMethods: 'https://apple.com/apple-pay',
        total: {
            label: 'Total',
            amount: {
                currency: 'USD',
                value: '15.00',
            },
        },
        additionalDisplayItems: [{
            label: 'Credit surcharge',
            amount: {
                currency: 'USD',
                value: '5.00',
            },
        }],
        data: {
            paymentMethodType: 'credit',
        },
    }],
}
```

However, for `store` cards (private-label credit cards), Apple Pay provided the merchant associated with the card with additional account information in `applepaypaymentmethodchanged` so that they can apply discounts based on factors like "points", premium account status, etc. In the `PaymentDetailsModifier` model, there is no clear way to do this, since the additional display items will differ from customer to customer.

While `PaymentDetailsModifier` will work fine for most use cases involving credit, debit, and prepaid cards, it would be nice to also have a solution in Payment Request for private-label cards.

One solution would be to add a standard `paymentinstrumentchange` event where payment methods can include arbitrary additional information, and use that as a replacement for `applepaypaymentmethodchanged` and in addition to `PaymentDetailsModifier`. I'd be interested to hear if there are other ideas, though.

cc: @marcoscaceres @rsolomakhin

marcoscaceres:
closes #662

 * [x] [Added Web platform tests](https://github.com/w3c/web-platform-tests/pull/10912)
 * [x] [added MDN Docs](https://developer.mozilla.org/en-US/docs/Web/API/PaymentRequest/onpaymentmethodchange)
 * [ ] added MDN [compat data](https://github.com/mdn/browser-compat-data)

Implementation commitment:

 * [ ] Safari (link to issue)
 * [ ] [Chrome](https://crbug.com/824760)
 * [x] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1468356)
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/695.html" title="Last updated on Jun 21, 2018, 3:51 PM GMT (6a79bfb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/695/12ef9c4...6a79bfb.html" title="Last updated on Jun 21, 2018, 3:51 PM GMT (6a79bfb)">Diff</a>


rsolomakhin (commented):
Does this change affect ShippingAddressUpdateEvent and ShippingOptionUpdateEvent?

marcoscaceres:
Only indirectly and in a backwards compatibile way. I hope to have a draft ready for review tomorrow.

rsolomakhin:
No rush! Just checking.

marcoscaceres:
Basic card integration https://github.com/w3c/payment-method-basic-card/pull/53

index.html
```diff
@@ -3299,6 +3368,48 @@ <h2>
...
+        <ol class="algorithm">
+          <li>Let <var>request</var> be the <a>PaymentRequest</a> object that
+          the user is interacting with.
+          </li>
+          <li>Let <var>name</var> be "<a>paymentmethodchanged</a>".
+          </li>
+          <li>Let <var>methodDetails</var> be null.
+          </li>
+          <li>If <var>request</var>'s <a>payment handler</a> defines a <a>user
+          changes payment method</a> algorithm, set <var>methodDetails</var> be
```
domenic: This link is not going anywhere in the output. It should probably be a `<dfn>`?

index.html
```diff
@@ -3299,6 +3368,48 @@ <h2>
...
+          a <a data-cite="!WEBIDL#idl-dictionary">dictionary</a> or
+          <a data-cite="!WEBIDL#idl-object">object</a> resulting the user
+          selecting a different <a>payment method</a>.
```
domenic: Shouldn't this be "resulting from running that algorithm"?

index.html
```diff
@@ -3299,6 +3368,48 @@ <h2>
...
+          </li>
+          <li>Let <var>methodName</var> be the <a>payment method identifier</a>
+          of the <a>payment handler</a> the user is interacting with.
```
domenic: Is "the payment handler the user is interacting with" different from "request's payment handler"?

domenic: Actually, given that the user is potentially switching payment handlers, is the original or the new one "request's payment handler"? Is the original or the new one "the payment handler the user is interacting with"?

domenic (changes requested):
Some issues, in particular confusion about which payment handler is being referenced.

index.html
```diff
@@ -515,13 +515,15 @@ <h2>
           Promise&lt;boolean&gt; canMakePayment();
 
           readonly attribute DOMString id;
+          readonly attribute object? methodDetails;
```
domenic: A good test to write will be that `pr.methodDetails === pr.methodDetails`. The spec makes this requirement clear but it's something implementers sometimes miss.

marcoscaceres: Will need to test this with Basic Card.

marcoscaceres:
Sent tests for review https://github.com/w3c/web-platform-tests/pull/10912

marcoscaceres:
Realized that when other events fire (e.g., `onshippingaddresschange`) or generally things happen on the user interaction task source (e.g., user aborts, `request.abort()`), the `methodDetails` needs to be reset to null... same with `methodName`.

marcoscaceres:
Hmmm... having spec'ed this out, now I don't know if I like this design anymore 😓 Doesn't feel right that the only thing that changes `.methodName` and `.methodData` is this event. 

Maybe this should switch back to having special event for this, as per https://github.com/w3c/payment-request/issues/662#issuecomment-359038775

The name `paymentmethodchange` is also super ambiguous tho, because it's really a "Instrument change" of a payment method. So, I I'm kind thinking "instrumentchange" is more appropriate as an event name.  

I think I'd like to change this to: 

```JS
PaymentMethodChangeEvent : PaymentRequestUpdateEvent {
   readonly attribute DOMString methodName;
   readonly attribute object? methodDetails;
}
```

So that:

```JS
request.oninstrumentchange = ev => {
  const { type: cardType } = ev.methodDetails;
  if (ev.methodName === "https://apple.com/apple-pay") {
    switch (cardType) {
      case "store":
        // do Apple Pay specific handling for store card...
        break;
    }
  }
  // finally...
  ev.updateWith(newStuff);
};
```

@aestes wdyt?

marcoscaceres:
Updated this to match https://github.com/w3c/payment-request/pull/695#issuecomment-387978220

marcoscaceres:
Updated WPTests to match updated spec.

index.html
marcoscaceres: Added the test you suggested:  
https://github.com/w3c/web-platform-
tests/pull/10912/commits/29a384b326fc514da854a9041cdfbde13e0eee54

marcoscaceres:
@domenic, this one too should now be good to go (🤞).

index.html
```diff
@@ -3476,6 +3606,49 @@ <h2>
           </li>
         </ol>
       </section>
+      <section>
+        <h2>
+          Payment method changed algorithm
+        </h2>
+        <p>
+          The user agent MUST run the <dfn>payment method changed
+          algorithm</dfn> runs when the user changes <a>payment method</a>. The
```
domenic: extra word "runs"

index.html
```diff
@@ -3476,6 +3606,49 @@ <h2>
...
+          algorithm takes <var>methodDetails</var>, which is either
+          <a data-cite="!WEBIDL#idl-dictionary">dictionary</a> or an
+          <a data-cite="!WEBIDL#idl-object">object</a> or null, and a
+          <var>methodName</var>, which is a DOMString that represents the
+          <a>payment method identifier</a> of the <a>payment handler</a> the
+          user is interacting with:
```
domenic: How does the user agent determine _methodName_ and _methodDetails_?

index.html
```diff
@@ -3476,6 +3606,49 @@ <h2>
...
+              terminate this algorithm and take no further action. Only one
+              update may take place at a time. The <a>user agent</a> SHOULD
+              ensure that this never occurs.
+              </li>
+              <li>If the <var>request</var>.<a>[[\state]]</a> is not set to
+              "<a>interactive</a>", then terminate this algorithm and take no
+              further action. The <a>user agent</a> user interface SHOULD
+              ensure that this never occurs.
+              </li>
+              <li data-link-for="PaymentMethodChangeEvent">Fire an event
```
domenic: Ideally "fire an event" should be linked

index.html
```diff
@@ -3204,8 +3223,119 @@ <h2>
...
+        </h2>
+        <pre class="idl">
+          [Constructor(DOMString type, optional PaymentMethodChangeEventInit eventInitDict), SecureContext, Exposed=Window]
+          interface PaymentMethodChangeEvent : PaymentRequestUpdateEvent {
+            readonly attribute DOMString methodName;
+            readonly attribute object? methodDetails;
+          };
+        </pre>
+        <section>
+          <h2>
```
domenic, line 3450: Soooo I know it goes against all previous advice and good sense, but currently for events we don't use internal slots :(. Instead we use the old-style, terrible, "When getting, returns the value it was initialized with". I think making this less terrible is mostly tracked by https://github.com/whatwg/dom/issues/364, /cc @annevk. But in the end we do want to end up in a world that doesn't require people to explicitly define slots and return them for every event property, so I would take the slots out for now. Especially since nothing in this spec sets them.

domenic (changes requested):
Confused how the payment method change steps are supposed to be implemented...

index.html
marcoscaceres: Um, ok... um... we kinda added [internal slots](https://w3c.github.io/payment-request/#internal-slots-2) `PaymentRequestUpdateEvent` also. 

I guess I should remove and rephrase those too?

domenic: No, that one makes sense, as its not exposed. It's the ones that are 1:1 with public properties that are less-good.

marcoscaceres: ah, got it. Ok :)

index.html
marcoscaceres: Will rephrase this, as it's actually the Payment Handler that needs to invoke this. From discussions over in Basic Card, we are going to make this optional... so Basic Card won't fire this, but Apply Pay will.

marcoscaceres:
@domenic, I think I've addressed all the feedback 🤞

index.html
```diff
@@ -3401,8 +3420,78 @@ <h2>
...
+            <dfn>methodDetails</dfn> attribute
+          </h2>
+          <p data-link-for="PaymentMethodChangeEventInit">
+            When getting, returns the value it was initialized with. See
+            <a>methodDetails</a> member for more information.
```
domenic: Here and below, "member of PaymentMethodChangeEventInit" perhaps? Otherwise before you click on it it looks like it'll just link to itself.

index.html
```diff
@@ -3671,6 +3760,49 @@ <h2>
...
+              update may take place at a time. The <a>user agent</a> SHOULD
+              ensure that this never occurs.
+              </li>
+              <li>If the <var>request</var>.<a>[[\state]]</a> is not set to
+              "<a>interactive</a>", then terminate this algorithm and take no
+              further action. The <a>user agent</a> user interface SHOULD
+              ensure that this never occurs.
+              </li>
+              <li data-link-for="PaymentMethodChangeEvent">
+                <a>Fire an event</a> "<a>paymentmethodchange</a>" at
```
domenic: Fire an event **named**

domenic (approved):
LGTM with nits

marcoscaceres:
Added link to Firefox tracking bug.

marcoscaceres:
Added developer documentation. Please review. https://developer.mozilla.org/en-US/docs/Web/API/PaymentRequest/onpaymentmethodchange

marcoscaceres:
PaymentMethodChangeEvent is now in Firefox Nightly behind a pref. We don't currently have any plans to fire it for Basic Card, but other payment handlers could theoretically make use of it.

PR #711, opened 2018-05-16
editorial: note about inputting regions

### LINKED ISSUE #710: Clarify that empty region/regionCode is ok for some locales
In https://github.com/w3c/contact-picker/issues/65 @mnoorenberghe wrote:

> ... Since users won't expect to be asked to select their subdivision for a shipping address in some countries that don't use them for mailing, I expect that in Firefox we won't show a subdivision field and therefore won't populate the region or regionCode fields for those countries such as Belgium, Germany, etc. **Perhaps the spec and/or MDN should be more clear that an empty string can be valid for mailing purposes.**

Should add a note to the spec as per above👆

marcoscaceres:
closes #710

- [x]  Update MDN

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/711.html" title="Last updated on May 16, 2018, 7:21 AM GMT (690d96a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/711/f0018f9...690d96a.html" title="Last updated on May 16, 2018, 7:21 AM GMT (690d96a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2799,7 +2799,25 @@
           <var>details</var>["<a>recipient</a>"] to the user-provided recipient
           of the transaction, or to the empty string if none was provided.
           </li>
-          <li>If "region" is not in <var>redactList</var>:
+          <li>
+            <p>
+              If "region" is not in <var>redactList</var>:
+            </p>
+            <p data-link-for="" class="note" title=
+            "Countries where regions are not commonly used">
+              In some countries (e.g., Belgium) it is uncommon for users to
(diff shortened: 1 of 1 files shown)


mnoorenberghe (approved):
Thanks!

marcoscaceres:
Ok, MDN updated too. https://developer.mozilla.org/en-US/docs/Web/API/PaymentAddress/region

Issue #705, opened 2018-05-02
Retrying a payment 


marcoscaceres:
@domenic, @aestes, @zkoch, @adrianba, @mnoorenberghe, @stpeter, @edenchuang, all, here is an attempt to solve the retry issue. It builds on the fine grained error recovery proposal at #647, and on payment method change event (#695).  

## Retrying a payment flow
I did a bunch of experimentation, and I think the most logical solution is to mixin the handlers from `PaymentRequest` into `PaymentResponse` (or just add them directly as below?... hi @domenic!), plus add a `onpayerdetailschange` and `onpaymentmethodchange` event handlers, and a `.retry()` method.

That means that we don't need to screw around with dead `PaymentRequest`s by resetting their state machines.

The rationale for adding the new `onpayerdetailschange` being that `PaymentRequest` never gets things like `payerName`, `payerEmail`, etc. so there is no way to detect those changing. Similarly, `onpaymentmethodchange` reflects the full payment handler response, which is not available on `PaymentRequest`. 

Thus... I'd like to propose:

### IDL changes

```JS
/* see https://github.com/w3c/payment-request/issues/647#issuecomment-385852164 */
dictionary PaymentErrors {
  PayerErrors payerErrors;
  AddressErrors shippingAddressErrors;
}

partial interface PaymentResponse {
  attribute EventHandler onpayerdetailschange;
  attribute EventHandler onpaymentdetailschange;
  Promise <void> retry(PaymentErrors errors);
}
```

### `retry()` method
The `retry()` method signals that something is wrong with the sheet.
What's actually wrong with the sheet is represented by the `PaymentErrors errors` argument.

It returns a promise that resolves when the user hits Pay again (after they hopefully fix `errors`), and gets rejected if the user aborts (e.g., they hit the "esc" key while the sheet is showing). 

`retry()` can only be called multiple times, but only after each returned promise settles. It tells the browser: "let the user change the requested inputs" (i.e., what's in `PaymentOptions` and the ). Calling it prematurely, rejects a `retryPromise` with `"InvalidStateError"` .  

Empty `PaymentErrors` dictionary (i.e., `.retry({})`) means "unknown" error, meaning the user should check all their inputs. Otherwise, fix the `errors`.

## `onpayerdetailschange` event handler
The `onpayerdetailschange` fires when the user changes `name`, `email`, `phone`, depending on whether `PaymentOptions` requested them.

[...]

PR #712, opened 2018-05-16
fine-grained errors reporting for PaymentAddress

### LINKED ISSUE #647: Fine grained error recovery for fields
[Similarly to Apple Pay](https://developer.apple.com/documentation/applepayjs/applepayerrorcontactfield), we need need to define a means to signal which fields in the payment sheet suffer from a validation error.

marcoscaceres:
closes #647
 
 * [x] [Added Web platform tests](https://github.com/web-platform-tests/wpt/pull/12396)
 * [ ] added MDN Docs (link)

Implementation commitment:

 * [x] Safari - already implemented.
 * [ ] Chrome (link to issue)
 * [x] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1462907)
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/712.html" title="Last updated on Nov 2, 2018, 9:58 PM GMT (a9aee3f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/712/00a0976...a9aee3f.html" title="Last updated on Nov 2, 2018, 9:58 PM GMT (a9aee3f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -490,6 +490,39 @@
           }
         </pre>
       </section>
+      <section>
+        <h3>
+          Fine-grained error reporting
+        </h3>
+        <p>
+          A developer can use the <a data-link-for=
+          "PaymentDetailsUpdate">shippingAddressErrors</a> member of the
+          <a>PaymentDetailsUpdate</a> dictionary to indicate that there are
+          validation errors with specific attributes of a
+          <a>PaymentAddress</a>. The <a data-link-for=
+          "PaymentDetailsUpdate">shippingAddressErrors</a> member is a
+          <a>AddressErrorFields</a> dictionary, whose members specifically
+          demarcate the fields of a <a>physical address</a> that are erroneous
+          while also providing helpful error messages to be displayed to the
+          end user.
+        </p>
+        <pre class="example">
+          request.onshippingaddresschange = ev =&gt; {
+            ev.updateWith(validateAddress(request.shippingAddress));
+          };
+          function validateAddress(shippingAddress) {
+            const error = "Can't ship to this address.";
+            const shippingAddressErrors = {
+              cityError: "FarmVille is not a real place.",
+              postalCodeError: "Unknown postal code for your country.",
+            };
+            // Empty shippingOptions implies that we can't ship
+            // to this address.
+            const shippingOptions = [];
+            return { error, shippingAddressErrors, shippingOptions };
+          }
+        </pre>
+      </section>
       <section data-link-for="PaymentResponse">
         <h3>
           POSTing payment response back to a server
@@ -1713,6 +1746,7 @@
           dictionary PaymentDetailsUpdate : PaymentDetailsBase {
             DOMString error;
             PaymentItem total;
+            AddressErrorFields shippingAddressErrors;
           };
         </pre>
         <p>
@@ -1755,6 +1789,14 @@
               "PaymentItem.amount">amount</a>.<a data-lt=
               "PaymentCurrencyAmount.value">value</a> is a negative number.
             </p>
+          </dd>
+          <dt>
+            <dfn>shippingAddressErrors</dfn> member
+          </dt>
+          <dd>
+            Represents validation errors with the shipping address that is
+            associated with the <a data-cite="!DOM#event-target">event
+            target</a>.
           </dd>
         </dl>
       </section>
@@ -2040,7 +2082,7 @@
         Physical addresses
       </h2>
       <p>
-        A <dfn>physical address</dfn> is composed of the following concepts.
+        A <dfn>physical address</dfn> is composed of the following parts.
       </p>
       <dl data-sort="">
         <dt>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
First draft ready for review. This part only deals with errors related to shipping addresses. 

Will start working on #705, which will then give us the ability to correct both address-releated errors and Payer Errors (e.g., bad email) in the `PaymentResponse`.

index.html
```diff
@@ -2719,6 +2761,152 @@ <h2>
...
+        </pre>
+        <p>
+          The members of the <a>AddressErrorFields</a> dictionary represent
+          validation errors with specific parts of a <a>physical address</a>.
+          Each dictionary member has a dual function: firstly, it denotes that
+          a particular part of an address is suffering from a validation error.
+          Secondly, the string value allows the developer to describe the
+          validation error (and possibly how the end user can fix the error).
+        </p>
+        <p class="note">
```
marcoscaceres: Kept this (languageCodeError) for completeness... but wonder if we should just ditch languageCodeError. 🤔

index.html
```diff
@@ -2719,6 +2761,152 @@ <h2>
...
         </dl>
       </section>
+      <section data-dfn-for="AddressErrorFields" data-link-for=
+      "AddressErrorFields">
+        <h2>
+          <dfn>AddressErrorFields</dfn> dictionary
+        </h2>
+        <pre class="idl">
+          dictionary AddressErrorFields {
+            DOMString addressLineError;
```
marcoscaceres, line 2756: Decided to postfix all these with "Error" in the name, as I felt it made it a little bit more clear when in userland code.

index.html
marcoscaceres: Clarified what "this" was above.

marcoscaceres:
We have a partial implementation of this in Firefox now (currently based on the original proposal - but will get updated to match the new one soon).

index.html
msujaws: The extra "Error" suffix feels redundant me, as it should be implied since these are members of the shippingAddressErrors property.

marcoscaceres: depends... the problem I ran into was in destructuring and using object short hands. Also, consider the `shippingAddressErrors` can be the result of another function. Thus:

```JS
function validateAddress(address){
   // whoops, now "postalCode" has been squatted. 
   const  { postalCode } = address;
   const postalCodeError =  !isValidPostCode(postCode) ? "Invalid zip code" : undefined;
   return { postalCodeError }
}

function collectErrors(response) {
   // same problem here!
   const { shippingAddress } = response;
   const  shippingAddressErrors = validateAddress(shippingAddress);
   return { shippingAddressErrors }
}
```

And so on...

msujaws: Fair enough :)

marcoscaceres:
@domenic, when you have a few mins, would appreciate a final review. We've landed this in Firefox.

index.html
domenic: I tend to agree this is redundant. I think you have chosen a code style with a combination of object literal shorthand, destructuring, and redundant variable declarations that makes it extra confusing. But I don't think that's a good reason to verbose-ify the API, personally. I would instead suggest writing your code in a more conventional way, such as

```js
function validateAddress(address){
   return {
    postalCode: !isValidPostalCode(address.postalCode) ? "Invalid zip code" : undefined
  };
}

function collectErrors(response) {
   return {
    shippingAddress: validateAddress(response.shippingAddress)
  };
}
```

(basically, don't declare variables you're only going to use once.

index.html
```diff
@@ -2704,6 +2746,153 @@ <h2>
...
+            DOMString recipientError;
+            DOMString regionError;
+            DOMString regionCodeError;
+            DOMString sortingCodeError;
+          };
+        </pre>
+        <p>
+          The members of the <a>AddressErrorFields</a> dictionary represent
+          validation errors with specific parts of a <a>physical address</a>.
+          Each dictionary member has a dual function: firstly, it denotes that
```
domenic: I'd say "firstly, its presence denotes..."

index.html
```diff
@@ -490,6 +490,39 @@ <h3>
...
+          function validateAddress(shippingAddress) {
+            const error = "Can't ship to this address.";
+            const shippingAddressErrors = {
+              cityError: "FarmVille is not a real place.",
+              postalCodeError: "Unknown postal code for your country.",
+            };
+            // Empty shippingOptions implies that we can't ship
+            // to this address.
+            const shippingOptions = [];
+            return { error, shippingAddressErrors, shippingOptions };
```
domenic, line 522: Empty shippingOptions should not be necessary, since not-present gets converted to empty in the processing model, I am pretty sure.

domenic (approved):
Mostly LGTM, although I agree with others who are concerned about the redundancy. Will leave it to you to make the call though.

index.html
marcoscaceres: In the "[update a PaymentRequest's details algorithm](https://www.w3.org/TR/payment-request/#dfn-update-a-paymentrequest-s-details-algorithm)", we do an incremental update to `request`, so:

> If the `shippingOptions` member of details is present, and `request.[[options]].requestShipping` is true, then:
>  1. Set request.[[details]].shippingOptions to shippingOptions.

Otherwise, the original shipping options are retained - so the developer needs to be explicitly thrash them.

marcoscaceres:
@ianbjacobs, could you please republish the document on TR?

ianbjacobs:
Hi @marcoscaceres, I have send a publication request for 7 June. Thanks for all the editing!

marcoscaceres:
Added tests https://github.com/web-platform-tests/wpt/pull/12396

PR #720, opened 2018-06-07
Define PaymentResponse.prototype.retry() method

marcoscaceres:
part 1 of #705 

 * [x] [Added Web platform tests](https://github.com/w3c/web-platform-tests/pull/11101)' 
 * [x] [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/API/PaymentResponse/retry)

Implementation commitment:

 * [x] [Safari](https://bugs.webkit.org/show_bug.cgi?id=190985)
 * [x] Chrome (link to issue)
 * [x] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1435161) - currently P3 bug
 * [ ] Edge (public signal)

Unknown. 

## Example
This pull request gets us here... the user doesn't yet know what's actually wrong with the payment, but at least they know something is wrong. 

```JS
async function doPaymentRequest() {
  try {
    await recursiveValidate(request, response);
  } catch (err) { // retry aborted.
    console.error(err);
    return;
  }
}

async function recursiveValidate(request, response) {
  const promisesToFixThings = [];
  const errors = await validate(request, response);
  if (!errors) {
    return;
  }
  await response.retry();
  return recursiveValidate(request, response);
}

doPaymentRequest();
```

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/720.html" title="Last updated on Aug 31, 2018, 2:21 AM GMT (9b1e54a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/720/0d2ab3a...9b1e54a.html" title="Last updated on Aug 31, 2018, 2:21 AM GMT (9b1e54a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -666,9 +666,8 @@
                     <var>paymentMethod</var>.<a data-lt=
                     "PaymentMethodData.supportedMethods">supportedMethods</a>.
                     If it returns false, then throw a <a>RangeError</a>
-                    exception and terminate this algorithm. Optionally, inform
-                    the developer that the payment method identifier is
-                    invalid.
+                    exception. Optionally, inform the developer that the
+                    payment method identifier is invalid.
                   </li>
                   <li>If the <a data-lt="PaymentMethodData.data">data</a>
                   member of <var>paymentMethod</var> is missing, let
@@ -842,6 +841,8 @@
           <li>Set <var>request</var>.<a>[[\serializedMethodData]]</a> to <var>
             serializedMethodData</var>.
           </li>
+          <li>Set <var>request</var>.<a>[[\response]]</a> to null.
+          </li>
           <li>Set the value of <var>request</var>'s <a data-lt=
           "PaymentRequest.shippingOption">shippingOption</a> attribute to <var>
             selectedShippingOption</var>.
@@ -900,9 +901,9 @@
         <ol class="algorithm">
           <li data-tests=
           "payment-request-show-method-manual.https.html, show-method-postmessage-manual.https.html">
-          If the method was not <a>triggered by user activation</a>, return a
-          promise rejected with a "<a>SecurityError</a>" <a>DOMException</a>
-          and terminate this algorithm.
+          If the method was not <a>triggered by user activation</a>, return <a>
+            a promise rejected with</a> with a "<a>SecurityError</a>"
+            <a>DOMException</a>.
           </li>
           <li>Let <var>request</var> be the <a>context object</a>.
           </li>
@@ -1120,10 +1121,14 @@
         <ol class="algorithm">
           <li>Let <var>request</var> be the <a>context object</a>.
           </li>
+          <li>If <var>request</var>.<a>[[\response]]</a> is not null, and <var>
+            request</var>.<a>[[\response]]</a>.<a>[[\retryPromise]]</a> is not
+            null, return <a>a promise rejected with</a> an
+            "<a>InvalidStateError</a>" <a>DOMException</a>.
+          </li>
           <li>If the value of <var>request</var>.<a>[[\state]]</a> is not
-          "<a>interactive</a>" then return a promise rejected with a
-          "<a>InvalidStateError</a>" <a>DOMException</a> and terminate this
-          algorithm.
+          "<a>interactive</a>" then return <a>a promise rejected with</a> an
+          "<a>InvalidStateError</a>" <a>DOMException</a>.
           </li>
           <li>Let <var>promise</var> be <a>a new promise</a>.
           </li>
@@ -1180,8 +1185,9 @@
           then return <a>a promise rejected with</a> an
           "<a>InvalidStateError</a>" <a>DOMException</a>.
           </li>
-          <li>Optionally, at the <a>user agent</a>'s discretion, return a
-          promise rejected with a "<a>NotAllowedError</a>" <a>DOMException</a>.
+          <li>Optionally, at the <a>user agent</a>'s discretion, return <a>a
+          promise rejected with</a> a "<a>NotAllowedError</a>"
+          <a>DOMException</a>.
             <p class="note" data-link-for="PaymentRequest">
               This allows user agents to apply heuristics to detect and prevent
               abuse of the <a>canMakePayment()</a> method for fingerprinting
@@ -1441,6 +1447,15 @@
               The pending <a>Promise</a> created during <a data-lt=
               "PaymentRequest.show">show</a> that will be resolved if the user
               accepts the payment request.
+            </td>
+          </tr>
+          <tr>
+            <td>
+              <dfn>[[\response]]</dfn>
+            </td>
+            <td>
+              Null, or the <a>PaymentResponse</a> instantiated by this
+              <a>PaymentRequest</a>.
             </td>
           </tr>
         </table>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Ok, implemented this and it seems to work for me locally.

marcoscaceres:
(again noting, this doesn't include passing the actual errors yet - that's part 2.)

marcoscaceres:
@domenic, this is ready for review when you have time. Let me know how you want to handler all the parts. I tried to limit the changes to make is easier to review.

index.html
```diff
@@ -3744,17 +3869,41 @@ <h2>
...
+              undefined.
+              </li>
+              <li>Set <var>response</var>.<a>[[\complete]]</a> to false.
+              </li>
+              <li>Set the <a data-lt="PaymentResponse.requestId">requestId</a>
+              attribute value of <var>response</var> to the value of
+              <var>request</var>.<a>[[\details]]</a>.<a data-lt=
+              "PaymentDetailsInit.id">id</a>.
+              </li>
+              <li>Set the <a data-lt=
```
marcoscaceres: @zkoch, question for you: upon a retry, will it be possible for a user to change payment handlers. For instance, can a user switch from "basic-card" to "https://bob.pay"? Seems like they probably should be able to.

ianbjacobs: My view is yes, they should be able to do so.

marcoscaceres:
Seeking additional implementation commitment for `.retry()` also. Currently, only have commitment from Mozilla.

marcoscaceres:
@ianbjacobs, @adrianhopebailie, could you evaluate if there is any impact on payment handler? I've not given it any thought. If there is, could you add details to the "Impact on Payment Handler spec?" Same applies for the other pull requests relating to `retry()`.

index.html
marcoscaceres: I'll move it down, so it's not part of the initialization.

index.html
```diff
@@ -3131,12 +3147,84 @@ <h2>
...
+          <li>If <var>response</var>.<a>[[\complete]]</a> is true, return <a>a
+          promise rejected with</a> an "<a>InvalidStateError</a>"
+          <a>DOMException</a>.
+          </li>
+          <li>If <var>response</var>.<a>[[\retrying]]</a> is true, return <a>a
+          promise rejected with</a> an "<a>InvalidStateError</a>"
+          <a>DOMException</a>.
+          </li>
+          <li>Set <var>response</var>.<a>[[\retrying]]</a> to true.
+          </li>
```
domenic: You may be able to get away with consolidating [[retryPromise]] and [[retrying]] into just [[retryPromise]], which will be null if you are not retrying. This might also be a good change because then you'd make it clearer that the UA can release the reference to the promise, when you set it to null.

index.html
```diff
@@ -3131,12 +3147,84 @@ <h2>
...
+                  <li>Set the <a>user agent</a>'s <a>payment request is
+                  showing</a> boolean to false.
+                  </li>
+                  <li>Reject <var>retryPromise</var> with an
+                  "<a>AbortError</a>" <a>DOMException</a>.
+                  </li>
+                </ol>
+                <p>
+                  Finally, when <var>retryPromise</var> settles, set
+                  <var>response</var><a>[[\retrying]]</a> to false.
```
domenic: There are a lot of missing periods in this PR. Search for `</var><a>[[` (should be `</var>.<a>[[`).

index.html
```diff
@@ -3131,12 +3147,84 @@ <h2>
...
+          <var>retryPromise</var>.
+          </li>
+          <li>
+            <a>In parallel</a>:
+            <ol>
+              <li>In the payments UI, indicate to the end-user that something
+              is wrong with the user-provided data of the payment response.
+              </li>
+              <li>
+                <p>
```
domenic: The first paragraph here doesn't seem like a numbered step. I would put it as a NOTE inside the step 10, "return retryPromise".

index.html
```diff
@@ -3131,12 +3147,84 @@ <h2>
...
+          </li>
+          <li>Set <var>request</var>.<a>[[\state]]</a> to "<a>interactive</a>".
+          </li>
+          <li>Let <var>retryPromise</var> be <a>a new promise</a>.
+          </li>
+          <li>Set <var>response</var>.<a>[[\retryPromise]]</a> to
+          <var>retryPromise</var>.
+          </li>
+          <li>
+            <a>In parallel</a>:
```
domenic: I don't think any of these steps should be done in parallel (on a background thread). Updating the UI needs to be done from the UI thread. And the other steps are basically just setting up callbacks on main-thread objects.

index.html
```diff
@@ -3131,12 +3147,84 @@ <h2>
...
+                  rejected by either the <a>user accepts the payment request
+                  algorithm</a> or the <a>user aborts the payment request
+                  algorithm</a>, which are triggered through interaction with
+                  the user interface.
+                </p>
+                <p data-test="rejects_if_not_active.https.html">
+                  If <var>document</var> stops being <a data-cite=
+                  "!HTML#fully-active">fully active</a> while the user
+                  interface is being shown, or no longer is by the time this
+                  step is reached, then:
```
domenic: I think you also want a fully active check earlier in the algorithm (before you present the UI in the first place), as is done in show().

index.html
```diff
@@ -3131,12 +3147,84 @@ <h2>
...
+                  <li>Close down the user interface.
+                  </li>
+                  <li>Set the <a>user agent</a>'s <a>payment request is
+                  showing</a> boolean to false.
+                  </li>
+                  <li>Reject <var>retryPromise</var> with an
+                  "<a>AbortError</a>" <a>DOMException</a>.
+                  </li>
+                </ol>
+                <p>
```
domenic: This paragraph should be its own standalone step.

index.html
```diff
@@ -3338,11 +3434,40 @@ <h2>
...
+              The <a>PaymentRequest</a> instance that instantiated this
+              <a>PaymentResponse</a>.
+            </td>
+          </tr>
+          <tr>
+            <td>
+              <dfn>[[\retryPromise]]</dfn>
+            </td>
+            <td>
+              When <a>[[\retrying]]</a> is true, a <a>Promise</a> resolves when
```
domenic: A Promise **that** resolves...

index.html
```diff
@@ -3744,21 +3869,43 @@ <h2>
...
+          <li>If <var>isRetry</var> if false, initialize the newly created
+          <var>response</var>:
+            <ol>
+              <li>Set <var>response</var>.<a>[[\request]]</a> to
+              <var>request</var>.
+              </li>
+              <li>Set <var>response</var>.<a>[[\retrying]]</a> to false.
+              </li>
+              <li>Set <var>response</var>.<a>[[\retryPromise]]</a> to
+              undefined.
```
domenic: null, not undefined, is what you used elsewhere for the default here

domenic (approved):
Mostly looks good, lots of editorial changes... Feel free to merge after fixing if you want to get a move on, or I can take another pass.

Also, as a nit on the commit message, it's PaymentResponse.prototype.retry() or paymentResponse.retry(), not PaymentResponse.retry().

index.html
marcoscaceres: Great suggestion! I'll make this change.

index.html
marcoscaceres: Good point. But we still need this for when the UI is being presented (as happens in `show()` step 19). Adding both checks.

marcoscaceres:
Added some developer docs:
https://developer.mozilla.org/en-US/docs/Web/API/PaymentResponse/retry

marcoscaceres:
Blink intent to implement https://groups.google.com/a/chromium.org/d/msg/blink-dev/wayZGnuBkrI/SCLOTxCACwAJ

aestes:
WebKit tracking bug: https://bugs.webkit.org/show_bug.cgi?id=190985

PR #724, opened 2018-06-12
Add PaymentResponse.prototype.onpayerdetailchange 

marcoscaceres:
part 4 of #705 - define eventing model for "payerdetailchange". 

 * [x] [Added Web platform tests](https://github.com/web-platform-tests/wpt/pull/11772)
 * [x] [added MDN Docs](https://developer.mozilla.org/en-US/docs/Web/API/PaymentResponse/onpayerdetailchange)

Implementation commitment:

 * [x] [Safari](https://bugs.webkit.org/show_bug.cgi?id=189249)
 * [ ] Chrome (link to issue)
 * [x] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1472026)
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/724.html" title="Last updated on Nov 2, 2018, 9:56 PM GMT (397c3b6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/724/b417638...397c3b6.html" title="Last updated on Nov 2, 2018, 9:56 PM GMT (397c3b6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3850,7 +3850,7 @@
             "DOM#event-target">target</a> is an instance of
             <a>PaymentResponse</a>, let <var>request</var> be
             <var>event</var>'s <a data-cite=
-            "DOM#event-target">target</a><a>[[\request]]</a>.
+            "DOM#event-target">target</a>.<a>[[\request]]</a>.
             </li>
             <li>Otherwise, let <var>request</var> be the value of
             <var>event</var>'s <a data-cite="DOM#event-target">target</a>.
@@ -4021,6 +4021,75 @@
           identifier</a> of the <a>payment handler</a> the user is interacting
           with:
         </p>
+                <ol class="algorithm">
+          <li>Let <var>request</var> be the <a>PaymentRequest</a> object that
+          the user is interacting with.
+          </li>
+          <li>
+            <a>Queue a task</a> on the <a>user interaction task source</a> to
+            run the following steps:
+            <ol>
+              <li>Assert: <var>request</var>.<a>[[\updating]]</a> is false.
+              Only one update can take place at a time.
+              </li>
+              <li>Assert: <var>request</var>.<a>[[\state]]</a> is
+              "<a>interactive</a>".
+              </li>
+              <li data-link-for="PaymentMethodChangeEvent">
+                <a>Fire an event</a> named "<a>paymentmethodchange</a>" at
+                <var>request</var> using <a>PaymentMethodChangeEvent</a>, with
+                its <a>methodName</a> attribute initialized to
+                <var>methodName</var>, and its <a>methodDetails</a> attribute
+                initialized to <var>methodDetails</var>.
+              </li>
+            </ol>
+          </li>
+        </ol>
+      </section>
+      <section>
+        <h2>
+          PaymentRequest updated algorithm
+        </h2>
+        <p data-tests="algorithms-manual.https.html">
+          The <dfn>PaymentRequest updated algorithm</dfn> is run by other
+          algorithms above to <a>fire an event</a> to indicate that a user has
+          made a change to a <a>PaymentRequest</a> called <var>request</var>
+          with an event name of <var>name</var>:
+        </p>
+        <ol class="algorithm">
+          <li>Assert: <var>request</var>.<a>[[\updating]]</a> is false. Only
+          one update can take place at a time.
+          </li>
+          <li>Assert: <var>request</var>.<a>[[\state]]</a> is
+          "<a>interactive</a>".
+          </li>
+          <li>Let <var>event</var> be the result of <a data-cite=
+          "!DOM#concept-event-create">creating an event</a> using the
+          <a>PaymentRequestUpdateEvent</a> interface.
+          </li>
+          <li>Initialize <var>event</var>'s <code><a data-cite=
+          "!DOM#dom-event-type">type</a></code> attribute to <var>name</var>.
+          </li>
+          <li>
+            <a data-cite="!DOM#concept-event-dispatch">Dispatch</a>
+            <var>event</var> at <var>request</var>.
+          </li>
+          <li data-link-for="PaymentRequestUpdateEvent">If
+          <var>event</var>.<a>[[\waitForUpdate]]</a> is true, disable any part
+          of the user interface that could cause another update event to be
+          fired.
+          </li>
+        </ol>
+      </section>
+      <section>
+        <h2>
+          Payer detail changed algorithm
+        </h2>
+        <p>
+          The user agent MUST run the <dfn>payer detail changed algorithm</dfn>
+          when the user changes the <var>payer name</var>, or the <var>payer
+          email</var>, or the <var>payer phone</var> in the user interface:
+        </p>
         <ol class="algorithm">
           <li>Let <var>request</var> be the <a>PaymentRequest</a> object that
           the user is interacting with.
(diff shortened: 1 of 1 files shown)


ianbjacobs:
@marcoscaceres, it has been pointed out that the proposal as written (if I read the code correctly) requires the developer to query all the fields to find out which thing changed. What would be your suggested good practice for figuring out which thing changed?

marcoscaceres:
> What would be your suggested good practice for figuring out which thing changed?

There is not way to detect which changed, just revalidate. So the question is mostly around validation of these values... 

```JS
response.onpayerdetailchange = ev => {
    const errors = {};
    const { payerName, payerEmail, payerPhone } = response; 
    object.assign(
       errors, validateEmail(payerEmail), validateName(payerName), validatePhone(payerPhone) 
    );
   if(Object.getOwnPropertyNames(x).length){
      response.retry(errors);
   }
}
```

marcoscaceres:
actually, to be safe, maybe the should be separate events 🧐... if each of those requires a network lookup for validation, then this could get a little bit sad. Question is, is anyone doing this validation over the network?

stpeter:
@marcoscaceres What would the separate events be?

marcoscaceres:
`onpayeremailchange`, `onpayerphonechange`, and `onpayernamechange`.

stpeter:
That's what I was afraid of, but I see why we might need it...

marcoscaceres:
The problem is that if UI autofills the three fields at the same time, we get into a race condition.

ianbjacobs:
Ah, I had not thought about the autofill scenario; I was only imagining independent manual interaction.

marcoscaceres:
Discussed with our UI folks too. The single event is better for us, because it gives us more flexibility as to when we fire the event (e.g., clicking "next").

marcoscaceres:
so, this is how one would handle having the single event:  

```JS
let {
  payerName: oldPayerName,
  payerEmail: oldPayerEmail,
  payerPhone: oldPayerPhone,
} = response;
response.onpayerdetailchange = async ev => {
  const promisesToValidate = [];
  const { payerName, payerEmail, payerPhone } = response;
  if (oldPayerName !== payerName) {
    promisesToValidate.push(validateName(payerName));
    oldPayerName = payerName;
  }
  if (oldPayerEmail !== payerEmail) {
    promisesToValidate.push(validateEmail(payerEmail));
    oldPayerEmail = payerEmail;
  }
  if (oldPayerPhone !== payerPhone) {
    promisesToValidate.push(validatePhone(payerPhone));
    oldPayerPhone = payerPhone;
  }
  const errors = await Promise.all(promisesToValidate).then(results =>
    results.reduce((errors, result), Object.assign(errors, result))
  );
  if (Object.getOwnPropertyNames(errors).length) {
    await response.retry(errors);
  }
};
```

DannyRussellWP:
Based on my experience of implementing payments request on multiple demo's for worldpay now. I would echo the recommendation of marco's ui guys and leave the single event. While a complicated handler is not perfect I admit it does keep the code a lot clearer than having a lot of separate event handlers

ianbjacobs:
Hi @marcoscaceres,

I have been asking around for views on the question of "one event" v. "multiple events". You can see that @DannyRussellWP supported a single event. I've heard back similarly from two other payments companies as well.

Ian

marcoscaceres:
Excellent, thanks for the update @ianbjacobs.

wanli:
@marcoscaceres @ianbjacobs  - The preference from Airbnb is that one event handler would suffice.

I asked internally and have updated this action regarding this:
https://www.w3.org/Payments/WG/track/actions/97

marcoscaceres:
Appreciate the feedback @wanli!

index.html
```diff
@@ -3812,8 +3846,14 @@ <h2>
             <li>If <var>event</var>.<a>[[\waitForUpdate]]</a> is true, then <a>
               throw</a> an "<a>InvalidStateError</a>" <a>DOMException</a>.
             </li>
-            <li>Let <var>request</var> be the value of <var>event</var>'s
-            <a data-cite="DOM#event-target">target</a>.
+            <li>If <var>event</var>'s <a data-cite=
+            "DOM#event-target">target</a> is an instance of
+            <a>PaymentResponse</a>, let <var>request</var> be
+            <var>event</var>'s <a data-cite=
+            "DOM#event-target">target</a><a>[[\request]]</a>.
```
domenic: Missing dot

index.html
```diff
@@ -3985,26 +4025,70 @@ <h2>
           <li>Let <var>request</var> be the <a>PaymentRequest</a> object that
           the user is interacting with.
           </li>
+          <li>If <var>request</var><a>[[\response]]</a> is null, terminate this
```
domenic: Missing dot

Use "return" instead of "terminate this algorithm"

index.html
```diff
@@ -3985,26 +4025,70 @@ <h2>
...
+          algorithm.
+          </li>
+          <li>Let <var>response</var> be
+          <var>request</var><a>[[\response]]</a>.
```
domenic: Missing dot

Move this step up so that you can check if _response_ is null.

index.html
```diff
@@ -3985,26 +4025,70 @@ <h2>
...
-              <li data-link-for="PaymentMethodChangeEvent">
-                <a>Fire an event</a> named "<a>paymentmethodchange</a>" at
-                <var>request</var> using <a>PaymentMethodChangeEvent</a>, with
-                its <a>methodName</a> attribute initialized to
-                <var>methodName</var>, and its <a>methodDetails</a> attribute
-                initialized to <var>methodDetails</var>.
+              <li>If <var>payer name</var> changed:
+                <ol>
+                  <li data-link-for="PaymentOptions">Assert:
+                  <var>request</var><a>[[\options]]</a>.<a>requestPayerName</a>
```
domenic: Search for `</var><a>[[`, so many missing dots...

index.html
```diff
@@ -3985,26 +4025,70 @@ <h2>
           <li>Let <var>request</var> be the <a>PaymentRequest</a> object that
```
domenic: The algorithms seem to be switched; this is the payment details changed algorithm.

domenic (changes requested):
A little hard to review due to the switched headings... let's get that fixed.

index.html
marcoscaceres: Argh, oops... screwed up the merge/

marcoscaceres:
@domenic, hopefully less bad now 🤞

index.html
```diff
@@ -3981,23 +4021,19 @@ <h2>
           identifier</a> of the <a>payment handler</a> the user is interacting
           with:
         </p>
-        <ol class="algorithm">
+                <ol class="algorithm">
```
domenic: Whitspace change seems bad

index.html
```diff
@@ -4049,6 +4081,88 @@ <h2>
...
+          The user agent MUST run the <dfn>payer detail changed algorithm</dfn>
+          when the user changes the <var>payer name</var>, or the <var>payer
+          email</var>, or the <var>payer phone</var> in the user interface:
+        </p>
+        <ol class="algorithm">
+          <li>Let <var>request</var> be the <a>PaymentRequest</a> object that
+          the user is interacting with.
+          </li>
+          <li>If <var>request</var>.<a>[[\response]]</a> is null, terminate
+          this algorithm.
```
domenic: "terminate this algorithm" -> "return"

index.html
```diff
@@ -4049,6 +4081,88 @@ <h2>
...
+              "<a>payerdetailchange</a>".
+              </li>
+              <li>
+                <a data-cite="!DOM#concept-event-dispatch">Dispatch</a>
+                <var>event</var> at <var>response</var>.
+              </li>
+              <li data-link-for="PaymentRequestUpdateEvent">If
+              <var>event</var>.<a>[[\waitForUpdate]]</a> is true, disable any
+              part of the user interface that could cause another update event
+              to be fired.
```
domenic: Is "another update event" appropriate here? Maybe "another change to the payer details"?

domenic (approved):
LGTM!

marcoscaceres:
Added Gecko tracking bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1472026

marcoscaceres:
Dev docs added https://developer.mozilla.org/en-US/docs/Web/API/PaymentResponse/onpayerdetailchange

marcoscaceres:
Added tests https://github.com/web-platform-tests/wpt/pull/11772

marcoscaceres:
Firefox implementation underway.

aestes:
WebKit tracking bug: https://bugs.webkit.org/show_bug.cgi?id=189249

marcoscaceres:
@romandev, any news on this one (`PaymentResponse.prototype.onpayerdetailchange`) from the Chrome side?

romandev:
Ah, I didn't create a new separate bug but it's already finished here:
https://chromium-review.googlesource.com/c/chromium/src/+/1206750

romandev:
Chrome tracking bug: https://bugs.chromium.org/p/chromium/issues/detail?id=901710

marcoscaceres:
Awesome. Looks like I need to do another run of WTP to update the implementation report.

PR #726, opened 2018-06-12
 Editorial: use internal slots

### LINKED ISSUE #705: Retrying a payment

marcoscaceres:
Finally, part 5 of #705  - closes #705 

Defines PaymentRequest's internal slots. 
In some cases, PaymentResponse just puts forward PaymentRequest's internal slot values.  

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/726.html" title="Last updated on Jun 12, 2018, 1:25 PM GMT (4c766a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/726/12ef9c4...4c766a5.html" title="Last updated on Jun 12, 2018, 1:25 PM GMT (4c766a5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3143,33 +3143,40 @@
           <dfn>retry()</dfn> method
         </h2>
         <ol class="algorithm">
-          <li>Let response be the context object.
+          <li>Let <var>response</var> be the <a>context object</a>.
           </li>
-          <li>Let request be response.[[\request]].
+          <li>Let <var>request</var> be
+          <var>response</var>.<a>[[\request]]</a>.
           </li>
-          <li>If response.[[\completeCalled]] is true, throw a
-          "InvalidStateError" DOMException.
+          <li>If <var>response</var>.<a>[[\complete]]</a> is true, throw a
+          "<a>InvalidStateError</a>" <a>DOMException</a>.
           </li>
-          <li>If request.[[\state]] is not "closed", throw a
-          "InvalidStateError" DOMException.
+          <li>If <var>request</var>.<a>[[\retrying]]</a> is true, throw a
+          "<a>InvalidStateError</a>" <a>DOMException</a>.
           </li>
-          <li>Set request.[[\state]] to "interactive".
+          <li>If <var>request</var>.<a>[[\state]]</a> is not "<a>closed</a>",
+          throw a "<a>InvalidStateError</a>" <a>DOMException</a>.
           </li>
-          <li>Set request.[[\retrying]] to true.
+          <li>Set <var>response</var>.<a>[[\retrying]]</a> to true.
           </li>
-          <li>Let promiseToRetry be a newly created promise.
+          <li>Set <var>request</var>.<a>[[\state]]</a> to "<a>interactive</a>".
           </li>
-          <li>In parallel:
+          <li>Set <var>request</var>.<a>[[\retryPromise]]</a> to a newly
+          created promise.
+          </li>
+          <li>
+            <a>In parallel</a>:
             <ol>
-              <li>If the <a>user accepts the retry</a>, resolve promiseToRetry
-              with undefined.
+              <li>Indicate to the end-user that something is wrong with the
+              data of the payment response.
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Closing this one... will send again.

PR #733, opened 2018-06-20
Fix a small bug in `abort`'s error handling.

mrbkap:
Because `PaymentRequest.abort` returns a `Promise<void>`, it shouldn't be specified to throw exceptions. Instead, it should return rejected promises.

Somewhat complicating this is that Safari does throw an exception (even though Firefox and Blink do not). This has an impact on web developers as an eager exception will stop their scripts and a rejected promise will not.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1113,8 +1113,9 @@
           <li>Let <var>request</var> be the <a>context object</a>.
           </li>
           <li>If the value of <var>request</var>.<a>[[\state]]</a> is not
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):

marcoscaceres:
Marked as non substantive for IPR from ash-nazg.

marcoscaceres:
Thanks, @mrbkap! Probably a few of those we missed... too much async/await corrupting our editorial practices :)

PR #749, opened 2018-07-06
Support requesting billing address

marcoscaceres:
Part 1 for #27 (part 2 is #750)

 * [x] [Added Web platform tests](https://github.com/web-platform-tests/wpt/pull/14255)
 * [x] added MDN Docs - will be done as part of the with Firefox work. 
 * [ ] added MDN [compat data](https://github.com/mdn/browser-compat-data)

Implementation commitment:

 * [ ] Safari (link to issue)
 * [ ] [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=909634)
 * [x] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1477117)
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/749.html" title="Last updated on Nov 28, 2018, 10:17 AM GMT (e91e6b4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/749/d8faeb9...e91e6b4.html" title="Last updated on Nov 28, 2018, 10:17 AM GMT (e91e6b4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4059,9 +4059,16 @@
             </li>
             <li>Set <var>event</var>.<a>[[\waitForUpdate]]</a> to true.
             </li>
+            <li>Let <var>identifer</var> be undefined.
+            </li>
+            <li>If <var>event</var> has a <a data-link-for=
+            "PaymentMethodChangeEvent">methodName</a> attribute, set
+            <var>identifer</var> to the <a data-link-for=
+            "PaymentMethodChangeEvent">methodName</a> attribute's value.
+            </li>
             <li>Run the <a>update a <code>PaymentRequest</code>'s details
-            algorithm</a> with <var>detailsPromise</var> and
-            <var>request</var>.
+            algorithm</a> with <var>detailsPromise</var>, <var>request</var>,
+            and <var>identifer</var>.
             </li>
           </ol>
         </section>
@@ -4614,9 +4621,10 @@
         <p>
           The <dfn>update a <code>PaymentRequest</code>'s details
           algorithm</dfn> takes a <a>PaymentDetailsUpdate</a>
-          <var>detailsPromise</var> and a <a>PaymentRequest</a>
-          <var>request</var>. The steps are conditional on the
-          <var>detailsPromise</var> settling. If <var>detailsPromise</var>
+          <var>detailsPromise</var>, a <a>PaymentRequest</a>
+          <var>request</var>, and optionally a DOMString <var>identifer</var>
+          (a <a>payment method identifier</a>). The steps are conditional on
+          the <var>detailsPromise</var> settling. If <var>detailsPromise</var>
           never settles then the payment request is blocked. The user agent
           SHOULD provide the user with a means to abort a payment request.
           Implementations MAY choose to implement a timeout for pending updates
@@ -4802,6 +4810,28 @@
                         </ol>
                       </li>
                     </ol>
+                  </li>
+                </ol>
+              </li>
+              <li>If the <a data-lt="PaymentDetailsUpdate.paymentMethodErrors">
+                paymentMethodErrors</a> member is present and
+                <var>identifier</var> is not undefined:
+                <ol>
+                  <li>If required by the specification that defines the
+                  <var>identifer</var>, then <a data-cite=
+                  "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a> <a>
+                    paymentMethodErrors</a> to an IDL value. Otherwise,
+                    <a data-cite=
+                    "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+                    <a>paymentMethodErrors</a> to <a data-cite=
+                    "!WEBIDL#idl-object">object</a>.
+                  </li>
+                  <li>If conversion results in a <a data-cite=
+                  "!WEBIDL#dfn-exception">exception</a> <var>error</var>,
+                  <a>abort the update</a> with <var>error</var>.
+                  </li>
+                  <li>The <a>payment handler</a> SHOULD display an error for
+                  each relevant erroneous field of <a>paymentMethodErrors</a>.
                   </li>
                 </ol>
               </li>
--- a/index.html
+++ b/index.html
@@ -1972,6 +1972,7 @@
       <pre class="idl">
         dictionary PaymentOptions {
           boolean requestPayerName = false;
+          boolean requestBillingAddress = false;
           boolean requestPayerEmail = false;
           boolean requestPayerPhone = false;
           boolean requestShipping = false;
@@ -1984,6 +1985,19 @@
         options desired for the payment request.
       </p>
       <dl>
+        <dt>
+          <dfn>requestBillingAddress</dfn> member
+        </dt>
+        <dd data-link-for="PaymentMethodChangeEvent">
+          A boolean that instructs the <a>user agent</a> to get the billing
+          address associated with a <a>payment method</a> (e.g., the billing
+          address associated with a credit card). Typically, the user agent
+          will return the billing address as part of the
+          <a>PaymentMethodChangeEvent</a>'s <a>methodDetails</a>, albeit
+          possibly with parts of the address redacted for privacy reasons. A
+          merchant can use this information to, for example, calculate tax in
+          certain jurisdictions.
+        </dd>
         <dt>
           <dfn>requestPayerName</dfn> member
         </dt>
--- a/index.html
+++ b/index.html
@@ -584,11 +584,13 @@
           readonly attribute PaymentAddress? shippingAddress;
           readonly attribute DOMString? shippingOption;
           readonly attribute PaymentShippingType? shippingType;
+          readonly attribute PaymentAddress? payerAddress;
 
           attribute EventHandler onmerchantvalidation;
           attribute EventHandler onshippingaddresschange;
           attribute EventHandler onshippingoptionchange;
           attribute EventHandler onpaymentmethodchange;
+          attribute EventHandler onpayeraddresschange;
         };
       </pre>
       <div class="note">
@@ -855,6 +857,10 @@
           </li>
           <li>Set the value of the <a data-lt="PaymentRequest.shippingAddress">
             shippingAddress</a> attribute on <var>request</var> to null.
+          </li>
+          <li>Set the value of the <a data-lt=
+          "PaymentRequest.payerAddress">payerAddress</a> attribute on
+          <var>request</var> to null.
           </li>
           <li>If <var>options</var>.<a data-lt=
           "PaymentOptions.requestShipping">requestShipping</a> is set to true,
@@ -1265,6 +1271,24 @@
           "PaymentOptions.shippingType">shippingType</a> member).
         </p>
       </section>
+      <section>
+        <h2>
+          <dfn>payerAddress</dfn> attribute
+        </h2>
+        <p>
+          If the <a data-lt=
+          "PaymentOptions.requestPayerAddress">requestPayerAddress</a> member
+          was set to true in the <a>PaymentOptions</a> passed to the
+          <a>PaymentRequest</a> constructor, then <a>payerAddress</a> will be
+          the payer's address chosen by the user.
+        </p>
+        <p class="note">
+          On some user agents, the some attributes of the <a>payerAddress</a>'s
+          might be redacted until the <a>user accepts the payment request</a> -
+          at which point all the user provided details of <a>payerAddress</a>
+          become available.
+        </p>
+      </section>
       <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
         <h2>
           <dfn>onmerchantvalidation</dfn> attribute
@@ -1317,6 +1341,16 @@
           A <a>PaymentRequest</a>'s <a>onpaymentmethodchange</a> attribute is
           an <a>EventHandler</a> for a <a>PaymentMethodChangeEvent</a> named
           "<a>paymentmethodchange</a>".
+        </p>
+      </section>
+      <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
+        <h2>
+          <dfn>onpayeraddresschange</dfn> attribute
+        </h2>
+        <p>
+          A <a>PaymentRequest</a>'s <a>onpayeraddresschange</a> attribute is an
+          <a>EventHandler</a> for a <a>PaymentRequestUpdateEvent</a> named
+          "<a>payeraddresschange</a>".
         </p>
       </section>
       <section>
@@ -1975,6 +2009,7 @@
           boolean requestPayerEmail = false;
           boolean requestPayerPhone = false;
           boolean requestShipping = false;
+          boolean requestPayerAddress = false;
           PaymentShippingType shippingType = "shipping";
         };
       </pre>
@@ -2036,6 +2071,15 @@
             The <a>shippingType</a> member only affects the user interface for
             the payment request.
           </p>
+        </dd>
+        <dt>
+          <dfn>requestPayerAddress</dfn> member
+        </dt>
+        <dd>
+          A boolean that instructs the <a>user agent</a> to collect and return
+          a payer's address (also commonly referred to as a "billing address").
+          A merchant can use this information to, for example, calculate tax in
+          certain jurisdictions.
         </dd>
       </dl>
     </section>
@@ -3059,6 +3103,7 @@
           readonly attribute DOMString? payerName;
           readonly attribute DOMString? payerEmail;
           readonly attribute DOMString? payerPhone;
+          readonly attribute PaymentAddress? payerAddress;
 
           [NewObject]
           Promise&lt;void&gt; complete(optional PaymentComplete result = "unknown");
@@ -3408,6 +3453,19 @@
           "PaymentRequest.id">id</a> that spawned this payment response.
         </p>
       </section>
+      <section>
+        <h2>
+          <dfn>payerAddress</dfn> attribute
+        </h2>
+        <p>
+          If the <a data-lt=
+          "PaymentOptions.requestPayerAddress">requestPayerAddress</a> member
+          was set to true in the <a>PaymentOptions</a> passed to the
+          <a>PaymentRequest</a> constructor, then <a data-lt=
+          "PaymentRequest.payerAddress">payerAddress</a> will be the
+          un-redacted payer's address chosen by the user.
+        </p>
+      </section>
       <section data-dfn-for="PaymentResponse" data-link-for="PaymentResponse">
         <h2>
           <dfn data-lt="complete(result)">complete()</dfn> method
@@ -3645,6 +3703,20 @@
             <td>
               The user chooses a different <a>payment method</a> within a
               <a>payment handler</a>.
+            </td>
+            <td>
+              <a>PaymentRequest</a>
+            </td>
+          </tr>
+          <tr>
+            <td>
+              <code><dfn>payeraddresschange</dfn></code>
+            </td>
+            <td>
+              <a>PaymentRequestUpdateEvent</a>
+            </td>
+            <td>
+              The user provides a new payer address.
             </td>
             <td>
               <a>PaymentRequest</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1272,6 +1278,24 @@ <h2>
...
+      <section>
+        <h2>
+          <dfn>payerAddress</dfn> attribute
+        </h2>
+        <p>
+          If the <a data-lt=
+          "PaymentOptions.requestPayerAddress">requestPayerAddress</a> member
+          was set to true in the <a>PaymentOptions</a> passed to the
+          <a>PaymentRequest</a> constructor, then <a>payerAddress</a> will be
+          the payer's address chosen by the user.
```
domenic: This paragraph seems like it could be more similar to the shipping address one. E.g., mentioning the payer address changed algorithm.

index.html
```diff
@@ -1272,6 +1278,24 @@ <h2>
...
+        </p>
+        <p class="note">
+          On some user agents, the some attributes of the <a>payerAddress</a>'s
```
domenic: "the some" -> "some"

"the payerAddress's" -> "payerAddress"

index.html
```diff
@@ -1272,6 +1278,24 @@ <h2>
...
+          might be redacted until the <a>user accepts the payment request</a> -
```
domenic: hyphen minus -> either em dash (—) or just use a comma

index.html
```diff
@@ -1272,6 +1278,24 @@ <h2>
...
+          at which point all the user provided details of <a>payerAddress</a>
```
domenic: "user provided" -> "user-provided"

index.html
```diff
@@ -4255,6 +4381,28 @@ <h2>
...
+              <code>PaymentAddress</code> from user-provided input</a> with
+              <var>redactList</var>.
+              </li>
+              <li>Set the <a data-lt=
+              "PaymentResponse.payerAddress">payerAddress</a> attribute value
+              of <var>response</var> to <var>payerAddress</var>.
+              </li>
+              <li>Set the <a data-lt=
+              "PaymentRequest.payerAddress">payerAddress</a> attribute value of
+              <var>request</var> to <var>payerAddress</var>.
```
domenic: It'll be worth testing that UAs create distinct objects here and change payerAddress to the new PaymentAddress object. That is, they can't mutate the existing object; per this spec they have to create a new one.

domenic (changes requested):
All nits

domenic (approved):

domenic:
Hmm, substantive concerns at https://github.com/w3c/payment-request/issues/27#issuecomment-403614316; I'd appreciate your (or others') thoughts there before merging this.

index.html
marcoscaceres: Will make sure to test that.

marcoscaceres:
Gecko bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1477117

marcoscaceres:
Closing, due to lack of consensus.

marcoscaceres:
@adrianhopebailie, @ianbjacobs, @rsolomakhin: 

Ok, so given feedback in #27 and teleconferences we've had. How about I do: 

 * Rename: `PaymentOptions.requestPayerAddress` -> `PaymentOptions.requestBillingAddress`

 * Drop: `PaymentRequest.payerAddress` and `PaymentResponse.payerAddress`. 

 * Add "redacted" `billingAddress` to Basic Card's event response: https://github.com/w3c/payment-method-basic-card/pull/53  - which is nulled when `requestBillingAddress` is false. Then, when the user changes instrument, the event fires and you get the `billingAddress` but not the card information. The merchant then can call `.updateWith()`. 

 * Lastly, the merchant always gets the billing address in the `PaymentResponse.details` - which they can request adjustments with via `.retry()` 
 
Agree?

rsolomakhin:
This looks good to me as is.

marcoscaceres:
@rsolomakhin, thanks for the feedback. The issue that come up previously is that billing address is always bound to an instrument, so it might make more sense to just attach the billing address to the "paymentmethodchange" event. 

Basically, the model that is currently specified stays the same... the deck chairs just get rearranged a bit.

marcoscaceres:
Ok, now matches https://github.com/w3c/payment-request/pull/749#issuecomment-412771286 and I've also updated w3c/payment-method-basic-card#53

rsolomakhin:
I do prefer a `paymentmethodchange` event that includes a redacting billing address if necessary. @marcoscaceres  :+1:

marcoscaceres:
@domenic, the updated PR is basically a rewrite (smaller tho:)). Do you have cycles to re-review it?  

@rsolomakhin, do you want to give it a good look over and see if the text matches how you would implement this?

index.html
```diff
@@ -3881,9 +3904,16 @@ <h2>
             </li>
             <li>Set <var>event</var>.<a>[[\waitForUpdate]]</a> to true.
             </li>
+            <li>Let <var>identifer</var> be undefined.
```
rsolomakhin: @marcoscaceres I would prefer to name this `methodIdentifier` to avoid confusion with the Payment Request identifier. `paymentMethodIdentifier` and `pmi` would also be OK.

index.html
```diff
@@ -4555,6 +4586,28 @@ <h2>
...
+                  "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a> <a>
+                    paymentMethodErrors</a> to an IDL value. Otherwise,
+                    <a data-cite=
+                    "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+                    <a>paymentMethodErrors</a> to <a data-cite=
+                    "!WEBIDL#idl-object">object</a>.
+                  </li>
+                  <li>If conversion results in a <a data-cite=
+                  "!WEBIDL#dfn-exception">exception</a> <var>error</var>,
+                  <a>abort the update</a> with <var>error</var>.
```
rsolomakhin, line 4817: @marcoscaceres Does "abort the update" result in abort of payment (i.e., the promise returned byu PaymentRequest.show() is rejected)?

rsolomakhin (approved):
Overall looks good. I have a question and a nitpick suggestion.

index.html
domenic: Also, "identi**fer**" -> "identi**fier**"

index.html
```diff
@@ -4364,9 +4394,10 @@ <h2>
         <p>
           The <dfn>update a <code>PaymentRequest</code>'s details
           algorithm</dfn> takes a <a>PaymentDetailsUpdate</a>
-          <var>detailsPromise</var> and a <a>PaymentRequest</a>
-          <var>request</var>. The steps are conditional on the
-          <var>detailsPromise</var> settling. If <var>detailsPromise</var>
+          <var>detailsPromise</var>, a <a>PaymentRequest</a>
+          <var>request</var>, and optionally a DOMString <var>identifer</var>
+          (a <a>payment method identifier</a>). The steps are conditional on
```
domenic: I don't think it should be optional. It should just either be a string or null. Spec-ese doesn't have the nice JavaScript thing where undefined = missing, so the below is actually broken when the third argument is not passed.

index.html
```diff
@@ -4555,6 +4586,28 @@ <h2>
...
               </li>
+              <li data-link-for="PaymentDetailsUpdate">If the
+              <a>paymentMethodErrors</a> member is present and
+              <var>identifier</var> is not undefined:
+                <ol>
+                  <li>If required by the specification that defines the
+                  <var>identifer</var>, then <a data-cite=
+                  "!WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a> <a>
+                    paymentMethodErrors</a> to an IDL value. Otherwise,
+                    <a data-cite=
```
domenic: It's already an object, so no need to convert.

domenic (commented):
I've lost track of the plot on what feature this implements and how, but at least editorially it seems reasonable, with only a few nits.

I guess my biggest substantive question is why we're adding requestBillingAddress to PaymentOptions when it sounds like we're keeping billing address as a payment method-specific thing. (E.g., it is not returned in any property of PaymentResponse.) If that's the case, shouldn't whether or not to request the billing address be part of the payment method data?

ianbjacobs:
@domenic,

billingAddress should be part of the payment method data. The thing is that in some cases, the Total depends on the billingAddress (e.g., VAT computation). The merchant wants to know about the billingAddress while the sheet is open to display a correct total. 

I believe the solution of this pull request will work for those payment methods build into the browser (e.g., basic card). However it will not work for some payment methods supported by payment handlers (e.g., for push payments or real-time credit financing) until we do something else. I'm synthesizing what I'm hearing here:
 https://github.com/w3c/payment-request/wiki/PushPayments

Ian

index.html
marcoscaceres: Yeah, true that. Will fix.

index.html
marcoscaceres: Ok yeah. I thought it was still a JS object and it was being converted into an IDL object. We might need to check other parts of the spec, as we do the same thing in other places.

index.html
marcoscaceres: Yes.

marcoscaceres:
@domenic wrote: 
> I guess my biggest substantive question is why we're adding requestBillingAddress to PaymentOptions when it sounds like we're keeping billing address as a payment method-specific thing. (E.g., it is not returned in any property of PaymentResponse.) If that's the case, shouldn't whether or not to request the billing address be part of the payment method data?

Yes, you are absolutely right! It should be part of the method data. I'll fix that. 

@ianbjacobs, wrote:
> I believe the solution of this pull request will work for those payment methods build into the browser (e.g., basic card).  However it will not work for some payment methods supported by payment handlers (e.g., for push payments or real-time credit financing) until we do something else.

I really need your help understanding what "something else" could be. I've read over everything provided thus far, and tried to make sure the updated PR addresses the push payment use case (as I understand it). If it doesn't address the use case, then I want to fix that - but I need your help showing me that what I've proposed here is actually unable to address the use case.  

Ian, would it be best to have a call about it? We've successfully resolved issues like this before when we've chatted (or you've successfully explained to me how wrong I am 😅). We only get one real shot at this, so we gotta make sure we get it right.

index.html
marcoscaceres: @rsolomakhin, so I totally agree: problem is we have `methodName` attributes on both  `PaymentResponse` and `PaymentMethodChangeEvent`, both standing for PMI. Changing this one to be `pmi` or `paymentMethodIdentifier` would make this one the odd one out 😢

Although `methodName` is a terrible name, I'd prefer we stay consistent. Is that ok?

marcoscaceres: @rsolomakhin sorry, I misread this out of context! You meant the local variable in the algorithm. Sure, I'll change it.

rsolomakhin:
@marcoscaceres : I think @ianbjacobs wants to make changes to Payment Handler, but this PR should work.

@ianbjacobs : Did I understand you correctly? Please add me to the call, if you have one at a convenient time.

ianbjacobs:
After chatting with @marcoscaceres and @rsolomakhin we would like clarification on one point: if a merchant needs a billing address (e.g., for VAT computation), will that merchant need the billing address for EVERY payment method the merchant would accept?

The answer to this question affects the design, roughly as follows:
 - If the merchant needs a billing address for EVERY payment method, it is easier to indicate that once through the API.
 - Otherwise, the merchant will need to request the billing address on a per payment-method basis. 

As an example, would a merchant ever need billing address for a basic card payment, and not need a billing address for payment by credit transfer, or cryptocurrency, or real-time financing, etc.?

@lyverovski, @Krystosterone, @michelle-stripe, @asolove-stripe, @mattsaxon, @vkuntz, @frank-hoffmann we welcome your input here.

asolove-stripe:
I think there are probably _some_ cases where you only need billing address for one method. I could imagine wanting it for card so you do [AVS on a card](https://en.wikipedia.org/wiki/Address_Verification_System), but don't care about it for other payment methods.

That said, the main use-case here is still where the billing address affects tax or other calculations controlling the total. For that use-case, it's really the same across all payment methods.

stpeter:
Thanks @asolove-stripe! As @adrianhopebailie mentioned on a call right now, if billing address is needed for VAT then it would be needed no matter what the payment method might be. I still worry a bit about non-VAT use cases where a user might be asked for a billing address when that's unexpected for their chosen payment method, but maybe we don't have a good handle on those yet.

marcoscaceres:
Ok, we can go back to having it as a "global" `requestBillingAddress` on `PaymentOptions`. 

In the future, if we need to, we can go more fine grained by adding `requestBillingAddress` on specific  `PaymentMethodData`'s if the case arrises.

stpeter:
This resolution is consistent with discussion on the WPWG call today, too: https://www.w3.org/2018/08/23-wpwg-minutes#item04

marcoscaceres:
Ok, awesome! Thanks for getting wider opinions and review from the WG. Will make these changes hopefully today.

marcoscaceres:
Ok, updated based on recent discussions. This is how it works in theory (it's a bit ugly, because `paymentMethodErrors` 🤷🏽‍♂️):

```JS
const options = {
  requestBillingAddress: true,
};
const request = new PaymentRequest(methods, details, options);

request.onpaymentmethodchange = ev => {
  // get billing address
  const { billingAddress } = ev.methodDetails;
  // get new total
  const updatedTotalPromise = asynCalcNewTotal(billingAddress);
  ev.updateWith(updatedTotalPromise);
};

await validAddresses(response);

async function validAddresses(response) {
  const {
    shippingAddress,
    details: { billingAddress },
  } = response;
  const [shipping, billingErrors] = await Promise.all([
    validateAddress(shippingAddress),
    validateAddress(billingAddress),
  ]);
  const billing = billingErrors
    ? { paymentMethodErrors: { billing: billingErrors } }
    : undefined;
  const errors = { shipping, billing };
  if (Object.getOwnPropertyNames(errors).length) {
    await response.retry(errors);
  }
  return validAddresses(response);
}
```

marcoscaceres:
Again, we do have the opportunity to move `billingAddress` to `PaymentResponse`, and we deprecate "basic-card"'s `.billingAddress`. That would make things a bit cleaner from a developer perspective.

marcoscaceres:
@domenic, @rsolomakhin, opinions on?: https://github.com/w3c/payment-request/pull/749#issuecomment-415663572

marcoscaceres:
fixed a couple of typos above

rsolomakhin:
Thank you, @marcoscaceres!

Re: code sample https://github.com/w3c/payment-request/pull/749#issuecomment-415663572 -- Ideally the merchant would also validate the shipping address in `shippingaddresschange` event and validate the billing address in `paymentmethodchange` event, which happens before the full response is received, so the chance of using `.retry()` would be reduced. (By the way, is `validAddresses` calling itself recursively without a termination condition?)

Re: moving `billingAddress` to `PaymentResponse` and/or `PaymentRequest`, similar to `shippingAddress` https://github.com/w3c/payment-request/pull/749#issuecomment-415664294 -- This would imply that the user agent needs to read the `billingAddress` field from 3rd party payment handler response as well, which is OK, but adds some complexity to keep in mind.

adrianhopebailie:
> Again, we do have the opportunity to move billingAddress to PaymentResponse, and we deprecate "basic-card"'s .billingAddress. That would make things a bit cleaner from a developer perspective.

Yes please!

@rsolomakhin wouldn't it be cleaner to have a `billingaddresschange` event (if we do what @marcoscaceres suggests above)? Seems much less surprising for a developer than different mechanisms for dealing with (to a developer) very similar data elements.

I can imagine implementations then deciding if a payment method change automatically changes the selected billing address or not, and if it does, firing both events. In that case, how would (could) the browser handle two `updateWith` calls?

rsolomakhin:
@adrianhopebailie: It seems that a billing address is always tied to a payment instrument, so decoupling the two events would result in `paymentmethodchange` event always being followed by `billingaddresschange` event, which feels unnecessary. If A and B always happen at the same time, then A==B is my thinking.

adrianhopebailie:
@rsolomakhin I think you are right but I'm not sure if that is ALWAYS true.

What if I pick a payment handler that has multiple underlying instruments? If we de-couple this we open up the possibility of allowing payment handlers to explicitly fire this event.

e.g. I pick BobPay which has two different instruments internally.
- Browser fires `paymentmethodchanged` event to signal the selection of BobPay
(Sidebar: Does this event still need an `updateWith()` if it doesn't signal address changes?)
- Inside the BobPay payment handler I select instrument 1
- Payment Handler calls a new synchronous `billingAddressChanged(address)` method
- Browser fires `billingaddresschanged` event
- Website calls `updateWith` with updated VAT and total
- Payment handler gets new payment details as response to calling `billingAddressChanged(address)`
- Inside the BobPay payment handler I select instrument 2
- Payment Handler calls `billingAddressChanged(address)` method again
- Browser fires `billingaddresschanged` event
- Website calls `updateWith` with updated VAT and total
- Payment handler gets new payment details as response to calling `billingAddressChanged(address)`
- etc

If the payment handler never calls 'billingAddressChanged()' then the browser can still fire the `billingaddresschanged` event when the payment handler returns (but before passing the response to the website). In this case, if the website returns with a new total the browser knows that the payment needs to be retried with the payment handler (some design work to figure out how that would be dealt with).

We could consider s similar mechanism for shipping address

marcoscaceres:
> Browser fires paymentmethodchanged event to signal the selection of BobPay.

Currently, "paymentmethodchanged"  only fires when the user picks an instrument, not when they pick a different or new handler.  

> Sidebar: Does this event still need an updateWith() if it doesn't signal address changes?

I would say, yes, for consistency - but only in as far as the event might contain useful information that the merchant would modify the total on (or modify some other aspect of the request on).

marcoscaceres:
> Re: code sample #749 (comment) -- Ideally the merchant would also validate the shipping address in shippingaddresschange event and validate the billing address in paymentmethodchange event, which happens before the full response is received, so the chance of using .retry() would be reduced.

Agree. 

> (By the way, is validAddresses calling itself recursively without a termination condition?)

oops! will fix. 

> Re: moving billingAddress to PaymentResponse and/or PaymentRequest, similar to shippingAddress #749 (comment) -- This would imply that the user agent needs to read the billingAddress field from 3rd party payment handler response as well, which is OK, but adds some complexity to keep in mind.

I'll just leave it as is.

rsolomakhin:
I've hooked this up into 3rd party payment handlers via https://github.com/w3c/payment-handler/pull/318.

marcoscaceres:
@sonakshisaxena1 has graciously offered to assist us with the testing. I'll work with her on it.

marcoscaceres:
@rsolomakhin @aestes, @zouhir just wanted to confirm if you are also planning to support this (`requestBillingAddress`)? 

Firefox's plan is always return `null` for billing address unless this is set to true.

rsolomakhin:
Yes, we plan to support this in Chrome, but the exact behavior needs to be thought through and fine-tuned on our end. The Firefox decision seems reasonable is in the realm of possibilities for us.

snez:
I am currently using Stripe's javascript API with the paymentRequestButton, and I have found the billing address to be available in the returned json object when the customer selects a card. I.e. the billing address registered with the specific card comes back as part of the card details json object. Not sure if the underlying implementation is something that Stripe has done, or if it is part of the paymentRequest API, but it solves the problem in my case.

marcoscaceres:
Hi @snez, you wrote:

> Not sure if the underlying implementation is something that Stripe has done, or if it is part of the paymentRequest API, but it solves the problem in my case.

If I may ask, could you help us understand what you are using the billing address for?

marcoscaceres:
Put up a test for review: https://github.com/web-platform-tests/wpt/pull/14255

marcoscaceres:
Just a quick update: Firefox patches for `requestBillingAddress` should be landing hopefully tomorrow.

marcoscaceres:
File [chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=909634), based on @rsolomakhin comment https://github.com/w3c/payment-request/pull/749#issuecomment-442093344

snez:
> Hi @snez, you wrote:
> 
> > Not sure if the underlying implementation is something that Stripe has done, or if it is part of the paymentRequest API, but it solves the problem in my case.
> 
> If I may ask, could you help us understand what you are using the billing address for?

The billing address is a minimum requirement for many large eCommerce systems in order to place an order (in my case Magento). I wouldn't be able to programmatically create an order without it, and thus the whole idea of accepting payments from the product page could not be implemented (see https://magento2.cryozonic.com/radiant-tee.html for an example), which is a perfect use for the paymentRequest API, and something that many merchants are trying to implement to improve their conversions.

marcoscaceres:
Thanks, @snez, for the example and details. That’s super helpful.

PR #806, opened 2018-11-05
Change what canMakePayment() means

### LINKED ISSUE #800: Meaning of canMakePayment()
We discovered during testing that each implementation have slightly different definitions of "can make payment". During the F2F, we came to a consensus that "canMakePayment" means "the user agent supports the PMIs" - but it does not mean that there is a payment instrument set up. We concluded that we might add something to check if there is an active instrument in the future. I'll file a different bug for that.

marcoscaceres:
During the F2F, we came to a consensus that "canMakePayment" means "the user agent supports the PMIs" - but it does not mean that there is a payment instrument set up. We concluded that we might add something to check if there is an active instrument in the future.

closes #800 

 * [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/13912)
 * [x] [Modified MDN Docs](https://developer.mozilla.org/en-US/docs/Web/API/PaymentRequest/canMakePayment) (link)

Implementation commitment:

 * [x] [Safari](https://trac.webkit.org/r237594)
 * [ ] Chrome (link to issue)
 * [x] Firefox - "Born this way" 👩‍🎤. 
 * [ ] Edge (public signal)

@ianbjacobs and @rsolomakhin, probably some impact! ☄️

### DIFF
--- a/index.html
+++ b/index.html
@@ -1216,10 +1216,10 @@
               time period in which those calls were made.
             </p>
           </li>
-          <li>Let <var>promise</var> be <a>a new promise</a>.
+          <li>Let <var>hasHandlerPromise</var> be <a>a new promise</a>.
           </li>
-          <li>Return <var>promise</var>, and perform the remaining steps <a>in
-          parallel</a>.
+          <li>Return <var>hasHandlerPromise</var>, and perform the remaining
+          steps <a>in parallel</a>.
           </li>
           <li>For each <var>paymentMethod</var> tuple in
           <var>request</var>.<a>[[\serializedMethodData]]</a>:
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Redid the test and updated MDN.

rsolomakhin:
How far in the future are we thinking about the active instrument check?

marcoscaceres:
I’m unsure - that was really unclear to me. Do we want it now? Can we live without it for V1?

rsolomakhin:
I would prefer to implement it soon, but don't want to force the other implementers or block the CR process.

marcoscaceres:
I guess we could at least draft it up and put everything in place, then check implementation plans.

rsolomakhin:
Good idea :+1:

marcoscaceres:
We couldn’t implement in Firefox for at least 3 month. We have a big backlog to finish just with the current stuff. That means 6 months till we would have something in release.

marcoscaceres:
Unrelated but, Mozilla folks put together this great wiki page that shows our implementation priorities, with explanations, and stuff we still need to do (@ianbjacobs and Chairs, you might like it): 

https://wiki.mozilla.org/Firefox/Features/Web_Payments/DOM

It gives a good sense of where we are at and updates dynamically as we fix things, so good to periodically check.

domenic (approved):

PR #810, opened 2018-11-20
Clarify when the user can abort the payment request algorithm

### LINKED ISSUE #796: Clarify when the user can abort the payment request algorithm
The [user aborts the payment request algorithm](https://w3c.github.io/payment-request/#user-aborts-the-payment-request-algorithm) says this:

> If the request.[[updating]] is true, then terminate this algorithm and take no further action. The user agent user interface SHOULD ensure that this never occurs.

I believe this is because user agents are expected to disable the payment UI while merchants are updating details.

In Apple Pay, we do prevent the user from authorizing payment while the merchant is updating, but we don't completely disable the interface. Notably, we don't disable the "cancel" button. So that means the user could abort a transaction in the midst of a details update.

To comply with the spec as-written, WebKit defers processing a user abort until the pending details update promise settles (see https://trac.webkit.org/r236922).

Should the spec acknowledge that some implementations might allow the user to abort in the midst of updating? Should it allow an abort during details update, or suggest that implementors defer the abort until the details settle?

cc: @marcoscaceres

marcoscaceres:
closes #796

Implementation commitment:

 * [ ] Safari (link to issue)
 * [ ] Chrome (link to issue)
 * [ ] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1510019)
 * [ ] Edge (public signal)

None

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/810.html" title="Last updated on Jan 10, 2019, 4:01 AM UTC (d5141c1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/810/3148aa3...d5141c1.html" title="Last updated on Jan 10, 2019, 4:01 AM UTC (d5141c1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -887,7 +887,7 @@
             The <a>show()</a> method is called when a developer wants to begin
             user interaction for the payment request. The <a>show()</a> method
             returns a <a>Promise</a> that will be resolved when the <a>user
-            accepts the payment request</a>, or rejected if the <a>user aborts
+            accepts the payment request</a> or rejected if the <a>user aborts
             the payment request</a>. Some kind of user interface will be
             presented to the user to facilitate the payment request after the
             <a>show()</a> method returns.
--- a/index.html
+++ b/index.html
@@ -887,10 +887,9 @@
             The <a>show()</a> method is called when a developer wants to begin
             user interaction for the payment request. The <a>show()</a> method
             returns a <a>Promise</a> that will be resolved when the <a>user
-            accepts the payment request</a> or rejected if the <a>user aborts
-            the payment request</a>. Some kind of user interface will be
-            presented to the user to facilitate the payment request after the
-            <a>show()</a> method returns.
+            accepts the payment request</a>. Some kind of user interface will
+            be presented to the user to facilitate the payment request after
+            the <a>show()</a> method returns.
           </p>
           <p>
             It is not possible to show multiple <a>PaymentRequest</a>s at the
--- a/index.html
+++ b/index.html
@@ -887,9 +887,10 @@
             The <a>show()</a> method is called when a developer wants to begin
             user interaction for the payment request. The <a>show()</a> method
             returns a <a>Promise</a> that will be resolved when the <a>user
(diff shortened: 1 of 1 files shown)


marcoscaceres:
So, we had in there that there could be a timeout or some other means of aborting. However, I think we were going down the wrong path... I've routed the "timeout" and the user hitting "esc" down the "[user aborts the payment request](https://www.w3.org/TR/payment-request/#user-aborts-the-payment-request-algorithm)" path instead. I also removed the "[[updating]]" check to allow for this.

marcoscaceres:
@domenic, if you have time, can you please sanity check this?

index.html
```diff
@@ -887,9 +887,10 @@ <h2>
             The <a>show()</a> method is called when a developer wants to begin
             user interaction for the payment request. The <a>show()</a> method
             returns a <a>Promise</a> that will be resolved when the <a>user
-            accepts the payment request</a>. Some kind of user interface will
-            be presented to the user to facilitate the payment request after
-            the <a>show()</a> method returns.
+            accepts the payment request</a> or rejected if the <a>user aborts
+            the payment request</a>. Some kind of user interface will be
```
domenic: There's a lot of failure conditions, and listing only one of them might be a bit more confusing than helpful.

domenic (approved):
The duplication between "user aborts the payment request" and "abort the update" made this hard to review, but in the end I think the switch to that algorithm is fine. We should consider deduplicating those two algorithms though, if we can.

index.html
marcoscaceres: Agree. Reverting.

marcoscaceres:
Filed bug about de-dupe abort the update and user aborts. https://github.com/w3c/payment-request/issues/814

marcoscaceres:
As this makes provisions for an optional UI feature (a "cancel" button or user triggered action, like a hitting "esc"), I'm ok with merging this as is. It doesn't really add any new conformance requirements. It just clarifies behavior that matches what browsers already do.

PR #811, opened 2018-11-21
Attach 'payment request is showing boolean' to top-level browsing context

### LINKED ISSUE #809: Rethink "payment request is showing" boolean
I think we should reconsider how "payment request is showing" boolean works and perhaps leave it to the user agent (or the payment handler, rather). 

For example, in Firefox, for "basic-card", there is no real reason why one shouldn't be able to have multiple payment sheets up for different tabs/windows at the same time. 

However, for payment handlers like Apple Pay, where authentication is required either via an Apple Watch or  an iPhone, it makes sense to only allow one payment sheet at a time (because there is a strong tie between the authentication device and the payment sheet being shown).  

Thus, I think what the spec currently states is incorrect: right now, if Sita A is showing a payment sheet, and Site B tries to show a payment sheet, Site B's payment request gets rejected.  

What I think we want is: if required by the Payment Handler, if Sita A is showing a payment sheet, and Site B tries to show a payment sheet, then Site A's payment gets aborted. 

cc @aestes, @rsolomakhin, @zkoch, @mnoorenberghe...

marcoscaceres:
closes #809

 * [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/14297)

Implementation commitment:

 * [ ] Safari (link to issue)
 * [ ] [Chrome](https://crbug.com/907509)
 * [ ] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1469419)
 * [ ] Edge (public signal)

Yes. Payment Handlers might need to have a means of keeping track of open payment sheets.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/811.html" title="Last updated on Jan 16, 2019, 7:06 AM UTC (ae602e9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/811/3148aa3...ae602e9.html" title="Last updated on Jan 16, 2019, 7:06 AM UTC (ae602e9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -610,9 +610,20 @@
         </p>
       </div>
       <p>
-        Each <a>top-level browsing context</a> has a <dfn>payment request is
-        showing</dfn> boolean, which prevents showing more than one payment UI
-        at a time.
+        A <var>request</var>'s <dfn>payment-relevant browsing context</dfn> is
+        that <a>PaymentRequest</a>'s <a data-cite=
+        "HTML#concept-relevant-global">relevant global object</a>'s browsing
+        context's <a>top-level browsing context</a>. Every <a>payment-relevant
+        browsing context</a> has a <dfn>payment request is showing</dfn>
+        boolean, which prevents showing more than one payment UI at a time.
+      </p>
+      <p class="Note">
+        A payment handlers can restrict the <a>user agent</a> to showing only
+        one payment UI across all browser windows and tabs. Other payment
+        handlers might allow showing a payment UI across disparate browser
+        tabs. However, in all cases, the <a>payment request is showing</a>
+        boolean simply prevents more than one payment UI being shown in a
+        single browser tab.
       </p>
       <section>
         <h2>
@@ -934,9 +945,10 @@
           then return <a>a promise rejected with</a> an
           "<a>InvalidStateError</a>" <a>DOMException</a>.
           </li>
-          <li>If the <a>top-level browsing context</a>'s <a>payment request is
-          showing</a> boolean is true, then return <a>a promise rejected
-          with</a> an "<a>AbortError</a>" <a>DOMException</a>.
+          <li>If <var>request</var>'s <a>payment-relevant browsing
+          context</a>'s <a>payment request is showing</a> boolean is true, then
+          return <a>a promise rejected with</a> an "<a>AbortError</a>"
+          <a>DOMException</a>.
           </li>
           <li>Set <var>request</var>.<a>[[\state]]</a> to "<a>interactive</a>".
           </li>
@@ -966,8 +978,8 @@
               advantage of this step.
             </p>
           </li>
-          <li>Set the <a>top-level browsing context</a>'s <a>payment request is
-          showing</a> boolean to true.
+          <li>Set <var>request</var>'s <a>payment-relevant browsing
+          context</a>'s <a>payment request is showing</a> boolean to true.
           </li>
           <li>Return <var>acceptPromise</var> and perform the remaining steps
           <a>in parallel</a>.
@@ -1000,8 +1012,9 @@
                   </li>
                   <li>Reject <var>acceptPromise</var> with <var>error</var>.
                   </li>
-                  <li>Set <a>top-level browsing context</a>'s <a>payment
-                  request is showing</a> boolean to false.
+                  <li>Set <var>request</var>'s <a>payment-relevant browsing
+                  context</a>'s <a>payment request is showing</a> boolean to
+                  false.
                   </li>
                   <li>Terminate this algorithm.
                   </li>
@@ -1031,8 +1044,8 @@
               <li>Reject <var>acceptPromise</var> with
               "<a>NotSupportedError</a>" <a>DOMException</a>.
               </li>
-              <li>Set <a>top-level browsing context</a>'s <a>payment request is
-              showing</a> boolean to false.
+              <li>Set <var>request</var>'s <a>payment-relevant browsing
+              context</a>'s <a>payment request is showing</a> boolean to false.
               </li>
               <li>Terminate this algorithm.
               </li>
@@ -1094,8 +1107,8 @@
             <ol>
               <li>Close down the user interface.
               </li>
-              <li>Set the <a>top-level browsing context</a>'s <a>payment
-              request is showing</a> boolean to false.
+              <li>Set <var>request</var>'s <a>payment-relevant browsing
+              context</a>'s <a>payment request is showing</a> boolean to false.
               </li>
               <li>Reject <var>acceptPromise</var> with an "<a>AbortError</a>"
               <a>DOMException</a>.
@@ -3169,8 +3182,8 @@
               "WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a> to
               <a data-cite="WEBIDL#idl-object">object</a>.
               </li>
-              <li>Set <a>top-level browsing context</a>'s <a>payment request is
-              showing</a> boolean to false.
+              <li>Set <var>request</var>'s <a>payment-relevant browsing
+              context</a>'s <a>payment request is showing</a> boolean to false.
               </li>
               <li>If conversion results in a <a data-cite=
               "WEBIDL#dfn-exception">exception</a> <var>error</var>:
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@ianbjacobs, you want to give this first read? 

@rsolomakhin, we will need to consider more fully what it means for Payment Handlers. I was a bit unsure if I should file a separate bug for Handlers itself.

rsolomakhin (commented):
A payment handler is able to enumerate all of its own windows through the [Clients](https://developer.mozilla.org/en-US/docs/Web/API/Clients) API in service worker, so it can take the appropriate actions inside of the `'paymentrequest'` event to either abort the current payment, abort other ongoing payments, or allow multiple ongoing payments. We should not need to add anything else to the payment handler spec for now. Does that make sense?

marcoscaceres:
Perfect! Thanks @rsolomakhin! 

Next big question: willing to change Chrome to match Firefox? :)

rsolomakhin:
Yes, sir! From what I understand, you're proposing that each browser tab have ability to show its own payment sheet. I would be willing to enable this in Chrome.

marcoscaceres:
Awesome! Ok, cool. I think we have all we need to move forward on this 🥳 

I’ll get a review from Domenic, just to make sure I didn’t miss anything.

marcoscaceres:
@rsolomakhin when possible, could you provide a link to a Chrome bug. No hurry, I know you all are on vacation next few days.

rsolomakhin:
Linked in description.

index.html
```diff
@@ -935,9 +934,9 @@ <h2>
           then return <a>a promise rejected with</a> an
           "<a>InvalidStateError</a>" <a>DOMException</a>.
           </li>
-          <li>If the <a>user agent</a>'s <a>payment request is showing</a>
-          boolean is true, then return <a>a promise rejected with</a> an
-          "<a>AbortError</a>" <a>DOMException</a>.
+          <li>If the <a>top-level browsing context</a>'s <a>payment request is
```
domenic: "the top-level browsing context" is not well-defined. Consider a same-origin popup window (i.e. two separate top-level windows which can synchronously access each other). Then call `window1.PaymentRequest.prototype.show.call(window2PaymentRequestInstance)`. Which top-level BC is this algorithm talking about in that case? The popup window or the original window?

You want "this `PaymentRequest`'s [relevant global object](https://html.spec.whatwg.org/#concept-relevant-global)'s browsing context's top-level browsing context". Maybe make a definition to shorten that (e.g. "payment-relevant browsing context of _obj_" or "payment request is showing in the same browsing context as _obj_").

And, if you're feeling evil, write tests for the case I described ^_^.

domenic (changes requested):
Needs more precision. "The top-level browsing context" is never correct; there are many of them per UA, which is why we're making this change, after all. So pick specific JS objects (presumably the PaymentRequest/PaymentResponse instances) and navigate from them to top-level BCs.

index.html
marcoscaceres: Thanks! I had a bad feeling about "top-level browsing context"... the suggestion is great. I'll definitely add the test 😼.

index.html
```diff
@@ -609,12 +609,10 @@ <h2>
...
-      <p data-link-for="PaymentRequest">
-        Because the simultaneous display of multiple <a>PaymentRequest</a> user
-        interfaces might confuse the user, this specification limits the
-        <a>user agent</a> to displaying one at a time via the <a>show()</a>
-        method. This is ensured by a <dfn>payment request is showing</dfn>
-        boolean.
+      <p>
+        Each <a>top-level browsing context</a> has a <dfn>payment request is
+        showing</dfn> boolean, which prevents showing more than one payment UI
+        at a time.
```
ianbjacobs: Hi @marcoscaceres,

Editorial suggestion for the above 

        ....showing</dfn> boolean to support <a>user agent</a> management of payment UI display. For example,
       a <a>user agent</a> might choose to show only one payment UI at a time.

marcoscaceres: Appreciate this suggestion, but it doesn't quite capture what we are trying to restrict here (don't show two payment UI in the same browser tab). I'll add a note describing what you are saying above.

index.html
marcoscaceres: > `window1.PaymentRequest.prototype.show.call(window2PaymentRequestInstance)` 

@domenic, just to make sure I got this right. 

Both of these would theoretically show the payment sheet in the popup: 

```JS
button.onclick = () => {
    const popup = window.open("some.html");
    await new Promise(resolve => (popup.onload = resolve));
    const r = new popup.PaymentRequest(methods, details);
    await r.show();
}
```
And: 

```JS
button.onclick = () => {
    const popup = window.open("some.html");
    await new Promise(resolve => (popup.onload = resolve));
    const r = new popup.PaymentRequest(methods, details);
    await PaymentRequest.prototype.show.call(r);
}
```

Is that correct? .... 

And then could go from `popup.opener`, and show a payment sheet in the opener from the popup.  

I've put up an initial set that use just an iframe:
https://github.com/w3c/payment-request/pull/811

marcoscaceres:
Initial test set: https://github.com/web-platform-tests/wpt/pull/14297

index.html
marcoscaceres: Err, sorry, copypasta, tests are here  https://github.com/web-platform-tests/wpt/pull/14297
(I've not run them - so probably buggy)

domenic (approved):

index.html
domenic: > Both of these would theoretically show the payment sheet in the popup:

Yeah, that's right!

> And then could go from popup.opener, and show a payment sheet in the opener from the popup.

I don't quite understand what you mean here.

> I've put up an initial set that use just an iframe:

I'll take a look, but iframes are not as good a test here, as they aren't top-level browsing contexts. The payment-relevant browsing context of a PaymentRequest created inside an iframe is its parent browsing context. Popups are special because they are top-level; they don't have a parent, just an opener.

marcoscaceres:
Sent an updated batch of tests for this pull request. Hope to merge this soon.

index.html
marcoscaceres: > And then could go from popup.opener, and show a payment sheet in the opener from the popup.

@domenic, imagine the following situation:

```JS
// this is actually a new browser tab...
const popupTab = window.open("tab.html");

// in which tab does the sheet show? In the popup tab or in opener tab? 
await new popupTab.PaymentRequest(method, details).show();
```

I guess "implementation detail"... 

Similarly, consider this one (little window):

```JS
// We make a little window... imagine, without toolbars, etc.
const popupWindow = window.open("window.html", "width=100,height=100,...");

// in which window does the sheet show? In the little window or in opener? 
await new popupWindow.PaymentRequest(method, details).show();
```

Similarly, I guess "implementation detail"?... don't think we can do much in the spec, but just wanted to point this out.

rsolomakhin:
This still looks good to me. Do you need anything else from us?

marcoscaceres:
@rsolomakhin no, it's all good from the spec side (appreciate you rechecking!). It was mostly on the test side of things. I was worried because I couldn't get the tests to run in Chrome... will follow up on the WPT side.

index.html
```diff
@@ -609,12 +609,21 @@ <h2>
...
-        interfaces might confuse the user, this specification limits the
-        <a>user agent</a> to displaying one at a time via the <a>show()</a>
-        method. This is ensured by a <dfn>payment request is showing</dfn>
-        boolean.
+      <p>
+        A <var>request</var>'s <dfn>payment-relevant browsing context</dfn> is
+        that <a>PaymentRequest</a>'s <a data-cite=
+        "HTML#concept-relevant-global">relevant global object</a>'s browsing
+        context's <a>top-level browsing context</a>. Every <a>payment-relevant
+        browsing context</a> has a <dfn>payment request is showing</dfn>
```
ianbjacobs, line 617: Hi @marcoscaceres,

The above sentence is quite dense. Is it correct? Should it be:

<pre>
       A <var>request</var>'s <dfn>payment-relevant browsing context</dfn> is
       that <a>PaymentRequest</a>'s <a data-cite=
        "HTML#concept-relevant-global">relevant global object</a>'s <a>top-level browsing context</a>.
</pre>

Or even better:

<pre>
    The <dfn>payment-relevant browsing context</dfn> of a <a>PaymentRequest</a>
    is the <a>top-level browsing context</a> of the <a data-cite=
        "HTML#concept-relevant-global">relevant global object</a>.
</pre>

marcoscaceres: Prefer we leave this for now, tbh. It's convoluted, but accurate.

Issue #753, opened 2018-07-12
Modifiers and [[serializedModifierData]] not actually used anywhere.


marcoscaceres:
Following on from #684, there are a bunch of huge issues in the spec relating to modifiers:

 * [[serializedModifierData]] is not actually used anywhere - it's undefined how that data actually gets applied/displayed.
 * [[serializedModifierData]] gathers all the data from EVERY modifier (without matching on PMI!), then passes ALL data to the chosen payment handler - in the spec, there is no way to distinguish between data for ApplePay and data for Basic-Card.  
 * The modifiers in `serializedModifierData` are never converted, only serialized, meaning that if they are  converted by the payment handler, they can throw: This is undefined behavior.  
 * The actual serialization also seems pointless. If that's for IPC, then that should be an implementation detail. It doesn't seem to serve any purpose in the spec (because [[serializedModifierData]] is never actually used for anything). 

I'm again strongly inclined towards dropping modifiers in favor of something else. They are super broken, underspecified, and unnecessarily complicated.

marcoscaceres:
Err... that should have been "serializedModifierData"

domenic:
FWIW the serialization was evolved from an earlier revision that checked if the objects were "JSON serializable". Assuming that's still a desirable property in the abstract, I remain convinced that serializing and passing to the payment handler spec is the right way to spec that property.

I'm less informed about the general question as to whether modifiers are worth keeping.

marcoscaceres:
> I remain convinced that serializing and passing to the payment handler spec is the right way to spec that property.

I don't mind so much (specially if a payment handler `JSON.parse()`s on the other end) - but we still have the problem that you get the wrong data. That's fixable by using  `tuple(PMI, serializedData)` - as it's done with `[[serializedMethodData]]`.

But the larger question of conversion kinda remains, as does `[[serializedModifierData]]` not actually being applied.

marcoscaceres:
I’ll send a patch to address some of the above.

stpeter:
Receiving a jumble of all the data for all the payment methods is bad enough, but I'd agree that there are even bigger problems with modifiers - I wonder if the feature can even be implemented reliably, let alone interoperably...

marcoscaceres:
From my testing today, Chrome Desktop does not seem to apply modifiers correctly. It seems to think a debit card is a credit card, and thus the modifiers applies the wrong total.

@zkoch, is this a known issue?

marcoscaceres:
[...]

PR #812, opened 2018-11-23
fix: convert PaymentMethodData.data

### LINKED ISSUE #813: Should PaymentMethodData be converted at construction time?
Related to #753 

Given the following: 

```JS
var details = {
  total: {
    label: "Total",
    amount: {
      currency: "USD",
      value: "0.00",
    },
  },
};
new PaymentRequest(
  [
    { supportedMethods: "basic-card", data: { supportedNetworks: "☠️" } },
  ],
  details
);
```

Firefox currently throws, because it tries to do the IDL conversion: 

 > TypeError: Fail to convert methodData.data to BasicCardRequest. 

Chrome doesn't throw. 

Per spec, Chrome is right... but it not doing the conversion seems not great.

marcoscaceres:
Closes #813 

Part 1 of  #753 

Converts PaymentMethodData.data to IDL type or object. 
 
 * [X] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/14205)
 * [X] Modified MDN Docs - obscure detail, not required. 

Implementation commitment:

 * [ ] Safari 
 * [x] Chrome - Implemented 
 * [x] Firefox - Implemented 
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/812.html" title="Last updated on Jan 21, 2019, 5:50 AM UTC (e473522)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/812/5459933...e473522.html" title="Last updated on Jan 21, 2019, 5:50 AM UTC (e473522)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -663,7 +663,8 @@
               </li>
             </ol>
           </li>
-          <li>Let <var>serializedMethodData</var> be an empty list.
+          <li>Let <var>paymentMethods</var> be an empty
+          sequence&lt;<a>PaymentMethodData</a>&gt;.
           </li>
           <li>Process payment methods:
             <ol data-link-for="PaymentMethodData">
@@ -678,24 +679,31 @@
                   to <a data-cite=
                   "payment-method-id#dfn-validate-a-payment-method-identifier">
                     validate a payment method identifier</a> with
-                    <var>paymentMethod</var>.<a data-lt=
-                    "PaymentMethodData.supportedMethods">supportedMethods</a>.
-                    If it returns false, then throw a <a>RangeError</a>
-                    exception. Optionally, inform the developer that the
-                    payment method identifier is invalid.
+                    <var>paymentMethod</var>.<a>supportedMethods</a>. If it
+                    returns false, then throw a <a>RangeError</a> exception.
+                    Optionally, inform the developer that the payment method
+                    identifier is invalid.
                   </li>
-                  <li>If the <a data-lt="PaymentMethodData.data">data</a>
-                  member of <var>paymentMethod</var> is missing, let
-                  <var>serializedData</var> be null. Otherwise, let
-                  <var>serializedData</var> be the result of
-                  <a>JSON-serializing</a>
-                    <var>paymentMethod</var>.<a data-lt="PaymentMethodData.data">data</a>
-                    into a string. Rethrow any exceptions.
+                  <li>If <var>paymentMethod</var>.<a>data</a> is present:
+                    <ol>
+                      <li>
+                        <a>JSON-serialize</a>
+                        <var>paymentMethod</var>.<a>data</a>. Rethrow any
+                        exceptions.
+                      </li>
+                      <li>If required by the specification that defines the
+                      <var>paymentMethod</var>.<a>supportedMethods</a>, replace
+                      <var>paymentMethod</var>.<a>data</a> with a
+                      <a>converted</a> IDL value of the type specified there
+                      (e.g., a <a data-cite=
+                      "payment-method-basic-card#dom-basiccardrequest">BasicCardRequest</a>
+                      in the case of [[?payment-method-basic-card]]). Rethrow
+                      any exceptions.
+                      </li>
+                    </ol>
                   </li>
-                  <li>Add the tuple (<var>paymentMethod</var>.<a data-lt=
-                  "PaymentMethodData.supportedMethods">supportedMethods</a>,
-                  <var>serializedData</var>) to
-                  <var>serializedMethodData</var>.
+                  <li>Append <var>paymentMethod</var> to
+                  <var>paymentMethods</var>.
                   </li>
                 </ol>
               </li>
@@ -770,8 +778,6 @@
               </li>
             </ol>
           </li>
-          <li>Let <var>serializedModifierData</var> be an empty list.
-          </li>
           <li data-link-for="PaymentDetailsBase">Process payment details
           modifiers:
             <ol>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Both Chrome and Firefox do the IDL conversion already.

index.html
```diff
@@ -805,11 +803,30 @@ <h2>
...
+                      "WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+                      <var>data</var> to an IDL value of the type specified
+                      there. Otherwise, <a data-cite=
+                      "WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+                      to <a data-cite="WEBIDL#idl-object">object</a>. Rethrow
+                      any exceptions.
+                      </li>
+                      <li>If the user agent does not have an available
+                      <a>payment handler</a> for <var>identifer</var>,
+                      continue.
```
marcoscaceres: We probably don't want to collect the unknown ones - only to convert them to retain current implementation behavior.

index.html
```diff
@@ -805,12 +803,34 @@ <h2>
...
+                      <a data-lt=
+                      "PaymentDetailsModifier.supportedMethods">supportedMethods</a>
+                      member of <var>modifier</var>.
+                      </li>
+                      <li data-tests=
+                      "constructor_convert_method_data.https.html">If required
+                      by the specification that defines the
+                      <var>identifier</var>, then <a data-cite=
+                      "WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+                      <var>data</var> to an IDL value of the type specified
```
domenic: What is _data_? I cannot find its declaration.

domenic (changes requested):

index.html
marcoscaceres: Oops, data was supposed to be `PaymentDetailsModifier.data`. Fixing.

marcoscaceres:
@domenic hopefully good now.

marcoscaceres:
@domenic, because in implementations, `PaymentMethod.data`'s JS-to-IDL conversion happens in the `PaymentRequest` constructor (instead of on `.show()`), it means we don't need the `[[serialized*]]` data internal slots because:

 * We can do the "JSON serializable" check at construction time and when `updateWith()` is called: this gives us assurances that `.data` doesn't contain any un-serializable JSON types and circular references....and, if ever needed, `.data` can be serialized to a string in the future. 
 * And because of the above, we can just internally hold *converted* IDL values for each PaymentRequest instance, without the need to hold them in string form.

This  also fixes what would otherwise be a mess when `show()`ing or `updateWith()`ing a payment sheet: the [[serializedModifierData]] array positions would need  to match the position of the deleted `.data` members. 

So, with this PR,  when a payment handler shows or gets updated, it always operates on its expected IDL type, without needing to redo any redundant JSON parsing, re-conversions, etc.

index.html
```diff
@@ -678,24 +679,31 @@ <h2>
...
+                      any exceptions.
+                      </li>
+                    </ol>
                   </li>
-                  <li>Add the tuple (<var>paymentMethod</var>.<a data-lt=
-                  "PaymentMethodData.supportedMethods">supportedMethods</a>,
-                  <var>serializedData</var>) to
-                  <var>serializedMethodData</var>.
+                  <li>Append <var>paymentMethod</var> to
+                  <var>paymentMethods</var>.
```
domenic, line 706: So the problem with not storing the JSON serialized version is that the JSON serialization now becomes meaningless. Consider the following:

```js
const bcr1 = { supportedNetworks: new Set("foo", "bar") };
```

With the new spec, the JSON-serialization will be `{ "supportedNetworks": {} }`. But the value stored in _paymentMethods_ will be `{ supportedNetworks: ["foo", "bar"] }` since Web IDL dictionary conversion converts all iterables to sequences.

This means that if someone, e.g., passed the JSON serialization across a process boundary and used it to attempt to reconstruct the result on the other side, they would not be able to match the spec. I thought our original spec was motivated exactly by that design.

domenic (changes requested):
Can you separate the editorial changes (removing lots of `data-lt`s) from the normative ones, to make this easier to review? I think I found the big normative change and the issue with it, but it's hard to tell what else might be there.

index.html
marcoscaceres: I believe the original design was motivated by some kind of security concern that was supposed to be mitigated by doing a JSON serialization on `.data` whose type is `object` to remove things like `Function` being added as properties.

domenic: It may be good to check with implementations then. But a spec that JSON serializes then disregards the result seems incoherent.

marcoscaceres: Gecko only does the IDL conversion, because we don't care about any payment method apart from "Basic Card". So, we don't store any `[[serializedMethodData]]` in our code... in fact, I don't see is doing any we JSON serialization of data at all in Gecko's code (@edenchuang?).

From the constructor, we check the method's details:
https://github.com/mozilla/gecko-dev/blob/master/dom/payments/PaymentRequest.cpp#L574  

We then just check `IsValidBasicCardRequest`: 
https://github.com/mozilla/gecko-dev/blob/master/dom/payments/PaymentRequest.cpp#L293

And then we convert it (to check throw if it fails):
https://github.com/mozilla/gecko-dev/blob/master/dom/payments/BasicCardPayment.cpp#L53

[WebKit does do the serialization](https://github.com/WebKit/webkit/blob/master/Source/WebCore/Modules/paymentrequest/PaymentRequest.cpp#L322) and stores it. However, @aestes also seems to suggest a [preference to not store the JSON serialization](https://github.com/w3c/payment-request/issues/813#issuecomment-441886747): "Browsers could convert from a JS object directly to a IDL dictionary rather than the current JS object > JSON string > JS object > IDL dictionary conversion algorithm".

domenic: Thanks. @rsolomakhin?

marcoscaceres:
Went back to look at https://github.com/w3c/payment-request/issues/346 ... I'd forgotten that the ideas was to support actually sending the JSON-serialized data no just across process boundaries (which I was imagining was just as per-origin process isolation) - but to other native applications acting as payment handlers. 

As such, I concur that it still makes sense to keep the JSON serialization.

PR #821, opened 2019-01-11
fix: close request if sheet is showing

### LINKED ISSUE #820: Clarification question for `show()` algorithm
**TL;DR:** Can `show()` be called on the same instance of `PaymentRequest` multiple times?

The `show()` algorithm includes the following steps:

> 6. If *request*.[[state]] is not "created" then return a promise rejected with an "InvalidStateError" DOMException.
> 7. If the user agent's "payment request is showing" boolean is true, then return a promise rejected with an "AbortError" DOMException.
> 8. Set *request*.[[state]] to "interactive".

It looks like `show()` can be rejected without altering the *request*.[[state]], so the merchant can call `show()` on the same instance of `PaymentRequest`. Is this the only edge case where `show()` does not change state? Or am I reading the spec wrong?

marcoscaceres:
closes #820

 * [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/commit/5f1cbd0a42811a6c66bcfdf25f8a5e7023bf900e)
 * [x] Modified MDN Docs - not needed. 
 * [x] Has undergone security/privacy review - bug fix. Not needed.
 
Implementation commitment:

 * [ ] Safari - @aestes, please provide link 🙏. 
 * [x] Chrome - current behavior. 
 * [x] Firefox - current behavior.
 * [ ] Edge - @zouhir - please confirm 🙏. 

None

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/821.html" title="Last updated on Jan 11, 2019, 3:13 AM UTC (bfd7c2e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/821/a1e773a...bfd7c2e.html" title="Last updated on Jan 11, 2019, 3:13 AM UTC (bfd7c2e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -936,8 +936,14 @@
           "<a>InvalidStateError</a>" <a>DOMException</a>.
           </li>
           <li>If the <a>user agent</a>'s <a>payment request is showing</a>
-          boolean is true, then return <a>a promise rejected with</a> an
-          "<a>AbortError</a>" <a>DOMException</a>.
+          boolean is true, then:
+            <ol>
+              <li>Set <var>request</var>.<a>[[\state]]</a> to "<a>closed</a>".
+              </li>
+              <li>Return <a>a promise rejected with</a> an
+              "<a>InvalidStateError</a>" <a>DOMException</a>.
+              </li>
+            </ol>
           </li>
           <li>Set <var>request</var>.<a>[[\state]]</a> to "<a>interactive</a>".
           </li>
--- a/index.html
+++ b/index.html
@@ -941,7 +941,7 @@
               <li>Set <var>request</var>.<a>[[\state]]</a> to "<a>closed</a>".
               </li>
               <li>Return <a>a promise rejected with</a> an
-              "<a>InvalidStateError</a>" <a>DOMException</a>.
+              "<a>AbortError</a>" <a>DOMException</a>.
               </li>
             </ol>
           </li>


marcoscaceres:
@zouhir, not sure if you can confirm this yet, but I guess Edge going forward will rely on Blink's implementation of Payment Request but still use Microsoft Wallet for payments. Is that a valid assumption? If yes, then I can just assume whatever is in Blink will work in Edge, in as far as the API mechanics are concerned (then I don't need to bug you to confirm things 😇 ).

marcoscaceres:
Added test: 
https://github.com/web-platform-tests/wpt/commit/5f1cbd0a42811a6c66bcfdf25f8a5e7023bf900e

rsolomakhin:
> assume whatever is in Blink will work in Edge.

May or may not, depending on where Chromium ends and Edge begins. Blink is a very thin layer that mostly forwards calls to the browser process, where most of the logic is implemented. The reason for that is we trust the renderer process as little as possible, since it parses and runs arbitrary HTML/JS from the web.

rsolomakhin (approved):

zouhir:
Thanks @marcoscaceres for checking with me - it is hard for me to confirm at the moment but the scnario you described is what I personally think is **ideal** for now. _generally speaking_ there is no intent to diverge. I believe in the following couple weeks I will have better comment on this 😊.

marcoscaceres:
Thanks @zouhir and @rsolomakhin for the details. 

I'll wait a few days for @aestes to confirm it's ok to make this small change, and there is not some obvious use case we might have overlooked.

marcoscaceres:
Going to move forward with this. Can revise if @aestes raises any concerns.

PR #822, opened 2019-01-15
Integrate with Feature Policy 

### LINKED ISSUE #600: Integration with Feature Policy
In V2, the spec should integrate with the [Feature Policy](https://wicg.github.io/feature-policy/) spec. We need to investigate what this means and/or reverse engineer the Web Platform Tests that Google has submitted around this.

### LINKED ISSUE #698: Disable Payment Request API in CSP/iframe sandbox
CSP/iframe sandbox is meant to host untrusted content (by locking down privildge of the untrusted 
 content). And powerful APIs such as Service Worker, AppCache, etc are not callable from sandboxed contents. For the same reason, Payment Request API should be disabled in sandxboxed content. If anyone feels that there's a valid use case of  Payment Request API in sandbox, then it should be only allowed with "allow-same-origin" keyword (Though I don't think there is such a use case).

marcoscaceres:
closes #600 
closes #698

Implementation commitment:

 * [ ] Safari (link to issue)
 * [x] Chrome (link to issue)
 * [X] Firefox (link to issue)
 * [ ] Edge (public signal)

None.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/822.html" title="Last updated on Jan 29, 2019, 1:24 AM UTC (1d9a28a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/822/5906b72...1d9a28a.html" title="Last updated on Jan 29, 2019, 1:24 AM UTC (1d9a28a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -644,14 +644,6 @@
           act as follows:
         </p>
         <ol data-link-for="PaymentDetailsBase" class="algorithm">
-          <li data-tests=
-          "allowpaymentrequest/active-document-cross-origin.https.sub.html, allowpaymentrequest/active-document-same-origin.https.html, allowpaymentrequest/removing-allowpaymentrequest.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest-timing.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest.https.sub.html">
-          If the <a>current settings object</a>'s <a data-cite=
-          "HTML#responsible-document">responsible document</a> is not
-          <a>allowed to use</a> the feature indicated by attribute name
-          <a>allowpaymentrequest</a>, then <a>throw</a> a
-          "<a>SecurityError</a>" <a>DOMException</a>.
-          </li>
           <li>Let <var>serializedMethodData</var> be an empty list.
           </li>
           <li>Establish the request's id:
@@ -3623,26 +3615,17 @@
     </section>
     <section class="informative">
       <h2>
-        <code>PaymentRequest</code> and <code>iframe</code> elements
-      </h2>
-      <p data-tests=
-      "allowpaymentrequest/active-document-cross-origin.https.sub.html, allowpaymentrequest/active-document-same-origin.https.html, allowpaymentrequest/allowpaymentrequest-attribute-cross-origin-bc-containers.https.html, allowpaymentrequest/allowpaymentrequest-attribute-same-origin-bc-containers.https.html, allowpaymentrequest/basic.https.html, allowpaymentrequest/no-attribute-cross-origin-bc-containers.https.html, allowpaymentrequest/no-attribute-same-origin-bc-containers.https.html, allowpaymentrequest/removing-allowpaymentrequest.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest-timing.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest.https.sub.html">
-        To indicate that a cross-origin <a>iframe</a> is allowed to invoke the
-        payment request API, the <a>allowpaymentrequest</a> attribute can be
-        specified on the <a>iframe</a> element.
-      </p>
-    </section>
-    <section class="informative">
-      <h2>
         Interaction with Feature Policy Specification
       </h2>
       <p>
         The <a data-cite="feature-policy">Feature Policy</a> specification
-        defines the "<code>payment</code>" feature, which controls whether the
-        current document is allowed to use the <a>PaymentRequest</a> interface.
-        Please refer to the <a data-cite="feature-policy">Feature Policy</a>
-        specification for usage examples, implementation details, and normative
-        conformance requirements.
+        defines the "<code>payment</code>" feature and the <code><dfn data-cit=
+        "feature-policy#attr-iframe-allowpaymentrequest">allowpaymentrequest</dfn></code>
+        attribute for <a>iframe</a> element, which control whether a particular
+        document, including cross-origin documents, is <a>allowed to use</a>
+        the <a>PaymentRequest</a> interface. Please refer to the <a data-cite=
+        "feature-policy">Feature Policy</a> specification for usage examples,
+        implementation details, and normative conformance requirements.
       </p>
       <p>
         Additionally, for details of how the "<code>payment</code>" feature and
@@ -5197,7 +5180,8 @@
           It is common for merchants and other payees to delegate checkout and
           other e-commerce activities to payment service providers through an
           <a>iframe</a>. This API supports payee-authorized cross-origin
-          iframes through [[HTML]]'s <a>allowpaymentrequest</a> attribute.
+          iframes through [[Feature-Policy]]'s <a>allowpaymentrequest</a>
+          attribute.
         </p>
         <p class="Note">
           <a>Payment handlers</a> have access to both the origin that hosts the
@@ -5347,9 +5331,7 @@
           "HTML/infrastructure.html#in-parallel">in parallel</dfn>, the
           <code><dfn data-cite=
           "HTML/iframe-embed-object.html#the-iframe-element">iframe</dfn></code>
-          element, and the <code><dfn data-cite=
-          "HTML/iframe-embed-object.html#attr-iframe-allowpaymentrequest">allowpaymentrequest</dfn></code>
-          attribute.
+          element.
         </dd>
         <dt>
           ECMAScript
--- a/index.html
+++ b/index.html
@@ -619,8 +619,9 @@
         <ol data-link-for="PaymentDetailsBase" class="algorithm">
           <li>If the <a>current settings object</a>'s <a data-cite=
           "HTML#responsible-document">responsible document</a> is not
-          <a>allowed to use</a> the "<a>payment</a>" feature, then <a>throw</a>
-          a "<a>SecurityError</a>" <a>DOMException</a>.
+          <a>allowed to use</a> the "<a data-lt="payment-feature">payment</a>"
+          feature, then <a>throw</a> a "<a>SecurityError</a>"
+          <a>DOMException</a>.
           </li>
           <li>Let <var>serializedMethodData</var> be an empty list.
           </li>
@@ -3486,7 +3487,7 @@
       </h2>
       <p>
         This specification defines a policy-controlled feature identified by
-        the string "<code><dfn data-lt="payment-feature" data-no-default=
+        the string "<code><dfn data-lt="payment-feature" data-nodefault=
         "">payment</dfn></code>". Its <a href=
         "feature-policy#default-allowlist">default allowlist</a> is
         '<code>self</code>'.
--- a/index.html
+++ b/index.html
@@ -646,9 +646,8 @@
         <ol data-link-for="PaymentDetailsBase" class="algorithm">
           <li>If the <a>current settings object</a>'s <a data-cite=
           "HTML#responsible-document">responsible document</a> is not
-          <a>allowed to use</a> the feature indicated by attribute name
-          <a>allowpaymentrequest</a>, then <a>throw</a> a
-          "<a>SecurityError</a>" <a>DOMException</a>.
+          <a>allowed to use</a> the "<a>payment</a>" feature, then <a>throw</a>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@ianbjacobs, can you give this a once over for me? Then I'll ping clelland to make sure it's technically accurate.

marcoscaceres:
Just noting some discussion happening here too:
https://github.com/web-platform-tests/wpt/pull/14855

marcoscaceres:
Annevk tells me that `allowpaymentrequest` has been moved out of the HTML spec into feature policy spec. 

I'll need to update the spec to remove reference to HTML for `allowpaymentrequest`.

index.html
```diff
@@ -635,14 +635,6 @@ <h2>
           act as follows:
         </p>
         <ol data-link-for="PaymentDetailsBase" class="algorithm">
-          <li data-tests=
-          "allowpaymentrequest/active-document-cross-origin.https.sub.html, allowpaymentrequest/active-document-same-origin.https.html, allowpaymentrequest/removing-allowpaymentrequest.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest-timing.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest.https.sub.html">
-          If the <a>current settings object</a>'s <a data-cite=
```
marcoscaceres, line 622: This is redundant now, given:
https://github.com/w3c/webappsec-feature-policy/blob/master/features.md#payment

@domenic, agree?

index.html
```diff
@@ -635,14 +635,6 @@ <h2>
           act as follows:
         </p>
         <ol data-link-for="PaymentDetailsBase" class="algorithm">
-          <li data-tests=
-          "allowpaymentrequest/active-document-cross-origin.https.sub.html, allowpaymentrequest/active-document-same-origin.https.html, allowpaymentrequest/removing-allowpaymentrequest.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest-timing.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest.https.sub.html">
```
marcoscaceres, line 621: Tests will be moving to Feature Policy directory as part of web-platform-tests/wpt#14855

marcoscaceres:
@clelland, whenever you can, would really appreciate if you could take a look at this PR. 

If you have input on web-platform-tests/wpt#14855, that would also be great.

index.html
```diff
@@ -3545,13 +3537,27 @@ <h2>
...
-        To indicate that a cross-origin <a>iframe</a> is allowed to invoke the
-        payment request API, the <a>allowpaymentrequest</a> attribute can be
-        specified on the <a>iframe</a> element.
+      <p>
+        The <a data-cite="feature-policy">Feature Policy</a> specification
+        defines the "<code>payment</code>" feature and the <code><dfn data-cit=
+        "feature-policy#attr-iframe-allowpaymentrequest">allowpaymentrequest</dfn></code>
+        attribute for <a>iframe</a> element, which control whether a particular
```
annevk: It only defines the processing model for this attribute, the attribute itself is defined by HTML.

index.html
```diff
@@ -3545,13 +3537,27 @@ <h2>
...
+        document, including cross-origin documents, is <a>allowed to use</a>
+        the <a>PaymentRequest</a> interface. Please refer to the <a data-cite=
+        "feature-policy">Feature Policy</a> specification for usage examples,
+        implementation details, and normative conformance requirements.
```
ianbjacobs: I think you can drop "normative" and just leave "conformance requirements."

ianbjacobs:
```suggestion
        implementation details, and conformance requirements.
```

index.html
marcoscaceres: Makes sense. Will change it back.

ianbjacobs (approved):
Hi Marcos,

I suggested one small change otherwise seems ok to me editorially.

marcoscaceres:
Appreciate the review @ianbjacobs!

domenic:
Is there any chance that we can kill `allowpaymentrequest` only use `allow="paymentrequest"` instead? Or is there too much deployed content?

index.html
domenic: No, this is wrong. The source of truth for how the PaymentRequest constructor behaves needs to be this spec. The fact that https://github.com/w3c/webappsec-feature-policy/blob/master/features.md#payment has a MUST requirement is just confusing. That document should be updated: <del>MUST throw a SecurityError</del><ins>will throw in the manner described in the PaymentRequest specification</ins>.

/cc @clelland to consider changes to avoid such confusion.

domenic (changes requested):

index.html
```diff
@@ -3545,13 +3537,27 @@ <h2>
...
+        Interaction with Feature Policy Specification
       </h2>
-      <p data-tests=
-      "allowpaymentrequest/active-document-cross-origin.https.sub.html, allowpaymentrequest/active-document-same-origin.https.html, allowpaymentrequest/allowpaymentrequest-attribute-cross-origin-bc-containers.https.html, allowpaymentrequest/allowpaymentrequest-attribute-same-origin-bc-containers.https.html, allowpaymentrequest/basic.https.html, allowpaymentrequest/no-attribute-cross-origin-bc-containers.https.html, allowpaymentrequest/no-attribute-same-origin-bc-containers.https.html, allowpaymentrequest/removing-allowpaymentrequest.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest-timing.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest.https.sub.html">
-        To indicate that a cross-origin <a>iframe</a> is allowed to invoke the
-        payment request API, the <a>allowpaymentrequest</a> attribute can be
-        specified on the <a>iframe</a> element.
+      <p>
+        The <a data-cite="feature-policy">Feature Policy</a> specification
+        defines the "<code>payment</code>" feature and the <code><dfn data-cit=
```
clelland: This spec should probably define the actual feature -- similarly to https://html.spec.whatwg.org/#policy-controlled-features or https://fullscreen.spec.whatwg.org/#feature-policy-integration, it's just a matter of declaring the feature name and the default allowlist ('self' in this case)

And, as @annevk mentioned, the allowpaymentrequest attribute is defined in HTML.

The note in the fullscreen spec is actually a pretty good fit for the situation here. `allowfullscreen` affects the container policy, unless overridden by `allow`.

index.html
clelland: Agreed. I'm removing anything that even smells like it might be normative for some other spec from that document.

This step here can be reinstated, just changed to use the phrasing "If the current settings object's responsible document is not allowed to use the 'payment' feature ... ", and link "payment" to the definition of the `payment` feature elsewhere in this doc (similarly to https://fullscreen.spec.whatwg.org/#feature-policy-integration)

marcoscaceres: For "V1" we are trying to avoid putting a normative dependency on Feature Policy (bc: document remains unpublished, unclear if WebKit will support it ATM). 

What I'd like to do instead is just say: "If the current settings object's responsible document is not allowed to use the API, throw a SecurityError" - implying that Feature Policy would deem the API usage "not allowed"... that hopefully gives us enough of a hook into Feature Policy without having a normative dependency on it. 

I can then point the reader to the sections Feature Policy and `allowpaymentrequest` below.

index.html
marcoscaceres: I agree. I'm kinda conflicted between us trying to get this to REC (important to a lot of members of the WG - aiming for April) and "doing the right thing" and just normatively defining "payment" here 🤔 Our plan is to publish a REC, but then quickly spin up a "v1.1" that would formally define the feature string.

index.html
marcoscaceres: Original text was actually fine, because it points to `allowpaymentrequest`, which itself names the "payment" feature and points to Feature Policy (while meeting our current requirements). Reverting.

marcoscaceres:
@domenic wrote: 

> Is there any chance that we can kill allowpaymentrequest only use allow="paymentrequest" instead? > Or is there too much deployed content?

@rsolomakhin, worth setting up a telemetry probe? It has shipped for about 2 years in Chrome, so might be too late... but never know.

index.html
annevk: You cannot get to REC without two implementations right? And implementations wouldn't be able to implement this any other way. Or would you leave this whole thing undefined or some such?

marcoscaceres: > You cannot get to REC without two implementations right? 

Correct. 

> And implementations wouldn't be able to implement this any other way. Or would you leave this whole thing undefined or some such?

The Feature Policy would be undefined, but there is sufficient prose in the spec (in the constructor) that an implementation that does implement either the `allowpaymentrequest` attribute or the Feature Policy spec would behave in conforming manner (and could be demonstratively shown to behave in a conforming manner).

domenic (changes requested):
To be clear, I don't think this is good in its current state, and I think PaymentRequest needs to define the feature. Feature Policy does not define individual features; the specs do.

marcoscaceres:
> To be clear, I don't think this is good in its current state, and I think PaymentRequest needs to define the feature. Feature Policy does not define individual features; the specs do.

I agree with this, and it's absolutely the right thing to do - and we might need to accept it as a critical piece of the design (particularly from a security stand point... even if it delays going to REC for a bit). 

Have you - or @clelland - heard anything from the WebKit folks about supporting Feature Policy? We've unfortunately not been able to get in touch with @aestes about WebKit's implementation plans.

marcoscaceres:
@domenic @clelland ok, done :) I based the text on what was stated in the Fullscreen spec.  

Please double check I didn't miss anything.

index.html
```diff
@@ -617,13 +617,10 @@ <h2>
...
         <ol data-link-for="PaymentDetailsBase" class="algorithm">
-          <li data-tests=
-          "allowpaymentrequest/active-document-cross-origin.https.sub.html, allowpaymentrequest/active-document-same-origin.https.html, allowpaymentrequest/removing-allowpaymentrequest.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest-timing.https.sub.html, allowpaymentrequest/setting-allowpaymentrequest.https.sub.html">
-          If the <a>current settings object</a>'s <a data-cite=
+          <li>If the <a>current settings object</a>'s <a data-cite=
           "HTML#responsible-document">responsible document</a> is not
-          <a>allowed to use</a> the feature indicated by attribute name
-          <a>allowpaymentrequest</a>, then <a>throw</a> a
-          "<a>SecurityError</a>" <a>DOMException</a>.
+          <a>allowed to use</a> the "<a>payment</a>" feature, then <a>throw</a>
```
clelland: Should this be `<a lt="payment-feature">payment</a>`, to link to the definition in `#feature-policy`?

clelland (commented):
This looks good, @marcoscaceres  -- one question about the internal cross-reference, but the text is exactly what I'd expect.

index.html
```diff
@@ -3475,12 +3472,49 @@ <h2>
...
         payment request API, the <a>allowpaymentrequest</a> attribute can be
-        specified on the <a>iframe</a> element.
+        specified on the <a>iframe</a> element. See <a href=
+        "#feature-policy"></a> for details of how <a>allowpaymentrequest</a>
+        and <a data-cite="feature-policy">Feature Policy</a> interact.
+      </p>
+    </section>
+    <section id="feature-policy">
+      <h2>
+        Feature Policy Integration
```
domenic:
```suggestion
        Feature Policy integration
```

domenic (approved):

index.html
marcoscaceres: @clelland that would be good. However, I did it this way because Feature Policy itself doesn't actually define the "payment" feature - thus I couldn't actually link to it. However, I'm happy update the link it once Feature Policy defines it.

clelland: I meant to link to the feature-policy section in this spec, where 'payment-feature' is actually defined, not to link to the feature policy spec. Features should ideally be defined in the spec where they are used, rather than centrally in the FP spec.

marcoscaceres: Ah, sorry for misunderstanding. Will fix.

PR #833, opened 2019-02-08
Add PaymentRequest.prototype.hasEnrolledInstrument()

### LINKED ISSUE #830: Spec hasEnrolledInstrument
I started implementing `hasEnrolledInstrument` in Chrome according to the WG agreement [1]. It will be great if this API and the change to `canMakePayment` can be added to the spec language. I'd like to kick off an Intent to Ship thread in blink-dev@ soon as an official announcement of the `canMakePayment` behavior change to web developers.

[1] https://lists.w3.org/Archives/Public/public-payments-wg/2018Nov/0018.html

danyao:
closes #830 

 * [x] Modified Web platform tests: https://github.com/web-platform-tests/wpt/pull/15306
 
Implementation commitment:

 * [x] [Safari](https://bugs.webkit.org/show_bug.cgi?id=197386)
 * [x] [Chrome](https://crbug.com/915907)
 * [ ] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1528663)
 * [ ] Edge (public signal)

After this change, `CanMakePaymentEvent` will not be triggered on any payment handler when `canMakePayment()` runs. It will be triggered when `hasEnrolledInstrument()` runs.

***
<a href="https://pr-preview.s3.amazonaws.com/danyao/payment-request/pull/833.html" title="Last updated on May 3, 2019, 2:50 AM UTC (6faac81)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/833/72de3ba...danyao:6faac81.html" title="Last updated on May 3, 2019, 2:50 AM UTC (6faac81)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1318,38 +1318,9 @@
           </p>
         </div>
         <p data-tests="payment-request-canmakepayment-method.https.html">
-          The <a>canMakePayment()</a> method MUST act as follows:
+          The <a>canMakePayment()</a> method MUST run the <a>can make payment
+          algorithm</a> with <var>checkForInstruments</var> set to false.
         </p>
-        <ol class="algorithm">
-          <li>Let <var>request</var> be the <a>PaymentRequest</a> object on
-          which the method was called.
-          </li>
-          <li>If <var>request</var>.<a>[[\state]]</a> is not "<a>created</a>",
-          then return <a>a promise rejected with</a> an
-          "<a>InvalidStateError</a>" <a>DOMException</a>.
-          </li>
-          <li>Let <var>hasHandlerPromise</var> be <a>a new promise</a>.
-          </li>
-          <li>Return <var>hasHandlerPromise</var>, and perform the remaining
-          steps <a>in parallel</a>.
-          </li>
-          <li>For each <var>paymentMethod</var> tuple in
-          <var>request</var>.<a>[[\serializedMethodData]]</a>:
-            <ol>
-              <li>Let <var>identifier</var> be the first element in the
-              <var>paymentMethod</var> tuple.
-              </li>
-              <li>If the user agent has a <a>payment handler</a> that supports
-              handling payment requests for <var>identifier</var>, or if it can
-              perform just-in-time installation of a suitable payment handler,
-              resolve <var>hasHandlerPromise</var> with true and terminate this
-              algorithm.
-              </li>
-            </ol>
-          </li>
-          <li>Resolve <var>hasHandlerPromise</var> with false.
-          </li>
-        </ol>
       </section>
       <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
         <h2>
@@ -1365,79 +1336,9 @@
           discretion.
         </p>
         <p data-tests="payment-request-hasenrolledinstrument-method.https.html">
-          The <a>hasEnrolledInstrument()</a> method MUST act as follows:
+          The <a>hasEnrolledInstrument()</a> method MUST run the <a>can make
+          payment algorithm</a> with <var>checkForInstruments</var> set to true.
         </p>
-        <ol class="algorithm">
-          <li>Let <var>request</var> be the <a>PaymentRequest</a> object on
-          which the method was called.
-          </li>
-          <li>If <var>request</var>.<a>[[\state]]</a> is not "<a>created</a>",
-          then return <a>a promise rejected with</a> an
-          "<a>InvalidStateError</a>" <a>DOMException</a>.
-          </li>
-          <li data-tests=
-          "payment-request/payment-request-hasenrolledinstrument-method-protection.https.html">
-          Optionally, at the <a>top-level browsing context</a>'s discretion,
-          return <a>a promise rejected with</a> a "<a>NotAllowedError</a>" <a>
-          DOMException</a>.
-            <p class="note" data-link-for="PaymentRequest">
-              This allows user agents to apply heuristics to detect and prevent
-              abuse of the <a>hasEnrolledInstrument()</a> method for
-              fingerprinting purposes, such as creating <a>PaymentRequest</a>
-              objects with a variety of supported <a>payment methods</a> and
-              calling <a>hasEnrolledInstrument()</a> on them one after the
-              other. For example, a user agent may restrict the number of
-              successful calls that can be made based on the <a>top-level
-              browsing context</a> or the time period in which those calls were
-              made.
-            </p>
-          </li>
-          <li>Let <var>promise</var> be <a>a new promise</a>.
-          </li>
-          <li>Return <var>promise</var> and perform the remaining steps <a>in
-          parallel</a>.
-          </li>
-          <li>For each <var>paymentMethod</var> tuple in
-          <var>request</var>.<a>[[\serializedMethodData]]</a>:
-            <ol>
-              <li>Let <var>identifier</var> be the first element in the
-              <var>paymentMethod</var> tuple.
-              </li>
-              <li>Let <var>data</var> be the result of <a data-cite=
-              "ECMASCRIPT#sec-json.parse">JSON-parsing</a> the second element
-              in the <var>paymentMethod</var> tuple.
-              </li>
-              <li>If required by the specification that defines the
-              <var>identifier</var>, then <a data-cite=
-              "WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
-              <var>data</var> to an IDL value. Otherwise, <a data-cite=
-              "WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a> to
-              <a data-cite="WEBIDL#idl-object">object</a>.
-              </li>
-              <li>If conversion results in an <a data-cite=
-              "WEBIDL#dfn-exception">exception</a> <var>error</var>, reject
-              <var>promise</var> with <var>error</var> and terminate this
-              algorithm.
-              </li>
-              <li>Let <var>handlers</var> be a <a>list</a> of registered
-              <a>payment handlers</a> that are authorized and can handle
-              payment request for <var>identifier</var>.
-              </li>
-              <li>For each <var>handler</var> in <var>handlers</var>:
-                <ol>
-                  <li>Let <var>hasEnrolledInstrument</var> be the result of
-                  running <var>handler</var>'s <a>steps to check if a payment
-                  can be made</a> with <var>data</var>.
-                  </li>
-                  <li>If <var>hasEnrolledInstrument</var> is true, resolve
-                  <var>promise</var> with true, and return.
-                  </li>
-                </ol>
-              </li>
-            </ol>
-          <li>Resolve <var>promise</var> with false.
-          </li>
-        </ol>
       </section>
       <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
         <h2>
@@ -4340,6 +4241,99 @@
       </section>
       <section>
         <h2>
+          Can make payment algorithm
+        </h2>
+        <p>
+          The <dfn>can make payment algorithm</dfn> checks if the <a>user
+          agent</a> supports making payment with a <a>PaymentRequest</a>. It
+          takes a boolean argument, <var>checkForInstruments</var>, that
+          specifies whether the algorithm should check for existence of enrolled
+          instruments in addition to basic support of a <a>payment method</a>.
+        </p>
+        <ol class="algorithm">
+          <li>Let <var>request</var> be the <a>PaymentRequest</a> object on
+          which the method was called.
+          </li>
+          <li>If <var>request</var>.<a>[[\state]]</a> is not "<a>created</a>",
+          then return <a>a promise rejected with</a> an
+          "<a>InvalidStateError</a>" <a>DOMException</a>.
+          </li>
+          <li data-tests=
+          "payment-request/payment-request-hasenrolledinstrument-method-protection.https.html">
+          If <var>checkForInstruments</var> is true, optionally, at the
+          <a>top-level browsing context</a>'s discretion, return <a>a promise
+          rejected with</a> a "<a>NotAllowedError</a>" <a> DOMException</a>.
+            <p class="note" data-link-for="PaymentRequest">
+              This allows user agents to apply heuristics to detect and prevent
+              abuse of the <a>hasEnrolledInstrument()</a> method for
+              fingerprinting purposes, such as creating <a>PaymentRequest</a>
+              objects with a variety of supported <a>payment methods</a> and
+              calling <a>hasEnrolledInstrument()</a> on them one after the
+              other. For example, a user agent may restrict the number of
+              successful calls that can be made based on the <a>top-level
+              browsing context</a> or the time period in which those calls were
+              made.
+            </p>
+          </li>
+          <li>Let <var>promise</var> be <a>a new promise</a>.
+          </li>
+          <li>Return <var>promise</var>, and perform the remaining
+          steps <a>in parallel</a>.
+          </li>
+          <li>For each <var>paymentMethod</var> tuple in
+          <var>request</var>.<a>[[\serializedMethodData]]</a>:
+            <ol>
+              <li>Let <var>identifier</var> be the first element in the
+              <var>paymentMethod</var> tuple.
+              </li>
+              <li>If <var>checkForInstruments</var> is false, resolve <var>
+              promise</var> with true and terminate this algorithm if either the
+              user agent has a <a>payment handler</a> that supports handling
+              payment requests for <var>identifier</var>, or if it can perform
+              just-in-time installation of a suitable payment handler.
+              </li>
+              <li>If <var>checkForInstruments</var> is true:
+                <ol>
+                  <li>Let <var>data</var> be the result of <a data-cite=
+                  "ECMASCRIPT#sec-json.parse">JSON-parsing</a> the second element
+                  in the <var>paymentMethod</var> tuple.
+                  </li>
+                  <li>If required by the specification that defines the
+                  <var>identifier</var>, then <a data-cite=
+                  "WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+                  <var>data</var> to an IDL value. Otherwise, <a data-cite=
+                  "WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a> to
+                  <a data-cite="WEBIDL#idl-object">object</a>.
+                  </li>
+                  <li>If conversion results in an <a data-cite=
+                  "WEBIDL#dfn-exception">exception</a> <var>error</var>, reject
+                  <var>promise</var> with <var>error</var> and terminate this
+                  algorithm.
+                  </li>
+                  <li>Let <var>handlers</var> be a <a>list</a> of registered
+                  <a>payment handlers</a> that are authorized and can handle
+                  payment request for <var>identifier</var>.
+                  </li>
+                  <li>For each <var>handler</var> in <var>handlers</var>:
+                    <ol>
+                      <li>Let <var>hasEnrolledInstrument</var> be the result of
+                      running <var>handler</var>'s <a>steps to check if a payment
+                      can be made</a> with <var>data</var>.
+                      </li>
+                      <li>If <var>hasEnrolledInstrument</var> is true, resolve
+                      <var>promise</var> with true and terminate this algorithm.
+                      </li>
+                    </ol>
+                  </li>
+                </ol>
+            </ol>
+          </li>
+          <li>Resolve <var>promise</var> with false.
+          </li>
+        </ol>
+      </section>
+      <section>
+        <h2>
           Shipping address changed algorithm
         </h2>
         <p data-tests=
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Thanks @danyao! I’ll try to review this ASAP!

danyao:
Thanks! I forgot to update the PaymentRequest interface with the new method in the first commit. This is fixed now.

index.html
```diff
@@ -1300,20 +1302,71 @@ <h2>
...
+          <li>Return <var>hasHandlerPromise</var>, and perform the remaining
+          steps <a>in parallel</a>.
+          </li>
+          <li>For each <var>paymentMethod</var> tuple in
+          <var>request</var>.<a>[[\serializedMethodData]]</a>:
+            <ol>
+              <li>Let <var>identifier</var> be the first element in the
+              <var>paymentMethod</var> tuple.
+              </li>
+              <li>If the user agent has a <a>payment handler</a> that support
```
ianbjacobs:
```suggestion
              <li>If the user agent has a <a>payment handler</a> that supports
```

index.html
```diff
@@ -1300,20 +1302,71 @@ <h2>
...
+      </section>
+      <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
+        <h2>
+          <dfn>hasEnrolledInstrument()</dfn> method
+        </h2>
+        <p class="note">
+          The <a>hasEnrolledInstrument()</a> method can be used by the developer
+          to determine if the <a>user agent</a> not only has support for one of
+          the desired <a>payment methods</a> but is also "ready for payment", (
+          e.g. when showing a "buy now" button). If the method is called too
```
ianbjacobs:
```suggestion
          e.g., when showing a "buy now" button). If the method is called too
```

index.html
```diff
@@ -1300,20 +1302,71 @@ <h2>
...
+          often, the user agent might instead return <a>a promise rejected
```
ianbjacobs:
```suggestion
          often, the user agent may return <a>a promise rejected
```

index.html
```diff
@@ -1323,40 +1376,66 @@ <h2>
...
               algorithm.
               </li>
+              <li>Let <var>handlers</var> be a <a>list</a> of registered
+              <a>payment handlers</a> that are authorized and can handle
+              payment request for <var>identifier</var>.
+              </li>
+              <li>For each <var>handler</var> in <var>handlers</var>:
+                <ol>
+                  <li>Let <var>hasEnrolledInstrument</var> be the result of
+                  running <var>handler</var>'s <a>steps to check if a payment
```
ianbjacobs: I have a question here: Instead of "steps to check if a payment can be made" should this be "steps to check if the handler has an enrolled instrument"? And does this mean that a payment method also needs to define those steps? Does a revised basic card spec need to change the steps to be about enrolled instrument?

(I don't have answers to these questions yet.)

[1] https://www.w3.org/TR/payment-method-basic-card/#steps-to-check-if-a-payment-can-be-made

ianbjacobs (dismissed):

index.html
danyao: In Chrome we're using the existing steps in the basic card spec to determine if a card should be considered "enrolled". For payment handler, I believe the equivalent is the CanMakePaymentEvent. In both cases, our rationale is that the new canMakePayment() method checks the existence of a payment handler for the specified method, and hasEnrolledInstrument() checks the actual "can-make-payment" property of the payment handler.

I think renaming "steps to check if a payment can be made" to "steps to check if the handler has an enrolled instrument" may be more clear in the long run, but it introduces a lot of changes to existing spec without changing any behavior. That's why I opted to reuse the existing language as much as possible.

index.html
```diff
@@ -1300,20 +1302,71 @@ <h2>
...
-          determine if the <a>PaymentRequest</a> object can be used to make a
-          payment, before they call <a>show()</a>. It returns a <a>Promise</a>
-          that will be fulfilled with true if the <a>user agent</a> supports
-          any of the desired <a>payment methods</a> supplied to the
-          <a>PaymentRequest</a> constructor, and false if none are supported.
-          If the method is called too often, the user agent might instead
-          return <a>a promise rejected with</a> a "<a>NotAllowedError</a>"
-          <a>DOMException</a>, at its discretion.
-        </p>
+        <div class="note">
```
```suggestion
        <div class="note" title="canMakePayment() vs hasEnrolledInstrument()">
```

index.html
```diff
@@ -1300,20 +1302,71 @@ <h2>
...
+          <p>
```
marcoscaceres, line 1318: I realize now that the first sentence of this paragraphs is somewhat misleading, but the second part is definitely correct. I'd suggest:

```HTML
          <p>
            The <a>canMakePayment()</a> method returns a <a>Promise</a>
            that is fulfilled with true if the <a>user agent</a> supports
            any of the <a data-link-for="PaymentMethodData">supportedMethods</a> supplied to the
            <a>PaymentRequest</a> <a data-lt="PaymentRequest.PaymentRequest()">constructor</a>, or false if none are supported.
          </p>
          <p>
            A true result from <a>canMakePayment()</a> does not imply that the
            user has an provisioned instrument ready for payment. For that, use
            <a>hasEnrolledInstrument()</a> instead.
          </p>
```

Needs a tidy :)

index.html
```diff
@@ -1300,20 +1302,71 @@ <h2>
...
+        </ol>
+      </section>
+      <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
+        <h2>
+          <dfn>hasEnrolledInstrument()</dfn> method
+        </h2>
+        <p class="note">
+          The <a>hasEnrolledInstrument()</a> method can be used by the developer
+          to determine if the <a>user agent</a> not only has support for one of
+          the desired <a>payment methods</a> but is also "ready for payment", (
```
```suggestion
          the desired <a>payment methods</a> but is also "ready for payment" (
```

index.html
```diff
@@ -1300,20 +1302,71 @@ <h2>
...
+          e.g., when showing a "buy now" button). If the method is called too
+          often, the user agent may return <a>a promise rejected
+          with </a> a "<a>NotAllowedError</a>" <a>DOMException</a>, at its
```
```suggestion
          with </a> a "<a>NotAllowedError</a>" <a>DOMException</a> at its
```

marcoscaceres (changes requested):
Hi @danyao, this is a great start - and the algorithm parts are looking pretty solid... 

However, I'd like to do this in two parts because there is some duplication in the two algorithms that I'd like to avoid (mostly an editorial/maintenance issue). Also note that the `hasEnrolledInstrument()` doesn't seem to integrate with the state machine like the other methods.

For example, right now:

```
pr = new PaymentRequest("")
await pr.abort();
await pr.canMakePayment(); //throws
await pr.hasEnrolledInstrument(); // doesn't throw.
``` 

I think we should have a single algorithm in the Algorithms sections. The algorithm should be called:

"can make payment algorithm" and have it take an argument Boolean `<var>checkForInstruments</var>` that if passed as "true", it does the instrument check.   

Then,  instead of:

```HTML
The <a>canMakePayment()</a> method MUST act as follows:
```

It would be:

```
The <a>canMakePayment()</a> method MUST run the <a>can make payment algorithm</a> with the <var>checkForInstruments</a> set to false.
```

Then,  instead of:

```HTML
The <a>hasEnrolledInstrument()</a> method MUST act as follows:
```

It would be:

```HTML
The <a>hasEnrolledInstrument()</a> method MUST run the <a>can make payment algorithm</a> with the <var>checkForInstruments</a> set to true.
```

marcoscaceres:
Now I'm wondering if we should just add a optional dictionary to `canMakePayment()`:

```JS
pr.canMakePayment({ withEnrolledInstruments: true });
```

webpayments-specs:
That would make feature detection difficult, so I would advise against it.

marcoscaceres:
Firefox would likely never return this information truthfully for built in methods regardless (i.e., we will always return true), so I don’t see what difference it makes from a feature detection POV?

marcoscaceres:
(Note the hasEnrolledInstruments() would introduce another fingerprinting vector - so we might want to document that in the Privacy and Security section, and we should make it ok for UAs to lie for privacy reasons).

rsolomakhin:
Chrome would return the `hasEnrolledInstrument()` truthfully for the built-in `basic-card` payment method. The feature detection matters for Chrome, because the old implementation of `canMakePayment()` was checking for instruments, whereas the new implementation would not be doing that. The only way for a website to determine which is which is to check for `hasEnrolledInstrument()` in the case of Chrome. Sorry about our legacy shipped code! :-(

> Note the hasEnrolledInstruments() would introduce another fingerprinting vector - so we might want to document that in the Privacy and Security section.

Privacy and Security section changes sound reasonable.

> We should make it ok for UAs to lie for privacy reasons.

Did you have in mind returning `NotAllowedError` similar to `canMakePayment()`? That sounds OK.

marcoscaceres:
Ok, let’s go with the new name 👍 about the exceptions, see my suggestion about using the same abstract algorithm for both this an canMakePayment(). Then they are both protected by the same abuse checking.

index.html
danyao: Done!

danyao:
Hi @marcoscaceres, thanks for the feedback!

> However, I'd like to do this in two parts because there is some duplication in the two algorithms that I'd like to avoid (mostly an editorial/maintenance issue).

Consolidating redundant algorithm makes sense. I went the other way initially because I felt it may be confusing to embed an "if" and the rate-limit note inside the loop. I gave it another try. PTAL.

> 
 Also note that the `hasEnrolledInstrument()` doesn't seem to integrate with the state machine like the other methods.
> 
> For example, right now:
> 
> ```
> pr = new PaymentRequest("")
> await pr.abort();
> await pr.canMakePayment(); //throws
> await pr.hasEnrolledInstrument(); // doesn't throw.
> ```

This would be a bug. How did you test it? I added a manual WPT test case for this scenario and it was passing on my debug build, but I could have overlooked something: https://github.com/web-platform-tests/wpt/blob/master/payment-request/payment-request-hasenrolledinstrument-method-manual.https.html#L81

> (Note the hasEnrolledInstruments() would introduce another fingerprinting vector - so we might want to document that in the Privacy and Security section, and we should make it ok for UAs to lie for privacy reasons).

Updated the Privacy Considerations section. PTAL.

Do you think it's important to keep the abuse protection for `canMakePayment()`? I feel it's no longer necessary because with the introduction of JIT and removal of handler-specific can make payment check, the new `canMakePayment()` is mostly an UA version detector. `hasEnrolledInstrument()` on the other hand still exposes user information, so should be kept under abuse protection.

danyao:
Friendly ping @marcoscaceres. I've responded to your feedback. PTAL.

index.html
```diff
@@ -4259,6 +4239,99 @@ <h2>
...
       </section>
+      <section>
+        <h2>
+          Can make payment algorithm
+        </h2>
+        <p>
+          The <dfn>can make payment algorithm</dfn> checks if the <a>user
+          agent</a> supports making payment with a <a>PaymentRequest</a>. It
+          takes a boolean argument, <var>checkForInstruments</var>, that
+          specifies whether the algorithm should check for existence of enrolled
```
```suggestion
          specifies whether the algorithm checks for existence of enrolled
```

index.html
```diff
@@ -4259,6 +4239,99 @@ <h2>
...
+          instruments in addition to basic support of a <a>payment method</a>.
```
```suggestion
          instruments in addition to supporting <a>payment method</a>.
```

index.html
```diff
@@ -4259,6 +4239,99 @@ <h2>
...
+          "<a>InvalidStateError</a>" <a>DOMException</a>.
+          </li>
+          <li data-tests=
+          "payment-request/payment-request-hasenrolledinstrument-method-protection.https.html">
+          If <var>checkForInstruments</var> is true, optionally, at the
+          <a>top-level browsing context</a>'s discretion, return <a>a promise
+          rejected with</a> a "<a>NotAllowedError</a>" <a> DOMException</a>.
+            <p class="note" data-link-for="PaymentRequest">
+              This allows user agents to apply heuristics to detect and prevent
+              abuse of the <a>hasEnrolledInstrument()</a> method for
```
```suggestion
              abuse of the calling method for
```

index.html
```diff
@@ -4259,6 +4239,99 @@ <h2>
...
+              fingerprinting purposes, such as creating <a>PaymentRequest</a>
```
marcoscaceres: Note that canMakePayment() is also a fingerprinting vector, so it’s prudent for this to apply to both.

index.html
```diff
@@ -4259,6 +4239,99 @@ <h2>
...
+          <li>For each <var>paymentMethod</var> tuple in
+          <var>request</var>.<a>[[\serializedMethodData]]</a>:
+            <ol>
+              <li>Let <var>identifier</var> be the first element in the
+              <var>paymentMethod</var> tuple.
+              </li>
+              <li>If <var>checkForInstruments</var> is false, resolve <var>
+              promise</var> with true and terminate this algorithm if either the
+              user agent has a <a>payment handler</a> that supports handling
+              payment requests for <var>identifier</var>, or if it can perform
```
marcoscaceres: I don't think we should mention the "just-in-time" feature... that's a Chrome implementation detail that falls under "supports".

index.html
```diff
@@ -4259,6 +4239,99 @@ <h2>
           </li>
         </ol>
       </section>
+      <section>
+        <h2>
+          Can make payment algorithm
+        </h2>
+        <p>
+          The <dfn>can make payment algorithm</dfn> checks if the <a>user
+          agent</a> supports making payment with a <a>PaymentRequest</a>. It
```
```suggestion
          agent</a> supports making payment with the <a>payment methods</a> with which the <a>PaymentRequest</a> was constructed. It
```

index.html
```diff
@@ -1300,65 +1302,43 @@ <h2>
...
+            The <a>canMakePayment()</a> method returns a <a>Promise</a>
+            that is fulfilled with true if the <a>user agent</a> supports
+            any of the <a data-link-for="PaymentMethodData">supportedMethods</a>
+            supplied to the <a>PaymentRequest</a>
+            <a data-lt="PaymentRequest.PaymentRequest()">constructor</a>, or
+            false if none are supported.
+          </p>
+          <p>
+            A true result from <a>canMakePayment()</a> does not imply that the
+            user has an provisioned instrument ready for payment. For that, use
```
```suggestion
            user has a provisioned instrument ready for payment. For that, use
```

index.html
```diff
@@ -5260,15 +5333,15 @@ <h2>
       </section>
       <section>
         <h2>
-          canMakePayment() protections
+          hasEnrolledInstrument() protections
```
```suggestion
          <code>hasEnrolledInstrument()</code> protections
```

index.html
```diff
@@ -5260,15 +5333,15 @@ <h2>
       </section>
       <section>
```
```suggestion
       <section id="hasEnrolledInstrument-privacy">
```

index.html
danyao: Done.

index.html
danyao: Makes sense. Removed the JIT language.

danyao:
@marcoscaceres All comments addressed. PTAL.

index.html
```diff
@@ -5260,15 +5333,15 @@ <h2>
...
+          hasEnrolledInstrument() protections
         </h2>
         <p data-link-for="PaymentRequest">
-          The <a>canMakePayment()</a> method enables the payee to call
-          <a>show()</a> if the user is ready to take advantage of the API, or
-          to fall back to a legacy checkout experience if not. Because this
-          method shares some information with the payee, user agents are
-          expected to protect the user from abuse of the method, for example,
-          by restricting the number or frequency of calls.
+          The <a>hasEnrolledInstrument()</a> method enables the payee to check
```
marcoscaceres: I'd suggest dropping the first sentence as it redefines what was already stated in the method's definition. Instead, just say something like: 

"The <a>hasEnrolledInstrument()</a> method has the potential to expose user information that could be abused for fingerprinting purposes. The API allows the user agent to restrict the number or frequency of calls to reduce the risk of fingerprinting. User agents MAY allow the user to control the response to <a>hasEnrolledInstrument()</a> via UI or provide their own means to protection (e.g., always returning with true)."

index.html
```diff
@@ -1300,65 +1302,43 @@ <h2>
...
-          payment, before they call <a>show()</a>. It returns a <a>Promise</a>
-          that will be fulfilled with true if the <a>user agent</a> supports
-          any of the desired <a>payment methods</a> supplied to the
-          <a>PaymentRequest</a> constructor, and false if none are supported.
-          If the method is called too often, the user agent might instead
-          return <a>a promise rejected with</a> a "<a>NotAllowedError</a>"
-          <a>DOMException</a>, at its discretion.
+          The <a>hasEnrolledInstrument()</a> method can be used by the developer
+          to determine if the <a>user agent</a> not only has support for one of
+          the desired <a>payment methods</a> but is also "ready for payment" (
```
marcoscaceres: I'd suggest dropping everything after "but also..." and replacing it with:

"and if a <a>payment handler</a> has an instrument ready for payment. See <a href="#"hasEnrolledInstrument-privacy"><a> for privacy considerations."

And also drop the actual details (i.e., the "If the method is called too often...").

marcoscaceres (changes requested):
Getting there... there is still a bit or repetition and some other little bits that I'd like to suggest we change.

index.html
```diff
@@ -4259,6 +4239,98 @@ <h2>
...
+          </li>
+          <li>If <var>request</var>.<a>[[\state]]</a> is not "<a>created</a>",
+          then return <a>a promise rejected with</a> an
+          "<a>InvalidStateError</a>" <a>DOMException</a>.
+          </li>
+          <li data-tests=
+          "payment-request-hasenrolledinstrument-method-protection.https.html, payment-request-canmakepayment-method-protection.https.html">
+          Optionally, at the <a>top-level browsing context</a>'s discretion,
+          return <a>a promise rejected with</a> a "<a>NotAllowedError</a>"
+          <a> DOMException</a>.
```
```suggestion
          <a>DOMException</a>.
```

index.html
```diff
@@ -4259,6 +4239,98 @@ <h2>
...
+                  in the <var>paymentMethod</var> tuple.
+                  </li>
+                  <li>If required by the specification that defines the
+                  <var>identifier</var>, then <a data-cite=
+                  "WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a>
+                  <var>data</var> to an IDL value. Otherwise, <a data-cite=
+                  "WEBIDL#dfn-convert-ecmascript-to-idl-value">convert</a> to
+                  <a data-cite="WEBIDL#idl-object">object</a>.
+                  </li>
+                  <li>If conversion results in an <a data-cite=
```
marcoscaceres: I think we can drop this `li`, because we did this check in the constructor so this can't ever throw.

index.html
```diff
@@ -4259,6 +4239,98 @@ <h2>
...
+        <h2>
+          Can make payment algorithm
+        </h2>
+        <p>
+          The <dfn>can make payment algorithm</dfn> checks if the <a>user
+          agent</a> supports making payment with the <a>payment methods</a> with
+          which the <a>PaymentRequest</a> was constructed. It takes a boolean
+          argument, <var>checkForInstruments</var>, that specifies whether the
+          algorithm checks for existence of enrolled instruments in addition to
+          supporting <a>payment method</a>.
```
```suggestion
          supporting a <a>payment method</a>.
```

index.html
danyao: Done.

index.html
danyao: Done. I also updated the corresponding note in canMakePayment() section above to match the wording, and also refer to the privacy section.

index.html
danyao: This is better! Thanks for the suggestion.

index.html
danyao: Done.

marcoscaceres:
Thanks for the speedy turn around. Will give it one more look over, but looking really good.

marcoscaceres (approved):
LGTM. Thanks for doing this @danyao.  

Before we merge, we need:
 1. Approval and implementation commitment from @aestes. 
 1. Firefox bug (I'll file one).

We also need to decide if we are going to fork 1.1 at this point. I'm waiting on final review feedback from various Editors of 1.0.

marcoscaceres:
Firefox bug https://bugzilla.mozilla.org/show_bug.cgi?id=1528663

danyao:
Friendly ping, @aestes.

index.html
```diff
@@ -5258,17 +5321,19 @@ <h2>
...
-          to fall back to a legacy checkout experience if not. Because this
-          method shares some information with the payee, user agents are
-          expected to protect the user from abuse of the method, for example,
-          by restricting the number or frequency of calls.
+          The <a>canMakePayment()</a> and <a>hasEnrolledInstrument()</a> methods
+          have the potential to expose user information that could be abused for
+          fingerprinting purposes. The API allows the user agent to restrict the
+          number or frequency of calls to reduce the risk of fingerprinting.
+          User agents MAY allow the user to control the response to either
+          method via UI or provide their own means to protection (e.g., always
```
aestes:
```suggestion
          method via UI or provide their own means of protection (e.g., always
```

aestes (approved):

index.html
danyao: Hi @ianbjacobs - did I answer your question? This is the last open review question blocking this PR. :-)

danyao:
Hi @marcoscaceres,

Now that this PR is approved, what would be the best next step? Should we create a 1.1 branch to commit this?

index.html
ianbjacobs: @danyao, sorry for the delay. Thank you for the answer. I'm ok with your proposal.

ianbjacobs:
@plehegar do you have guidance here for introducing a new branch?

Ian

marcoscaceres:
@danyao @plehegar @ianbjacobs, my preference is for us not to spin up new branches as we are almost ready to republish in the next 2 weeks. 

@danyao, just hang tight, we are almost done with 1.0. Once the new CR is out, we don't expect any significant feedback - so it will be easier to deal with back porting, or uplifting, changes as needed (and all this stuff will go into the `gh-pages` branch).

marcoscaceres:
If we can get commitment from one more implementer we can land this.

ianbjacobs:
@danyao, @rsolomakhin, @zouhir, @aestes see above request from @marcoscaceres. :)

rsolomakhin:
@ianbjacobs : This is [shipped in Chrome](https://crbug.com/915907) and there's a [bug in WebKit](https://bugs.webkit.org/show_bug.cgi?id=197386). Is that sufficient?

marcoscaceres:
@rsolomakhin, excellent! thanks for doing the archeology! Yep, that's good enough.

ianbjacobs:
That sounds like 2 implementation commitments. @marcoscaceres, seem ok?

marcoscaceres:
@danyao could you please check the merge conflict? I can modernize the markup a bit after we merge.

danyao:
Yep I'll take a pass later today. Thanks for keeping on top of this!

danyao:
Hi @marcoscaceres - I fixed the merge conflicts. PTAL!

marcoscaceres:
Thanks @danyao! looks great.

PR #838, opened 2019-02-28
Editorial Fit & Finish Changes

rvm4:
Editorial changes, though I did have additional editorial comments/questions that I wanted to sanity check in the pull request discussion:

1) There is no mention to the payment handler spec even though handlers are mentioned several times. Should there be a mention in the dependencies or refs section?
2) Why is there both a dependencies and refs section, don't they signify the same thing, especially given the breakdown to normative and informative in the refs section.
3) This is more of an architectural comment, but having the PaymentResponse not be an immutable object (with respect to the retry operation) seems like something we may want to reconsider in favor of immutability and an explicit return value for retry (if we haven't already).
4) The first paragraph after the NOTE block in section 3 (19th page) is incredibly hard to read. I don't have much context on the concept it is describing, so I didn't feel confident I could write an accurate and cleaner sentence.
5) The description of "error" in 6.3 is not very clear. Is it accurate to describe this as a global error message or a fallback error message when a more specific error field is not appropriate?
6) Section 11.4 pulls the term redactList out of nowhere, with no context. Is it fair to say that this is a user agent constructed list of field thought to provide security benefits if redacted?
7) It looks funky that the Assert statements in algorithm sections get their own highlighting box when no other standard term does, is that avoidable?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/838.html" title="Last updated on Feb 28, 2019, 2:10 PM UTC (7b4b8d6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/838/c4e599b...7b4b8d6.html" title="Last updated on Feb 28, 2019, 2:10 PM UTC (7b4b8d6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -393,7 +393,7 @@
               amount: { currency: "USD", value: "25.00" }
             },
           ];
-          Object.assign(details, { "shippingOptions": shippingOptions });
+          Object.assign(details, { shippingOptions });
         </pre>
       </section>
       <section>
--- a/index.html
+++ b/index.html
@@ -124,7 +124,7 @@
           Changes since last publication
         </h3>
         <p>
-          Substantive changes to Payment Request API since the 9 July 2018
+          Substantive changes to the Payment Request API since the 9 July 2018
           version are as follows. The complete list of changes, including all
           editorial changes, is viewable in the <a href=
           "https://github.com/w3c/payment-request/commits/gh-pages">commit
@@ -344,7 +344,7 @@
           second argument of the constructor (<var>details</var>) to provide
           the details of the transaction that the user is being asked to
           complete. This includes the total of the order and, optionally, some
-          line items that provide can provide a detailed breakdown of what is
+          line items that can provide a detailed breakdown of what is
           being paid for.
         </p>
         <pre class="example js" title="The `details` argument">
@@ -393,7 +393,7 @@
               amount: { currency: "USD", value: "25.00" }
             },
           ];
-          Object.assign(details, { shippingOptions });
+          Object.assign(details, { "shippingOptions": shippingOptions });
         </pre>
       </section>
       <section>
@@ -1398,7 +1398,7 @@
               <li>Let <var>identifier</var> be the first element in the
               <var>paymentMethod</var> tuple.
               </li>
-              <li>If there user agent has a <a>payment handler</a> that support
+              <li>If the user agent has a <a>payment handler</a> that supports
               handling payment requests for <var>identifier</var>, resolve
               <var>hasHandlerPromise</var> with true and terminate this
               algorithm.
@@ -1441,7 +1441,7 @@
         <p data-tests="onmerchantvalidation-attribute.https.html">
           A <a>PaymentRequest</a>'s <a>onmerchantvalidation</a> attribute is an
           <a>EventHandler</a> for a <a>MerchantValidationEvent</a> named
-          "<a>merchantvalidation</a>".
+          <a>merchantvalidation</a>.
         </p>
       </section>
       <section data-dfn-for="PaymentRequest" data-link-for="PaymentRequest">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1441,7 +1441,7 @@ <h2>
         <p data-tests="onmerchantvalidation-attribute.https.html">
           A <a>PaymentRequest</a>'s <a>onmerchantvalidation</a> attribute is an
           <a>EventHandler</a> for a <a>MerchantValidationEvent</a> named
-          "<a>merchantvalidation</a>".
+          <a>merchantvalidation</a>.
```
marcoscaceres: Going to keep these, as they map nicely to literal strings in c++.

marcoscaceres (approved):
This is awesome @rvm4! great finds. Thanks so much for taking the time to read through it.

marcoscaceres:
> 1) There is no mention to the payment handler spec even though handlers are mentioned several times. Should there be a mention in the dependencies or refs section?

Our intention was not to make it a dependency, as, for instance, Safari and Firefox don't implement the payment handlers API. If the changes in the future, then yeah :)  

> 2) Why is there both a dependencies and refs section, don't they signify the same thing, especially given the breakdown to normative and informative in the refs section.

The dependencies section is basically an somewhat redundant editorial tool that helps us with spec writing. Eventually, we want the dependencies section to be automatically generated (i.e., have a section "Terms cited by reference".... but ReSpec can't do that yet. Hopefully later in the year.) 

However, I can't get rid of it because it would break a bunch of things :( 

> 3) This is more of an architectural comment, but having the PaymentResponse not be an immutable object (with respect to the retry operation) seems like something we may want to reconsider in favor of immutability and an explicit return value for retry (if we haven't already).

That one is complicated, but yeah... the problem is that it has its own state machine inside that controls various things. It's as immutable as it can be, given the constraints that we have... ideally, we should have never had PaymentResponse... I think Apple's Session model was a better solution... water... bridge... personal opinions. 

But seriously, unless it's going to impede certain kinds of interactions, I'd be inclined to just live with it. It's "good enough"™️

> 4) The first paragraph after the NOTE block in section 3 (19th page) is incredibly hard to read. I don't have much context on the concept it is describing, so I didn't feel confident I could write an accurate and cleaner sentence.

Let's do that in a new bug. 

> 5) The description of "error" in 6.3 is not very clear. Is it accurate to describe this as a global error message or a fallback error message when a more specific error field is not appropriate?

As above. Let's do that in a new bug. 

> 6) Section 11.4 pulls the term redactList out of nowhere, with no context. Is it fair to say that this is a user agent constructed list of field thought to provide security benefits if redacted?

Privacy, but you are correct. Could probably be better introduced.  

7) It looks funky that the Assert statements in algorithm sections get their own highlighting box when no other standard term does, is that avoidable?

It is, but it's convention from other specs... so we are just cargo culting. We make them stand out because `assert:` literally means crash the browser. We don't ever want to hit those, so we want browser implementers to be sanity checking what we wrote via those asserts.

index.html
rvm4: Can you elaborate on this? Who would reasonably make that connection and why are the other handful of events in this list not in the same situation?

rvm4:
Oh, I forgot one question/comment. The section on merchantvalidation is not very clear. I assume the context is provided in the payment handler spec, but all the more reason to add that as an informative reference.

rvm4:
1) I get that, but we have failed to do some in the spec itself. If we are forced to reference handlers in order to accurately describe the spec we should give people a place to look for more information. Is there harm in linking payment handlers as an informative reference?

2) Fair enough.

3) V2 it is!

4) Roger.

5) Roger.

6) I'll attempt this in another request.

index.html
marcoscaceres: Ah, we should fix those to make sure they have quotes. 

This is what I mean about the string literals in c++: 
https://searchfox.org/mozilla-central/source/dom/payments/PaymentRequest.cpp#1001

marcoscaceres:
> is there harm in linking payment handlers as an informative reference?

There is no harm, so I’d be open to it. Just needs to be phrased carefully.

Issue #842, opened 2019-02-28
Richer negotiation re: address redaction?


samuelweiler:
As I think about how "billing addresses" are used in practice in the US: sometimes portions of them (e.g. only a postal (ZIP) code) are used by card issuers for validation.  Sometimes merchants want complete addresses for other reasons.  Does it make sense to have a requester explain exactly what address elements it wants, and why?

Consolidated explanation of my concern below at:
https://github.com/w3c/payment-request/issues/842#issuecomment-473907871

ianbjacobs:
Hi @samuelweiler,

Can you say more about the use cases?

Our initial choice of redaction was driven by the use case of "tax computation." At least in version 1, that seems like a good starting point. After consideration of more use cases, we can see whether some merchant/user negotiation would add valuable.

Ian

samuelweiler:
In the US, some merchants, particularly ones that are not shipping a physical product, request (only) a "credit card billing ZIP code", which appears to be used as part of the card acceptance validation.  I'm imagining that it would be convenient - and lead to less transaction abandonment - if the merchant could ask for only the info needed (e.g. just the post code) and not anything else.

I also imagine that different regions have quite different patterns of practices like this, hence the suggestion for a more generalized negotiation mechanism.

npdoty:
I'm not sure user-driven negotiation is particularly useful here (although generally I like that as a strategy) since the user doesn't typically know or set the rules for required data to complete a payment or shipment. But merchant-driven minimization of information would definitely fall under the principle of data minimization, and that can do things like improve user comfort and even encourage payments. It sounded like this feature had been proposed for a future version and maybe Ian could link us to the corresponding issue.

samuelweiler:
npdoty: the problem I see is that merchants often desire more data than is, in fact, required.  Making easier for merchants to get the data (by defaulting to sending it) is a lose for the human sending the payment.

npdoty:
@samuelweiler but are end users in a good position to know what information is required, or to do anything besides refuse payment altogether if the site wants information that might not be necessary?

I think there are some times that negotiation might be a good approach even though it can be complicated to specify and implement, but I think those are cases where users have a lot of insight and there can be progressive levels of functionality based on level of information. But payment seems complex, externally regulated and mostly binary, though perhaps the analog example of haggling over prices would be a non-binary example.

samuelweiler:
[...]

Issue #777, opened 2018-09-12
To address branded button issues and improve privacy, add browser support for payment method selection before the sheet


ianbjacobs:
Hi all,
cc @adimallikarjunareddy

This post is inspired by:
 https://github.com/w3c/payment-request/issues/774

I would like to start discussion around a proposal for adding browser support for payment method selection before the sheet. Perhaps we can discuss at TPAC in October.

It seems to me that canMakePayment() is serving multiple roles. Perhaps we can tease those apart into at least least two needs:

 (1) The merchant wants to know whether to provide a Payment Request API user experience or a traditional Web form.
 (2) Having determined PR API is available, the merchant wants to know which payment buttons to show in a manner consistent with brand-specific requirements. 

Perhaps we can find away to limit canMakePayment() usage to just the first one. This would reduce the number of calls, improving privacy and perhaps avoiding rate limiting issues.

To address the second need, I propose that the browser offer a new user experience.

The browser knows the intersection of the payment methods accepted by the merchant and those for which the user has payment handlers.  Today the browser shows that intersection in the sheet.
What we are hearing is that we need something before the sheet as well.

Therefore, I propose the following (at a high level):

* The browser computes the intersection of merchant-accepted and user-supported payment methods.
* The browser automatically displays that information in the page at a location determined through code by the merchant (implementation details to be determined). However, the set of payment methods displayed must not be shared with the merchant (thus, for example, does not change
the DOM).
* For the display, the browser fetches an icon associated with each payment method (e.g., through the corresponding payment method manifest).
* If the user selects an icon, the browser launches PR API for just that payment method. The browser can optimize (as Chrome already does), showing the sheet if necessary or jumping straight to a single relevant payment handler.

Some advantages to this approach occur to me:

 * The user experience would match user expectations. They expect to see buttons on page, and they do.
 * The merchant would have less work to achieve that user experience. They just call PR API and magic happens to get the right logos.
 * The user experience is optimized for the user. The list of buttons is scoped to the intersection, so that improves on the NASCAR effect.
 * User privacy is protected. The merchant does not receive information about the user's environment before the user has chosen a payment method.
 * The sheet is still relevant (when collecting shipping and contact information).
 * Merchants have flexibility. Merchants could highlight payment buttons they want to highlight, and still use Payment Request this way for other payment methods.
[...]

PR #843, opened 2019-03-01
Changes resulting from 28 February PING privacy review

ianbjacobs:
@rsolomakhin, @danyao, @marcoscaceres,

Based on discussion at the Privacy Interest Group call yesterday, I took an action to propose some changes; see the minutes for details:
 https://www.w3.org/2019/02/28-privacy-minutes

Specifically:

 * Merged the "Privacy and Security Considerations" and "Privacy Considerations" sections into a single section ("Privacy and Security Considerations").

 * Added a forward pointer from 3.5 to the canMakePayment() protections section (in security and privacy considerations). Removed the Note that was in the middle of the algorithm of 3.5, and merged it with the (rewritten) canMakePayment protections section.

 * Expanded the canMakePayment protections section based on PING conversation.

NOTE: I will also point PING at the edit.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/843.html" title="Last updated on Mar 12, 2019, 10:08 PM UTC (fa87e58)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/843/2b23c2f...fa87e58.html" title="Last updated on Mar 12, 2019, 10:08 PM UTC (fa87e58)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5295,7 +5295,7 @@
         <p data-link-for="PaymentRequest">
           The <a>canMakePayment()</a> method enables the payee to determine
           — before calling <a>show()</a> — whether the user agent knows of any <a>payment handlers</a> available to the user that support the  <a>payment methods</a> provided to the <a>PaymentRequest<a> <a data-lt="PaymentRequest.PaymentRequest()">constructor</a>.
-          advantage of the API. This enables the payee to fall back to a legacy
+         If no <a>payment handlers</a> support the <a>payment methods</a>, this enables the payee to fall back to a legacy
           checkout experience. Because this method shares some information with
           the payee, user agents are expected to protect the user from abuse of
           the method. For example, user agents may reduce user fingerprinting
--- a/index.html
+++ b/index.html
@@ -5302,12 +5302,10 @@
           by:
         </p>
         <ul data-link-for="PaymentRequest">
-          <li>allowing the user to configure the user agent to turn off
-          <a>canMakePayment()</a>;
+          <li>Allowing the user to configure the user agent to turn off
+          <a>canMakePayment()</a>.
           </li>
-          <li>informing the user when <a>canMakePayment()</a> is called;
-          </li>
-          <li>rate-limiting the frequency of calls to <a>canMakePayment()</a>
+          <li>Rate-limiting the frequency of calls to <a>canMakePayment()</a>
           with different parameters.
           </li>
         </ul>
@@ -5315,9 +5313,10 @@
           For rate-limiting the user agent might look at repeated calls from:
         </p>
         <ul>
-          <li>the same effective top-level domain plus one (eTLD+1);
+          <li>the same effective top-level domain plus one (eTLD+1).
           </li>
-          <li>the top-level browsing context;
+          <li>the top-level browsing context. Alternatively, the user agent may
+          block access to the API entirely for origins know to be bad actors.
           </li>
           <li>for an iframe, the origin of the iframe content.
           </li>
--- a/index.html
+++ b/index.html
@@ -5318,7 +5318,7 @@
           <li>the top-level browsing context. Alternatively, the user agent may
           block access to the API entirely for origins know to be bad actors.
           </li>
-          <li>for an <a>iframe</a>, the origin of the <a>iframe</a> content.
+          <li>the origin of an <a>iframe</a> or popup window.
           </li>
         </ul>
         <p>
--- a/index.html
+++ b/index.html
@@ -5294,7 +5294,7 @@
         </h2>
         <p data-link-for="PaymentRequest">
           The <a>canMakePayment()</a> method enables the payee to determine
-          —before calling <a>show()</a>— whether the user is ready to take
+          — before calling <a>show()</a> — whether the user agent knows of any <a>payment handlers</a> available to the user that support the  <a>payment methods</a> provided to the <a>PaymentRequest<a> <a data-lt="PaymentRequest.PaymentRequest()">constructor</a>.
           advantage of the API. This enables the payee to fall back to a legacy
           checkout experience. Because this method shares some information with
           the payee, user agents are expected to protect the user from abuse of
--- a/index.html
+++ b/index.html
@@ -5183,7 +5183,7 @@
         </ol>
       </section>
     </section>
-    <section class="informative" id="privacy">
+    <section id="privacy">
       <h2>
         Privacy and Security Considerations
       </h2>
@@ -5263,7 +5263,7 @@
           <a>payment method identifier</a>.
         </p>
       </section>
-      <section class="informative">
+      <section>
         <h2>
           Exposing user information
         </h2>
--- a/index.html
+++ b/index.html
@@ -5305,9 +5305,6 @@
           <li>Allowing the user to configure the user agent to turn off
            <a>canMakePayment()</a>, which would return <a>a promise rejected with</a> a "<a>NotAllowedError</a>" <a>DOMException</a>.
           </li>
-          <li>Informing the user of what data is shared by
-          <a>canMakePayment()</a>;
-          </li>
           <li>Rate-limiting the frequency of calls to <a>canMakePayment()</a>
           with different parameters.
           </li>
--- a/index.html
+++ b/index.html
@@ -5304,7 +5304,6 @@
         <ul data-link-for="PaymentRequest">
           <li>Allowing the user to configure the user agent to turn off
            <a>canMakePayment()</a>, which would return <a>a promise rejected with</a> a "<a>NotAllowedError</a>" <a>DOMException</a>.
-            DOMException</a>.
           </li>
           <li>Informing the user of what data is shared by
           <a>canMakePayment()</a>;
--- a/index.html
+++ b/index.html
@@ -5305,7 +5305,7 @@
           <li>Allowing the user to configure the user agent to turn off
           <a>canMakePayment()</a>.
           </li>
-          <li>informing the user of what data is shared by
+          <li>Informing the user of what data is shared by
           <a>canMakePayment()</a>;
           </li>
           <li>Rate-limiting the frequency of calls to <a>canMakePayment()</a>
--- a/index.html
+++ b/index.html
@@ -5293,17 +5293,22 @@
           <code>canMakePayment()</code> protections
         </h2>
         <p data-link-for="PaymentRequest">
-          The <a>canMakePayment()</a> method enables the payee to determine
-          — before calling <a>show()</a> — whether the user agent knows of any <a>payment handlers</a> available to the user that support the  <a>payment methods</a> provided to the <a>PaymentRequest<a> <a data-lt="PaymentRequest.PaymentRequest()">constructor</a>.
-         If no <a>payment handlers</a> support the <a>payment methods</a>, this enables the payee to fall back to a legacy
-          checkout experience. Because this method shares some potentially unique information with
+          The <a>canMakePayment()</a> method enables the payee to determine —
+          before calling <a>show()</a> — whether the user agent knows of any
+          <a>payment handlers</a> available to the user that support the
+          <a>payment methods</a> provided to the <a>PaymentRequest</a>
+          <a data-lt="PaymentRequest.PaymentRequest()">constructor</a>. If no
+          <a>payment handlers</a> support the <a>payment methods</a>, this
+          enables the payee to fall back to a legacy checkout experience.
+          Because this method shares some potentially unique information with
           the payee, user agents are expected to protect the user from abuse of
           the method. For example, user agents can reduce user fingerprinting
           by:
         </p>
         <ul data-link-for="PaymentRequest">
           <li>Allowing the user to configure the user agent to turn off
-           <a>canMakePayment()</a>, which would return <a>a promise rejected with</a> a "<a>NotAllowedError</a>" <a>DOMException</a>.
+          <a>canMakePayment()</a>, which would return <a>a promise rejected
+          with</a> a "<a>NotAllowedError</a>" <a>DOMException</a>.
           </li>
           <li>Rate-limiting the frequency of calls to <a>canMakePayment()</a>
           with different parameters.
--- a/index.html
+++ b/index.html
@@ -1375,17 +1375,9 @@
           "payment-request/payment-request-canmakepayment-method-protection.https.html">
           Optionally, at the <a>top-level browsing context</a>'s discretion,
           return <a>a promise rejected with</a> a "<a>NotAllowedError</a>" <a>
-            DOMException</a>.
-            <p class="note" data-link-for="PaymentRequest">
-              This allows user agents to apply heuristics to detect and prevent
-              abuse of the <a>canMakePayment()</a> method for fingerprinting
-              purposes, such as creating <a>PaymentRequest</a> objects with a
-              variety of supported <a>payment methods</a> and calling
-              <a>canMakePayment()</a> on them one after the other. For example,
-              a user agent may restrict the number of successful calls that can
-              be made based on the <a>top-level browsing context</a> or the
-              time period in which those calls were made.
-            </p>
+            DOMException</a>. As described in section <a href=
+            "#canmakepayment-protections"></a>, the user agent may limit the
+            rate at which a page can call <a>canMakePayment()</a>.
           </li>
           <li>Let <var>hasHandlerPromise</var> be <a>a new promise</a>.
           </li>
@@ -5191,7 +5183,7 @@
         </ol>
       </section>
     </section>
-    <section class="informative">
+    <section class="informative" id="privacy">
       <h2>
         Privacy and Security Considerations
       </h2>
@@ -5271,12 +5263,7 @@
           <a>payment method identifier</a>.
         </p>
       </section>
-    </section>
-    <section id="privacy">
-      <h2>
-        Privacy Considerations
-      </h2>
-      <section>
+      <section class="informative">
         <h2>
           Exposing user information
         </h2>
@@ -5301,17 +5288,45 @@
           consent.
         </p>
       </section>
-      <section>
-        <h2>
+      <section class="informative">
+        <h2 id="canmakepayment-protections">
           canMakePayment() protections
         </h2>
         <p data-link-for="PaymentRequest">
-          The <a>canMakePayment()</a> method enables the payee to call
-          <a>show()</a> if the user is ready to take advantage of the API, or
-          to fall back to a legacy checkout experience if not. Because this
-          method shares some information with the payee, user agents are
-          expected to protect the user from abuse of the method, for example,
-          by restricting the number or frequency of calls.
+          The <a>canMakePayment()</a> method enables the payee to determine
+          —before calling <a>show()</a>— whether the user is ready to take
+          advantage of the API. This enables the payee to fall back to a legacy
+          checkout experience. Because this method shares some information with
+          the payee, user agents are expected to protect the user from abuse of
+          the method. For example, user agents may reduce user fingerprinting
+          by:
+        </p>
+        <ul data-link-for="PaymentRequest">
+          <li>allowing the user to configure the user agent to turn off
+          <a>canMakePayment()</a>;
+          </li>
+          <li>informing the user when <a>canMakePayment()</a> is called;
+          </li>
+          <li>rate-limiting the frequency of calls to <a>canMakePayment()</a>
+          with different parameters.
+          </li>
+        </ul>
+        <p>
+          For rate-limiting the user agent might look at repeated calls from:
+        </p>
+        <ul>
+          <li>the same effective top-level domain plus one (eTLD+1);
+          </li>
+          <li>the top-level browsing context;
+          </li>
+          <li>for an iframe, the origin of the iframe content.
+          </li>
+        </ul>
+        <p>
+          These rate-limiting techniques intend to increase the cost associated
+          with repeated calls, whether it is the cost of managing multiple
+          eTLDs or the user experience friction of opening multiple windows
+          (tabs or pop-ups).
         </p>
       </section>
     </section>
--- a/index.html
+++ b/index.html
@@ -5305,6 +5305,9 @@
           <li>Allowing the user to configure the user agent to turn off
           <a>canMakePayment()</a>.
           </li>
+          <li>informing the user of what data is shared by
+          <a>canMakePayment()</a>;
+          </li>
           <li>Rate-limiting the frequency of calls to <a>canMakePayment()</a>
           with different parameters.
           </li>
--- a/index.html
+++ b/index.html
@@ -5303,7 +5303,8 @@
         </p>
         <ul data-link-for="PaymentRequest">
           <li>Allowing the user to configure the user agent to turn off
-          <a>canMakePayment()</a>.
+          <a>canMakePayment()</a>, which would cause the user agent to return <a>a promise rejected with</a> a "<a>NotAllowedError</a>" <a>
+            DOMException</a>.
           </li>
           <li>Informing the user of what data is shared by
           <a>canMakePayment()</a>;
--- a/index.html
+++ b/index.html
@@ -5321,7 +5321,7 @@
           <li>the top-level browsing context. Alternatively, the user agent may
           block access to the API entirely for origins know to be bad actors.
           </li>
-          <li>for an iframe, the origin of the iframe content.
+          <li>for an <a>iframe</a>, the origin of the <a>iframe</a> content.
           </li>
         </ul>
         <p>
--- a/index.html
+++ b/index.html
@@ -5298,7 +5298,7 @@
          If no <a>payment handlers</a> support the <a>payment methods</a>, this enables the payee to fall back to a legacy
           checkout experience. Because this method shares some potentially unique information with
           the payee, user agents are expected to protect the user from abuse of
-          the method. For example, user agents may reduce user fingerprinting
+          the method. For example, user agents can reduce user fingerprinting
           by:
         </p>
         <ul data-link-for="PaymentRequest">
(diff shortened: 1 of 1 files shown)


npdoty:
Combining the privacy and security considerations section and the privacy considerations section has one problem: the high-level category is listed as non-normative, but some of the subsections have normative requirements in them. I think you could fix that by removing the marking of the high-level section as informative, and then evaluating each subsection to see whether it's informative or normative.

rsolomakhin (commented):
Addition to "canmakepayment-protections" section LGTM

ianbjacobs:
@npdoty, thanks for catching that. I have adopted your proposal. I do not believe I have changed what was already the case - a single normative section under Security and Privacy Considerations.

marcoscaceres:
Thanks for making a start on this! I’ll try to review it over the weekend.

pes10k:
Thanks much for these @ianbjacobs.  

1) Under mitigations / protections, I'm wondering if there isn't the possibility for greater protections / stronger suggestions here.  For example, is there any plausible any scenario where multiple iframes would need to call the canMakePayment endpoint in the same top-level context?

2) What are the avenues for suggesting / proposing that the mitigations become normative?

3) @samuelweiler had suggestions / concerns about the current proposal and negotiating what payment fields are shared (looks like the [options](https://www.w3.org/TR/payment-request/#the-options-argument) argument would be a natural fit).  Will that be part of these changes?

Thanks much!

ianbjacobs:
HI @snyderp,

> Under mitigations / protections, I'm wondering if there isn't the possibility for greater protections / 
> stronger suggestions here. For example, is there any plausible any scenario where multiple iframes
 > would need to call the canMakePayment endpoint in the same top-level context?

That was one of the suggestions from the call that made it into the pull request: to limit in the face of multiple calls with different parameters from the top-level browsing context. Are there concrete text suggestions you have for making that clearer?

> What are the avenues for suggesting / proposing that the mitigations become normative?

We can continue to discuss them on this thread. I think if the implementers agree that a given mitigation should be standardized, then I expect we can add it. However, I would not expect that to happen in version 1.

>  @samuelweiler had suggestions / concerns about the current proposal and negotiating what 
> payment fields are shared (looks like the options argument would be a natural fit). Will that be
> part of these changes?

I would not expect negotiation of which data is shared to be a version 1 feature of Payment Request API. That has not been part of the specification to date and I think we would need to have more discussion of real-world payment systems. Sam has opened a new issue where have started discussion:
 https://github.com/w3c/payment-request/issues/842

A related topic (outside of Payment Request API) involves device fingerprinting (through JavaScript and other mechanisms). Our ongoing discussions in the working group involve topics like how to manage user privacy preferences with demands for data from risk engines. 

Ian

index.html
```diff
@@ -5301,17 +5288,45 @@ <h2>
...
+          checkout experience. Because this method shares some information with
+          the payee, user agents are expected to protect the user from abuse of
+          the method. For example, user agents may reduce user fingerprinting
+          by:
+        </p>
+        <ul data-link-for="PaymentRequest">
+          <li>allowing the user to configure the user agent to turn off
+          <a>canMakePayment()</a>;
+          </li>
+          <li>informing the user when <a>canMakePayment()</a> is called;
```
marcoscaceres: This seem impractical. I'm against including this suggestion.

index.html
```diff
@@ -5301,17 +5288,45 @@ <h2>
...
+          —before calling <a>show()</a>— whether the user is ready to take
+          advantage of the API. This enables the payee to fall back to a legacy
+          checkout experience. Because this method shares some information with
+          the payee, user agents are expected to protect the user from abuse of
+          the method. For example, user agents may reduce user fingerprinting
+          by:
+        </p>
+        <ul data-link-for="PaymentRequest">
+          <li>allowing the user to configure the user agent to turn off
+          <a>canMakePayment()</a>;
```
```suggestion
          <a>canMakePayment()</a>.
```

index.html
```diff
@@ -5301,17 +5288,45 @@ <h2>
...
+          </li>
+          <li>informing the user when <a>canMakePayment()</a> is called;
+          </li>
+          <li>rate-limiting the frequency of calls to <a>canMakePayment()</a>
```
```suggestion
          <li>Rate-limiting the frequency of calls to <a>canMakePayment()</a>
```

index.html
```diff
@@ -5301,17 +5288,45 @@ <h2>
...
+          with different parameters.
+          </li>
+        </ul>
+        <p>
+          For rate-limiting the user agent might look at repeated calls from:
+        </p>
+        <ul>
+          <li>the same effective top-level domain plus one (eTLD+1);
```
```suggestion
          <li>the same effective top-level domain plus one (eTLD+1).
```

index.html
```diff
@@ -5301,17 +5288,45 @@ <h2>
...
+          </li>
+          <li>the top-level browsing context;
```
```suggestion
          <li>the top-level browsing context - or block access to the API entirely for origins know to be bad actors.</li> 
```

pes10k:
@ianbjacobs 
> That was one of the suggestions from the call that made it into the pull request: to limit in the face of multiple calls with different parameters from the top-level browsing context. Are there concrete text suggestions you have for making that clearer?

I see the "For example, a user agent may restrict the number of successful calls that can be made based on the top-level browsing context" text.  I'm not sure I understand the "successful calls" text there.  Whats an "unsuccessful" call?  If "unsuccessful" means `canMakePayment` returns false, then merely rate-limiting on "successful" calls won't do much to reduce its finger-print-ability.  Apologies if I'm miss understanding but could you explain / clarify whats meant here?

> A related topic (outside of Payment Request API) involves device fingerprinting (through JavaScript and other mechanisms). Our ongoing discussions in the working group involve topics like how to manage user privacy preferences with demands for data from risk engines.

Great!  Where is the best place to stay up to date on these discussions?

ianbjacobs:
Hi @snyderp,

I'll start by saying that this topic has been a challenging one for the Working Group, and new ideas are welcome. We are relying on implementaitons to help mitigate abuse while at the same time responding to merchant requirements to be able to decide what checkout experience they can provide that will afford the least friction.

The particular text that you are referring to no longer exists in my pull request. However, it may be that you have the same question over the new text. 

I invite @marcoscaceres, @rsolomakhin, @danyao, and @aestes to weigh in on their implementation experience.

There have been other ideas on this topic that have not gained traction for other reasons; see #777 for example.

Regarding data gathering for risk analysis, that seems to lie partly within the scope of several W3C groups. The Web Payments WG has had some discussions with EMVCo about 3-D Secure, which gathers some data. We've also chatted with the Web Authentication Working Group about whether WebAuthn (e.g., attestations) might be helpful in reducing reliance on JavaScript. And FIDO and EMVCo are also discussing that topic.

To help facilitate some of these conversations we have been discussing the creation of a W3C Interest Group on Web payment security. A draft charter was reviewed by the W3C Membership, EMVCo Board, and FIDO Board; we are currently addressing feedback. 

In short: I would like to make it easier to discover and participate in Web payment security topics. I hope this proposed IG can help fulfill that goal.

marcoscaceres (changes requested):
FWIW, rate limiting still feels ineffectual.

pes10k:
@ianbjacobs 

Thanks again for the reply.  Sure enough, i was looking at the wrong text.  Thanks for the clarification!

One other item related to mitigations, and one more reason you may want to include mitigation standardization in v1 (even if it slows things down): how does the user agent signal to the website "I am interested in using the Payment API, but I don't want to allow the fingerprinting risk implied by canMakePayment."?

Say the user agent is operating in this privacy-protecting state and takes some of the mitigation steps mentioned in the text, and turns `canMakePayment` off (e.g. "allowing the user to configure the user agent to turn off").  Is there some way the UA can say "ask me explicitly what I support, by sending me a form, or similar, but no passive techniques"?  I dont see anyway to express this in the API as described.

Throwing `SecurityError` seems semantically wrong (there is no security related issue in this scenario).  The `NotAllowedError` is only specified for the rate limit case.  And in either case, the website might reasonably interpret the response to be "UA never allows using Web Payment API" or "i've been rate limited", instead of "UA doesn't allow the website to ask what payment methods are supported, passively."  In which case, the `canMakePayment` functionality, as currently imagined might _increase_ friction (since the site could assume Web Payments is disabled).

I suggest adding an additional exception type, or return value to the `canMakePayment` promise indicating "user is still interested in Web Payments, but no background / passive checking, to maximize use of Web Payments API.  Thoughts?

marcoscaceres:
> I suggest adding an additional exception type, or return value to the canMakePayment promise indicating "user is still interested in Web Payments, but no background / passive checking, to maximize use of Web Payments API. Thoughts?

The use case remains that "can make payment?" is intended to be used to present, for example, an Google Pay specific button directly in the web page. Asking the user is not practical, because the site is effectively "port scanning" what supported payment methods are.  

I think the only feasible solution here is for the user to have control over this (like Safari does). Basically, an opt-in this check: 

> "[ ] Tell sites I support "basic-card" (when not in Private Browsing mode)."

marcoscaceres:
This is how Safari works: 

screenshot 2019-03-05 16 12 37

pes10k:
@marcoscaceres

I dont we're disagreeing.  The correct answer is definitely to prompt, and to effectively not allow anything like how `canMakePayment` is designed.  We discussed this on the PING call. 

But, whether or not `canMakePayment` comes out or always leads to a prompt, or whatever else, i'm saying there needs to be more granularity in the reason the user is not returning accurate info on `canMakePayment`.  That there needs to be a way to explicitly say to the site "i'm not going to tell you the payment methods I support passively, but I would still like to use the Web Payments API checkout flow."

marcoscaceres:
> That there needs to be a way to explicitly say to the site "i'm not going to tell you the payment methods I support passively, but I would still like to use the Web Payments API checkout flow."

heh, returns "maybe" :) But seriously, it's why in Firefox we always lie and just return true.

pes10k:
@marcoscaceres that seems like a reasonable approach, but suggests it shouldn't be in the standard in the first place :)

Brave will very likely do something similar.  Is anyone familiar with what Safari is planning on doing for non-Apple pay methods?

aestes:
> Is anyone familiar with what Safari is planning on doing for non-Apple pay methods?

Currently, Safari's only supported payment method is Apple Pay.

rsolomakhin:
@snyderp wrote:

> Whats an "unsuccessful" call?

Rejecting `canMakePayment` with `NotAllowedError` and a message `Exceeded query quota`.

pes10k:
> Currently, Safari's only supported payment method is Apple Pay.

Thanks!  I should have asked differently though: Does anyone know who, other than Chrome, plans to implement the spec / `canMakePayment` as described in the document?  Sounds like definitely not Firefox, definitely not Brave, and _maybe_ not Apple?

rsolomakhin:
@snyderp : Which part of the spec differs among the implementations?

From https://w3c.github.io/payment-request/#canmakepayment-method :

> If the user agent has a payment handler that supports handling payment requests for identifier, resolve hasHandlerPromise with true and terminate this algorithm.

Firefox has a built-in payment handler for "basic-card", so they will always return "true". @marcoscaceres, correct me if I'm wrong. So Firefox implementation seems to match to spec, right?

ianbjacobs:
Adding @romandev and @zouhir (and assuming @rsolomakhin and @danyao already watching).

Ian

rsolomakhin:
@snyderp wrote:

> I suggest adding an additional exception type, or return value to the canMakePayment promise indicating "user is still interested in Web Payments, but no background / passive checking, to maximize use of Web Payments API. Thoughts?

I like this idea. Our current behavior in this case is to return "false" in `canMakePayment()`, but a different error message would definitely be useful.

ianbjacobs:
@snyderp and @rsolomakhin, would you support creating a new issue on the topic of finer-grain error reporting when canMakePayment() returns "false"? 

Some observations (which I can carry over to a new issue as well).
 - An error message (e.g., "okToUseAPI") that indicates that the user does not want to share detailed information but would like to use the API under show() conditions would reveal some information about the user's browser configuration. For consideration: is that an improvement?
- @rsolomakhin, you mentioned fine-grain error reporting for debugging. What are the other scenarios that you (or Google Pay) team would like to reflect in these messages?

Ian

rsolomakhin:
@ianbjacobs : A separate issue sounds good. Does this need to go into the spec or best practices for implementers? I understand the error codes go into spec. Not sure about error messages.

Other scenarios that would benefit any payment handler or merchant:

1. Errors in SSL. For example, self-signed SSL causes Chrome to always return "false" to canMakePayment() and reject show() with "AbortError: Request cancelled."
1. Reasons why the payment sheet closed, e.g., "AbortError: User closed the payment handler", "AbortError: The payment handler cancelled the transaction", or "AbortError: User closed the card CVV entry form."

pes10k:
> @snyderp : Which part of the spec differs among the implementations?

@rsolomakhin I was basing what I said (that Firefox isn't implementing the spec as written) based on @marcoscaceres comment "But seriously, it's why in Firefox we always lie and just return true.". I take that to mean they don't follow the `supportedNetworks` part of the algorithm. 

Brave will likely do the same (decision hasn't been 100% made yet) and return true no matter what, at least with "Shields Up".

rsolomakhin:
@snyderp : Where in the `canMakePayment` algorithm are `supportedNetworks` mentioned? I was under the impression that current the spec says that Chrome, Firefox, and Edge should always return `true` to `canMakePayment()` query for `basic-card`, while Safari should always return `true` for `https://apple.com/apple-pay`. (Chrome is in compliance starting at version 74.)

rsolomakhin:
@snyderp wrote:

> Brave will likely do the same (decision hasn't been 100% made yet) and return true no matter what, at least with "Shields Up".

Sounds good. I believe that's in compliance with the spec and that's how the other implementers are behaving for `canMakePayment`, too, when the query is for a built-in payment handler, such as `basic-card`, or Apple Pay.

index.html
zouhir: after reading the privacy minutes, I actually think this is an OK idea and npdoty questions was good.

Pros:
- Website are likely to be calling this method more responsibly if there is a little UI indication to the users and only do so in the right page and time where users expect this to happen and checkout.

Cons:
- Some users might freak out if they see such a UI notification or indication. there has to be a lot of carefulness and UX research from implementers side how to present this.

Why does it seem impractical?

pes10k: I'm in favor of it too.  Short of standardizing the mitigations (which would be better, but seems to be off the table), including the above seems like a second best option

pes10k:
@rsolomakhin 

> Where in the canMakePayment algorithm are supportedNetworks mentioned?

Ah, I see my error here.  [3.5](https://w3c.github.io/payment-request/#dom-paymentrequest-canmakepayment) says `supportedNetworks` doesn't matter.  But in the definition of the `show()` algorithm, in step 15.6.1 it says `canMakePayment` is the result of [checking the result of each handler](https://w3c.github.io/payment-request/#dfn-steps-to-check-if-a-payment-can-be-made) (which would include [`supportedNetworks`](https://www.w3.org/TR/payment-method-basic-card/#steps-to-check-if-a-payment-can-be-made)).

So, 100% my error, but would be good to rename `canMakePayment` the variable in the [`show()`](https://w3c.github.io/payment-request/#show-method) algorithm, to something else, to avoid confusion with the `canMakePayment` method.  `isHandlerSupported`? Thoughts?

This is even more confusing because the `canMakePayment` note says "[canMakePayment checks if] the user agent supports any of the desired payment methods".  When I click on `payment methods`, i'm taken to a link that says "the party that provides the means (e.g., credit card) that the payer uses to pay". Since "Visa" and "Mastercard" are parties, but "a credit card" is not a party, the "NOTE" text seems to suggest the `canMakePayment` text is describing "does the user agent have a visa card set up" and not  "does the user agent have a credit card" set up. 

On looking closer, i see that "payment method" is in bold, followed by provider (not in bold), so I'm left even more confused whether this is a definition of a payment method, or a provider of a payment method (and if so, why was I was linked here? and I still don't know what a payment method is).

I suggest cleaning all this up, and explicitly defining more terms.  I see that I've misunderstood what `canMakePayment` (method, not the variable) is intended to do.  But I hope I've shown why the document isn't helping ;)

pes10k:
@rsolomakhin @ianbjacobs 

Just so that my understanding is clear, can someone describe what `canMakePayment` should return in each of the following scenarios:

1. I'm using Firefox, I haven't ever used any Credit Card info in FF (i.e. it doesn't know if I have a visa).  Website calls `canMakePayment` with credit-card and visa.
2. Same as above, but I've disabled Payment Request API (`dom.payments.request.enabled = false`)
3. I'm using Firefox, but `supportedMethods = "apple-pay".

@marcoscaceres just trying to understand what you mean by "FF lies…".  Can you describe what you meant there?

rsolomakhin:
@pes wrote:

> @rsolomakhin, Just so that my understanding is clear, can someone describe what `canMakePayment` should return in each of the following scenarios:
> 
> 1. I'm using Firefox, I haven't ever used any Credit Card info in FF (i.e. it doesn't know if I have a visa).  Website calls `canMakePayment` with credit-card and visa.

According to the spec, FF should return `true` for `canMakePayment` here, because the built-in `basic-card` payment handler is available for use.

> 2. Same as above, but I've disabled Payment Request API (`dom.payments.request.enabled = false`)

Not familiar with `dom.payments.request.enabled`. @marcoscaceres ?

> 3. I'm using Firefox, but `supportedMethods = "apple-pay".

According to the spec, FF should return `false` for `canMakePayment` here.

marcoscaceres:
@snyderp, @rsolomakhin 
> According to the spec, FF should return false for canMakePayment here.

Yes, you are correct - sorry... I wasn't very clear and didn't give enough context in my response above. The original `canMakePayment()` worked in the same way as the proposed `hasEnrolledInstrument()` - that is, "supports PMI AND has enrolled an instrument".  We used to lie about having the enrolled instrument (because it further adds to finger printing)... 

So, basically we were returning `true` for only known PMIs (of which Firefox knows only about "basic-card"). 

So, @snyderp, to answer your questions. 

> I'm using Firefox, I haven't ever used any Credit Card info in FF (i.e. it doesn't know if I have a visa). Website calls canMakePayment with credit-card and visa.

True. 

> Same as above, but I've disabled Payment Request API (dom.payments.request.enabled = false)

False. 

Note `dom.payments.request.enabled` would literally turns off the entire API... so rather, you would get a "`canMakePayment()` is no a function" JS exception. 

> I'm using Firefox, but `supportedMethods = "apple-pay".

False.

rsolomakhin:
@snyderp wrote:

> I suggest cleaning all this up, and explicitly defining more terms. I see that I've misunderstood what canMakePayment (method, not the variable) is intended to do.

SGTM

index.html
```diff
@@ -5301,17 +5288,47 @@ <h2>
...
+        <p>
+          For rate-limiting the user agent might look at repeated calls from:
+        </p>
+        <ul>
+          <li>the same effective top-level domain plus one (eTLD+1).
+          </li>
+          <li>the top-level browsing context. Alternatively, the user agent may
+          block access to the API entirely for origins know to be bad actors.
+          </li>
+          <li>for an iframe, the origin of the iframe content.
```
```suggestion
          <li>for an <a>iframe</a>, the origin of the <a>iframe</a> content.
```

index.html
```diff
@@ -5301,17 +5288,47 @@ <h2>
           consent.
         </p>
       </section>
-      <section>
-        <h2>
+      <section class="informative">
+        <h2 id="canmakepayment-protections">
           canMakePayment() protections
```
```suggestion
          <code>canMakePayment()</code> protections
```

index.html
```diff
@@ -5301,17 +5288,47 @@ <h2>
...
         </h2>
         <p data-link-for="PaymentRequest">
-          The <a>canMakePayment()</a> method enables the payee to call
-          <a>show()</a> if the user is ready to take advantage of the API, or
-          to fall back to a legacy checkout experience if not. Because this
-          method shares some information with the payee, user agents are
-          expected to protect the user from abuse of the method, for example,
-          by restricting the number or frequency of calls.
+          The <a>canMakePayment()</a> method enables the payee to determine
+          —before calling <a>show()</a>— whether the user is ready to take
```
```suggestion
          — before calling <a>show()</a> — whether the user agent knows of any <a>payment handlers</a> available to the user that support the  <a>payment methods</a> provided to the <a>PaymentRequest<a> <a data-lt="PaymentRequest.PaymentRequest()">constructor</a>.
```

marcoscaceres: Should be: ... whether the user agent supports the payment handlers with which the payment request was constructed.

index.html
```diff
@@ -5301,17 +5288,47 @@ <h2>
...
+          advantage of the API. This enables the payee to fall back to a legacy
```
```suggestion
         If no <a>payment handlers</a> support the <a>payment methods</a>, this enables the payee to fall back to a legacy
```

index.html
```diff
@@ -5301,17 +5288,47 @@ <h2>
...
+          checkout experience. Because this method shares some information with
```
ianbjacobs:
```suggestion
          checkout experience. Because this method shares some potentially unique information with
```

index.html
```diff
@@ -5301,17 +5288,47 @@ <h2>
...
+          the payee, user agents are expected to protect the user from abuse of
+          the method. For example, user agents may reduce user fingerprinting
+          by:
+        </p>
+        <ul data-link-for="PaymentRequest">
+          <li>Allowing the user to configure the user agent to turn off
+          <a>canMakePayment()</a>.
+          </li>
+          <li>informing the user of what data is shared by
```
ianbjacobs:
```suggestion
          <li>Informing the user of what data is shared by
```

index.html
```diff
@@ -5301,17 +5288,47 @@ <h2>
...
-          to fall back to a legacy checkout experience if not. Because this
-          method shares some information with the payee, user agents are
-          expected to protect the user from abuse of the method, for example,
-          by restricting the number or frequency of calls.
+          The <a>canMakePayment()</a> method enables the payee to determine
+          — before calling <a>show()</a> — whether the user agent knows of any <a>payment handlers</a> available to the user that support the  <a>payment methods</a> provided to the <a>PaymentRequest<a> <a data-lt="PaymentRequest.PaymentRequest()">constructor</a>.
+         If no <a>payment handlers</a> support the <a>payment methods</a>, this enables the payee to fall back to a legacy
+          checkout experience. Because this method shares some information with
+          the payee, user agents are expected to protect the user from abuse of
+          the method. For example, user agents may reduce user fingerprinting
```
```suggestion
          the method. For example, user agents can reduce user fingerprinting
```

index.html
ianbjacobs:
```suggestion
          <li>Informing the user of what data is shared by
```

index.html
```diff
@@ -5301,17 +5288,47 @@ <h2>
...
+          —before calling <a>show()</a>— whether the user is ready to take
+          advantage of the API. This enables the payee to fall back to a legacy
+          checkout experience. Because this method shares some information with
+          the payee, user agents are expected to protect the user from abuse of
+          the method. For example, user agents may reduce user fingerprinting
+          by:
+        </p>
+        <ul data-link-for="PaymentRequest">
+          <li>Allowing the user to configure the user agent to turn off
+          <a>canMakePayment()</a>.
```
ianbjacobs:
```suggestion
          <a>canMakePayment()</a>, which would return <a>a promise rejected with</a> a "<a>NotAllowedError</a>" <a>DOMException</a>.
```

index.html
```diff
@@ -5301,17 +5288,48 @@ <h2>
...
+          — before calling <a>show()</a> — whether the user agent knows of any <a>payment handlers</a> available to the user that support the  <a>payment methods</a> provided to the <a>PaymentRequest<a> <a data-lt="PaymentRequest.PaymentRequest()">constructor</a>.
+         If no <a>payment handlers</a> support the <a>payment methods</a>, this enables the payee to fall back to a legacy
+          checkout experience. Because this method shares some potentially unique information with
+          the payee, user agents are expected to protect the user from abuse of
+          the method. For example, user agents can reduce user fingerprinting
+          by:
+        </p>
+        <ul data-link-for="PaymentRequest">
+          <li>Allowing the user to configure the user agent to turn off
+          <a>canMakePayment()</a>, which would cause the user agent to return <a>a promise rejected with</a> a "<a>NotAllowedError</a>" <a>
```
```suggestion
           <a>canMakePayment()</a>, which would return <a>a promise rejected with</a> a "<a>NotAllowedError</a>" <a>DOMException</a>.
```

index.html
```diff
@@ -5301,17 +5288,48 @@ <h2>
...
+         If no <a>payment handlers</a> support the <a>payment methods</a>, this enables the payee to fall back to a legacy
+          checkout experience. Because this method shares some potentially unique information with
+          the payee, user agents are expected to protect the user from abuse of
+          the method. For example, user agents can reduce user fingerprinting
+          by:
+        </p>
+        <ul data-link-for="PaymentRequest">
+          <li>Allowing the user to configure the user agent to turn off
+           <a>canMakePayment()</a>, which would return <a>a promise rejected with</a> a "<a>NotAllowedError</a>" <a>DOMException</a>.
+            DOMException</a>.
```
```suggestion

```

index.html
```diff
@@ -5301,17 +5288,47 @@ <h2>
...
+        <p>
+          For rate-limiting the user agent might look at repeated calls from:
+        </p>
+        <ul>
+          <li>the same effective top-level domain plus one (eTLD+1).
+          </li>
+          <li>the top-level browsing context. Alternatively, the user agent may
+          block access to the API entirely for origins know to be bad actors.
+          </li>
+          <li>for an <a>iframe</a>, the origin of the <a>iframe</a> content.
```
```suggestion
          <li>the origin of an <a>iframe</a> or popup window.
```

marcoscaceres (approved):

aestes:
> This is how Safari works:
> 
> screenshot 2019-03-05 16 12 37

I should clarify something here. This option influences the return value of the [canMakePaymentsWithActiveCard API](https://developer.apple.com/reference/applepayjs/applepaysession/1778000-canmakepaymentswithactivecard) in Apple Pay JS (and soon the hasEnrolledInstrument API in Payment Request).

When that option is unchecked (or the window is in Private Browsing), canMakePaymentsWithActiveCard will behave as if canMakePayments were called instead, returning true whenever Apple Pay is supported even if there are no enrolled instruments.

Neither this checkbox nor Private Browsing influence the result of canMakePayments in Apple Pay JS or canMakePayment in Payment Request.

PR #873, opened 2019-06-28
Allow incremental request of billing and shipping address

### LINKED ISSUE #842: Richer negotiation re: address redaction?

danyao:
closes #842

Add the ability for the merchant to request fine-grained parts of billing address and shipping address. This is based on @adrianhopebailie's idea from https://github.com/w3c/payment-method-basic-card/issues/72#issuecomment-481275334.

We are making these changes because merchants (e.g. Netflix) has requested the ability to receive less information (e.g. Netflix).

At a high level, this pull request can be summarized as the following:

- The existing `requestBillingAddress` and `requestShipping` booleans of `PaymentOptions` will live on to provide backward compatibility:
  - The new `requestBillingAddressParts` and `requestShippingAddressParts` will only take effect when the respective boolean flag is true.
  - This provides backward compatibility:
    - old code + new browser: booleans are used, lists default to [] => old behavior
    - new code + old browser: booleans are used, lists are ignored => old behavior
    - new code + new browser: booleans and lists are both considered => new behavior
- `requestBillingAddressParts` and `requestShippingAddressParts` affect both intermediate results (e.g. those returned in `PaymentRequestUpdateEvent` and `PaymentMethodChangeEvent`) and final results in `PaymentResponse`
  - For billing address, the responsibility is on the payment handler. The specifics is left to the payment method specific specs (e.g. Basic Card and Payment Handler API).
  - For shipping address, a new `[[requestedShippingAddressParts]]` internal slot is added to `PaymentRequest` to keep track of the cumulative address parts requested.
    - “Create a PaymentAddress from user-provided input” algorithm is updated to take an additional `requestedAddressParts` argument that callers will set using the new internal slot.
  - A payee that desires full address at the end must explicitly call `requestBillingAddress([])` and/or `requestShippingAddress([])`.
- Implication on event firing:
  - `shippingaddresschange` event (type: `PaymentRequestUpdateEvent`) must be fired at least once if `requestShippingAddressParts` is non-empty to give payee an opportunity to request full shipping address.
  - `paymentmethodchange` event (type: `PaymentMethodChangeEvent`) must be fired at least once if `requestBillingAddressParts` is non-empty to give payee an opportunity to request full billing address.

Implementation commitment:

 * [ ] Safari (link to issue)
 * [ ] Chrome (link to issue)
 * [ ] Firefox (link to issue)
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/danyao/payment-request/pull/873.html" title="Last updated on Sep 14, 2020, 8:37 PM UTC (5f5bbc1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/873/c289211...danyao:5f5bbc1.html" title="Last updated on Sep 14, 2020, 8:37 PM UTC (5f5bbc1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -891,7 +891,7 @@
           <li>Set |request|.<a>[[\options]]</a> to |options|.
           </li>
           <li>Set |request|.<a>[[\requestedShippingAddressParts]]</a> to
-          |options|.<a data-lt="PaymentOptions.requestShippingAddressParts">requestShippingAddressParts</a>.
+          |options|.{{PaymentOptions/requestShippingAddressParts}}.
           <li>Set |request|.<a>[[\state]]</a> to "<a>created</a>".
           </li>
           <li>Set |request|.<a>[[\updating]]</a> to false.
--- a/index.html
+++ b/index.html
@@ -2212,7 +2212,7 @@
           parts of the billing address that the payee wishes to receive when the
           <a>requestBillingAddress</a> member is true. An empty list means no
           restriction, i.e. the payee requests the full billing address. This
-          field has no effect if the <a>requestBillingAddress</a> boolean is
+          field has no effect if the <a>requestBillingAddress</a> member is
           false.
           <aside class="note">
             The billing address is part of the payment-method specific response.
--- a/index.html
+++ b/index.html
@@ -2199,7 +2199,7 @@
           method</a> (e.g., the billing address associated with a credit card).
           The <a>user agent</a> returns the billing address as part of the
           {{PaymentMethodChangeEvent}}'s {{PaymentMethodChangeEvent/methodDetails}} attribute and the
-          {{PaymentResponse}}'s <a>details</a> attribute. A payee can use this
+          {{PaymentResponse}}'s {{PaymentResponse/details}} attribute. A payee can use this
           information to, for example, calculate tax in certain jurisdictions
           and update the displayed total. See below for privacy considerations
           regarding <a href="#user-info">exposing user information</a>.
--- a/index.html
+++ b/index.html
@@ -2265,7 +2265,7 @@
           attribute and the {{PaymentResponse}}'s
           {{PaymentResponse/shippingAddress}}
           attribute. The former is subject to a
-          <a data-lt="shipping-redact-list">redact list</a> per the
+          <a>shipping redact list</a> per the
           <a>shipping address changed algorithm</a>.
           <aside class="note">
             <p>
--- a/index.html
+++ b/index.html
@@ -4081,7 +4081,7 @@
           </h2>
           <aside class="note">
             <p>
-              <a>PaymentMethodChangeEvent.requestBillingAddress()</a> allows the
+              The {{PaymentMethodChangeEvent/requestBillingAddress()}} method allows the
               developer to incrementally request additional pieces of the
               billing address that were not initially requested in
               {{ PaymentOptions.requestBillingAddressParts }}.
--- a/index.html
+++ b/index.html
@@ -4076,7 +4076,7 @@
               If a developer wants to incrementally request more granular details
               of the user's shipping address, for example, to re-calculate
               shipping cost, then they can call <a>requestShippingAddress()</a>
-              and provide a list of <a>AddressParts</a>, which will return a
+              and provide a sequence of <a>AddressParts</a>, which will return a
               promise that resolves to an <a>PaymentAddress</a> with the
               requested address fields populated. This method can be called
               repeatedly to incrementally request more parts as needed.
--- a/index.html
+++ b/index.html
@@ -2208,7 +2208,7 @@
           <dfn>requestBillingAddressParts</dfn> member
         </dt>
         <dd>
-          A sequence of <a>AddressParts</a> that specifies which fine-grained
+          A sequence of {{AddressParts}} that specifies which fine-grained
           parts of the billing address that the payee wishes to receive when the
           <a>requestBillingAddress</a> boolean is true. An empty list means no
           restriction, i.e. the payee requests the full billing address. This
--- a/index.html
+++ b/index.html
@@ -1503,7 +1503,7 @@
               the payee-requested shipping address parts, initially supplied to
               the constructor via {{PaymentOptions}}'s
               {{PaymentOptions/requestShippingAddressParts}} and then updated with calls to
-              requestShippingAddress()</a>.
+              {{PaymentRequestUpdateEvent/requestShippingAddress()}}.
             </td>
           </tr>
           <tr>
--- a/index.html
+++ b/index.html
@@ -2287,7 +2287,7 @@
           restriction, i.e. the <a>user agent</a> will return the full shipping
           address, subject to the
           <a>shipping redact list</a>. This field has no effect if the
-          <a>requestShipping</a> member is false. The payee can request
+          {{PaymentOptions/requestShipping}} member is false. The payee can request
           additional pieces incrementally using {{PaymentRequestUpdateEvent}}'s
           {{PaymentRequestUpdateEvent/requestShippingAddress()}}.
         </dd>
--- a/index.html
+++ b/index.html
@@ -1502,7 +1502,7 @@
               The current <a>sequence</a>&lt;{{AddressParts}}&gt; specifying
               the payee-requested shipping address parts, initially supplied to
               the constructor via {{PaymentOptions}}'s
-              requestShippingAddressParts</a> and then updated with calls to
+              {{PaymentOptions/requestShippingAddressParts}} and then updated with calls to
               <a data-lt="PaymentRequestUpdateEvent.requestShippingAddress">
               requestShippingAddress()</a>.
             </td>
--- a/index.html
+++ b/index.html
@@ -2261,7 +2261,7 @@
           when physical goods need to be shipped by the payee to the user). The
           <a>user agent</a> returns the shipping address in the
           {{PaymentRequest}}'s {{PaymentRequest/shippingAddress}}
-          shippingAddress</a> attribute and the {{PaymentResponse}}'s
+          attribute and the {{PaymentResponse}}'s
           <a data-lt="PaymentResponse.shippingAddress">shippingAddress</a>
           attribute. The former is subject to a redact list per the <a>shipping
           address changed algorithm</a>.
--- a/index.html
+++ b/index.html
@@ -4087,7 +4087,7 @@
               billing address that were not initially requested in
               {{ PaymentOptions.requestBillingAddressParts }}.
             </p>
-            <pre class="example js" title="how to use `requestBillingAddress()` correct.">
+            <pre class="example js" title="how to use `requestBillingAddress()`.">
               request.onpaymentmethodchange = ev =&gt; {
                 const promiseForUpdatingTax =
                   ev.requestBillingAddress(["country", "postalCode"])
--- a/index.html
+++ b/index.html
@@ -2984,7 +2984,7 @@
           The steps to <dfn>create a `PaymentAddress` from user-provided
           input</dfn> are given by the following algorithm. The algorithm takes
           a <a>list</a> |redactList| and a
-          <a>sequence</a>&lt;<a>AddressParts</a>&gt; |requestedParts|.
+          <a>sequence</a>&lt;{{AddressParts}}&gt; |requestedParts|.
         </p>
         <div class="note" title=
         "Privacy of recipient information (the redactList)">
--- a/index.html
+++ b/index.html
@@ -2262,7 +2262,7 @@
           <a>user agent</a> returns the shipping address in the
           {{PaymentRequest}}'s {{PaymentRequest/shippingAddress}}
           attribute and the {{PaymentResponse}}'s
-          <a data-lt="PaymentResponse.shippingAddress">shippingAddress</a>
+          {{PaymentResponse/shippingAddress}}
           attribute. The former is subject to a redact list per the <a>shipping
           address changed algorithm</a>.
           <aside class="note">
--- a/index.html
+++ b/index.html
@@ -4037,7 +4037,7 @@
           [SecureContext, Exposed=Window]
           interface PaymentRequestUpdateEvent : Event {
             constructor(DOMString type, optional PaymentRequestUpdateEventInit eventInitDict = {});
-            Promise&lt;PaymentAddress&gt; requestShippingAddress(AddressParts addressParts);
+            Promise&lt;PaymentAddress&gt; requestShippingAddress(sequence<AddressParts> addressParts);
             void updateWith(Promise&lt;PaymentDetailsUpdate&gt; detailsPromise);
           };
         </pre>
--- a/index.html
+++ b/index.html
@@ -892,6 +892,7 @@
           </li>
           <li>Set |request|.<a>[[\requestedShippingAddressParts]]</a> to
           |options|.{{PaymentOptions/requestShippingAddressParts}}.
+          </li>
           <li>Set |request|.<a>[[\state]]</a> to "<a>created</a>".
           </li>
           <li>Set |request|.<a>[[\updating]]</a> to false.
@@ -1140,7 +1141,8 @@
             <p>
               Pass the <a data-lt="converted to an IDL value">converted</a>
               second element in the |paymentMethod| tuple, |modifiers|,
-              |request|.<a>[[\options]]</a>.{{PaymentOptions/requestBillingAddress}} and
+              |request|.<a>[[\options]]</a>.{{PaymentOptions/requestBillingAddress}}
+              and
               |request|.<a>[[\options]]</a>.{{PaymentOptions/requestBillingAddressParts}}.
               Optionally, the user agent SHOULD send the appropriate data from
               |request| to the user-selected <a>payment handler</a> in order to
@@ -1502,7 +1504,8 @@
               The current <a>sequence</a>&lt;{{AddressParts}}&gt; specifying
               the payee-requested shipping address parts, initially supplied to
               the constructor via {{PaymentOptions}}'s
-              {{PaymentOptions/requestShippingAddressParts}} and then updated with calls to
+              {{PaymentOptions/requestShippingAddressParts}} and then updated
+              with calls to
               {{PaymentRequestUpdateEvent/requestShippingAddress()}}.
             </td>
           </tr>
@@ -1744,8 +1747,8 @@
           the developer that the currency is invalid.
           </li>
           <li>If |amount|.<a>value</a> is not a <a>valid decimal monetary
-          value</a>, throw a {{TypeError}}, optionally informing the
-          developer that the currency is invalid.
+          value</a>, throw a {{TypeError}}, optionally informing the developer
+          that the currency is invalid.
           </li>
           <li>Set |amount|.<a>currency</a> to the result of <a>ASCII
           uppercase</a> |amount|.<a>currency</a>.
@@ -2198,30 +2201,31 @@
           and return the billing address associated with a <a>payment
           method</a> (e.g., the billing address associated with a credit card).
           The <a>user agent</a> returns the billing address as part of the
-          {{PaymentMethodChangeEvent}}'s {{PaymentMethodChangeEvent/methodDetails}} attribute and the
-          {{PaymentResponse}}'s {{PaymentResponse/details}} attribute. A payee can use this
-          information to, for example, calculate tax in certain jurisdictions
-          and update the displayed total. See below for privacy considerations
-          regarding [[[#user-info]]].
+          {{PaymentMethodChangeEvent}}'s
+          {{PaymentMethodChangeEvent/methodDetails}} attribute and the
+          {{PaymentResponse}}'s {{PaymentResponse/details}} attribute. A payee
+          can use this information to, for example, calculate tax in certain
+          jurisdictions and update the displayed total. See below for privacy
+          considerations regarding [[[#user-info]]].
         </dd>
         <dt>
           <dfn>requestBillingAddressParts</dfn> member
         </dt>
         <dd>
           A sequence of {{AddressParts}} that specifies which fine-grained
-          parts of the billing address that the payee wishes to receive when the
-          <a>requestBillingAddress</a> member is true. An empty list means no
-          restriction, i.e. the payee requests the full billing address. The <a>requestBillingAddressParts</a>
-          member has no effect if the <a>requestBillingAddress</a> member is
-          false.
+          parts of the billing address that the payee wishes to receive when
+          the <a>requestBillingAddress</a> member is true. An empty list means
+          no restriction, i.e. the payee requests the full billing address. The
+          <a>requestBillingAddressParts</a> member has no effect if the
+          <a>requestBillingAddress</a> member is false.
           <aside class="note">
-            The billing address is part of the payment-method specific response.
-            The <a>user agent</a> simply passes the payee-specified
+            The billing address is part of the payment-method specific
+            response. The <a>user agent</a> simply passes the payee-specified
             <a>requestBillingAddressParts</a> to the payment handler. It is the
             responsibility of the payment handler to ensure no more than the
             requested pieces is returned. A payment handler may choose to not
-            return some requested pieces (e.g., as a result of applying a <a>
-            redact list</a>). The payee can request additional pieces
+            return some requested pieces (e.g., as a result of applying a
+            <a>redact list</a>). The payee can request additional pieces
             incrementally using {{PaymentMethodChangeEvent}}'s
             {{PaymentMethodChangeEvent/requestBillingAddress()}}.
           </aside>
@@ -2261,19 +2265,18 @@
           and return a shipping address as part of the payment request (e.g.,
           when physical goods need to be shipped by the payee to the user). The
           <a>user agent</a> returns the shipping address in the
-          {{PaymentRequest}}'s {{PaymentRequest/shippingAddress}}
-          attribute and the {{PaymentResponse}}'s
-          {{PaymentResponse/shippingAddress}} attribute. The former is subject
-          to a <a>shipping redact list</a> per the <a>shipping address changed
-          algorithm</a>.
+          {{PaymentRequest}}'s {{PaymentRequest/shippingAddress}} attribute and
+          the {{PaymentResponse}}'s {{PaymentResponse/shippingAddress}}
+          attribute. The former is subject to a <a>shipping redact list</a> per
+          the <a>shipping address changed algorithm</a>.
           <aside class="note">
             <p>
-              A payee may not need the full shipping address of a user until the
-              payment request is completed. For example, just having "country"
-              may be sufficient to calculate shipping cost. A payee sensitive to
-              the user's privacy can use <a>requestShippingAddressParts</a> to
-              restrict the initial set of address parts returned by the
-              <a>user agent</a>.
+              A payee may not need the full shipping address of a user until
+              the payment request is completed. For example, just having
+              "country" may be sufficient to calculate shipping cost. A payee
+              sensitive to the user's privacy can use
+              <a>requestShippingAddressParts</a> to restrict the initial set of
+              address parts returned by the <a>user agent</a>.
             </p>
           </aside>
         </dd>
@@ -2285,10 +2288,11 @@
           parts of the shipping address that the payee wishes to receive when
           the <a>requestShipping</a> member is true. An empty list means no
           restriction, i.e. the <a>user agent</a> will return the full shipping
-          address, subject to the
-          <a>shipping redact list</a>. The {{PaymentOptions/requestShippingAddressParts}} member has no effect if the
-          {{PaymentOptions/requestShipping}} member is false. The payee can request
-          additional pieces incrementally using {{PaymentRequestUpdateEvent}}'s
+          address, subject to the <a>shipping redact list</a>. The
+          {{PaymentOptions/requestShippingAddressParts}} member has no effect
+          if the {{PaymentOptions/requestShipping}} member is false. The payee
+          can request additional pieces incrementally using
+          {{PaymentRequestUpdateEvent}}'s
           {{PaymentRequestUpdateEvent/requestShippingAddress()}}.
         </dd>
         <dt>
@@ -3006,9 +3010,9 @@
           </p>
         </div>
         <ol data-link-for="AddressInit">
-          <li>If |requestedParts| is empty, set it to «
-          "addressLine", "city", "country", "dependentLocality", "organization",
-          "phone", "postalCode", "recipient", "region", "sortingCode"».
+          <li>If |requestedParts| is empty, set it to « "addressLine", "city",
+          "country", "dependentLocality", "organization", "phone",
+          "postalCode", "recipient", "region", "sortingCode"».
           </li>
           <li>Let |details:AddressInit| be an <a>AddressInit</a> dictionary
           with no members present.
@@ -3028,8 +3032,8 @@
           was provided.
           </li>
           <li>If "phone" is in |requestedParts| and is not in |redactList|, set
-          |details|["<a>phone</a>"] to the user-provided phone number, or to the
-          empty string if none was provided.
+          |details|["<a>phone</a>"] to the user-provided phone number, or to
+          the empty string if none was provided.
             <aside class="note" title="Privacy of phone number">
               <p>
                 To maintain users' privacy, implementers need to be mindful
@@ -3054,10 +3058,10 @@
           user-provided recipient organization, or to the empty string if none
           was provided.
           </li>
-          <li>If "postalCode" is in |requestedParts| and is not in |redactList|,
-          set |details|["<a>postalCode</a>"] to the user-provided postal code,
-          or to the empty string if none was provided. Optionally, redact part
-          of |details|["<a>postalCode</a>"].
+          <li>If "postalCode" is in |requestedParts| and is not in
+          |redactList|, set |details|["<a>postalCode</a>"] to the user-provided
+          postal code, or to the empty string if none was provided. Optionally,
+          redact part of |details|["<a>postalCode</a>"].
             <div class="note" title="Privacy of Postal Codes">
               <p>
                 <a>Postal codes</a> in certain countries can be so specific as
@@ -3098,9 +3102,10 @@
               </li>
             </ol>
           </li>
-          <li>If "sortingCode" is in |requestedParts| and is not in |redactList|,
-          set |details|["<a>sortingCode</a>"] to the user-provided sorting code,
-          or to the empty string if none was provided.
+          <li>If "sortingCode" is in |requestedParts| and is not in
+          |redactList|, set |details|["<a>sortingCode</a>"] to the
+          user-provided sorting code, or to the empty string if none was
+          provided.
           </li>
           <li>
             <a data-lt="PaymentAddress.PaymentAddress()">Internally construct a
@@ -3714,9 +3719,9 @@
               <dfn>[[\retryPromise]]</dfn>
             </td>
             <td>
-              Null, or a {{Promise}} that resolves when a <a>user accepts
-              the payment request</a> or rejects if the <a>user aborts the
-              payment request</a>.
+              Null, or a {{Promise}} that resolves when a <a>user accepts the
+              payment request</a> or rejects if the <a>user aborts the payment
+              request</a>.
             </td>
           </tr>
         </table>
@@ -4059,9 +4064,8 @@
         </pre>
         <section>
           <h3>
-             <dfn data-lt="PaymentMethodChangeEvent.PaymentMethodChangeEvent()">
-               Constructor
-             </dfn>
+            <dfn data-lt=
+            "PaymentMethodChangeEvent.PaymentMethodChangeEvent()">Constructor</dfn>
           </h3>
           <p>
             The <a data-lt=
@@ -4074,6 +4078,7 @@
             |type| and |eventInitDict|.
             </li>
             <li>Set |event|.<a>[[\waitForMethodDetails]]</a> to false.
+            </li>
             <li>Return |event|.
             </li>
           </ol>
@@ -4100,19 +4105,19 @@
         </section>
         <section>
           <h2>
-            <dfn data-lt="requestBillingAddress(addressParts)">
-              requestBillingAddress()
-            </dfn> method
+            <dfn data-lt=
+            "requestBillingAddress(addressParts)">requestBillingAddress()</dfn>
+            method
           </h2>
           <aside class="note">
             <p>
               The {{PaymentMethodChangeEvent/requestBillingAddress()}} method
-              allows the developer to incrementally request additional pieces of
-              the billing address that were not initially requested in
-              {{ PaymentOptions/requestBillingAddressParts }}.
+              allows the developer to incrementally request additional pieces
+              of the billing address that were not initially requested in {{
+              PaymentOptions/requestBillingAddressParts }}.
             </p>
-            <pre class="example js"
-              title="how to use `requestBillingAddress()` correctly.">
+            <pre class="example js" title=
+            "how to use `requestBillingAddress()` correctly.">
               // ✅ Good: request more address parts to compute new total. The
               // final response will only contain "country" and "postalCode".
               request.onpaymentmethodchange = (ev) =&gt; {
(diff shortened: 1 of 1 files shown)


ianbjacobs:
Thank you, @danyao! 
cc @samuelweiler 

To summarize the proposal:

* A new requestShippingAddress() method enables the merchant to request partial shipping address information.
* To incrementally request address data, repeated calls to requestShippingAddress() can be chained with a final promise to compute new details using the address information. A call to updateWith() with the promise chain can be used to update the payment details.
* For full shipping address information, the merchant can still use the requestShipping boolean. This also helps with backwards compatibility.

We have begun discussions [1] about the delegation of the request for shipping addresses to payment handlers. If the current proposal is adopted, that discussion will need to include a mechanism for the browser to request shipping address information from the payment handler in order to respond to the merchant's incremental request.

As @danyao notes, this proposal does not address billing addresses. Note that:

* By default, merchants do not receive billing address information;  they must explicitly request it (via requestBillingAddress).
* The Basic Card specification defines a redact list that intends to provide enough information for payment authorization.
* This proposal does not require changes to Basic Card.

[1] https://github.com/w3c/payment-handler/issues/337

index.html
```diff
@@ -3955,6 +4040,7 @@ <h2>
         <pre class="idl">
           [Constructor(DOMString type, optional PaymentRequestUpdateEventInit eventInitDict), SecureContext, Exposed=Window]
           interface PaymentRequestUpdateEvent : Event {
+            Promise&lt;PaymentAddress&gt; requestShippingAddress(AddressParts addressParts);
```
rsolomakhin:
```suggestion
            Promise&lt;PaymentAddress&gt; requestShippingAddress(FrozenArray<AddressParts> addressParts);
```

The description on line 4083 talks about a list of AddressParts.

rsolomakhin (approved):
Technically this should work.

index.html
```diff
@@ -3985,6 +4071,39 @@ <h3>
...
+        <section>
+          <h2>
+            <dfn data-lt="requestShippingAddress(addressParts)">requestShippingAddress()</dfn> method
+          </h2>
+          <aside class="note">
+            <p>
+              If a developer wants to incrementally request more granular details
+              of the user's shipping address, for example, to re-calculate
+              shipping cost, then they can call <a>requestShippingAddress()</a>
+              and provide a list of <a>AddressParts</a>, which will return a
```
```suggestion
              and provide a sequence of <a>AddressParts</a>, which will return a
```

index.html
```suggestion
            Promise&lt;PaymentAddress&gt; requestShippingAddress(sequence<AddressParts> addressParts);
```

marcoscaceres: Note that here it should take a sequence... FrozenArrays are returned by methods, but generally not passed into a method call.  Will add a suggestion.

marcoscaceres:
I'm not following why we need the incremental increase in granularity? Is that really necessary? It seems like it would make implementing an interface much more difficult, because you'd need to be constantly updating it to add, remove, different address parts... that might be really confusing for users.

marcoscaceres:
to be clear, I don't follow why we need this: 

> To incrementally request address data, repeated calls to requestShippingAddress() can be chained with a final promise to compute new details using the address information. A call to updateWith() with the promise chain can be used to update the payment details.

Why not just ask up front?

aestes:
@marcoscaceres asking for everything up front requires user agents to over-share in some cases. For instance, maybe you offer flat-rate shipping in countryA but variable-per-region shipping in countryB. To cover both cases up front you'd have to ask for country and region, even though if you had only asked for country and gotten back countryA then you didn't need to know the user's region.

So with incremental requests you can start at least-specific and drill down one level at a time (until you reach the redacted levels) until you have just enough to compute a details update and resolve your updateWith() promise.

marcoscaceres:
Ah, I see. Thanks for clarifying @aestes.

samuelweiler:
This architecture seems like a good approach for both shipping and billing address. 

I would prefer if the "just ask for everything" option were removed - if only as a way to encourage the (admittedly harder to implement) incremental requests.  Failing that, I want to see some specific text discouraging the use of the "just ask for everything".  Perhaps as a sidebar in section 10?  Point out that if potential customer disables it - assuming a UI has such a feature - the merchant will need to fall back to incremental requests anyway or lose the sale.  And if they have to implement the incremental anyway, then why keep the "all"? 

Whether that boolean is removed or not, there should be some specific text encouraging the use of the incremental approach.  That could be in 18.4.2, perhaps as a sidebar.  It should also be mentioned in the security considerations, in 20.6.

@aestes Thank you for reiterating the need for this.  @joshkaroly, thank you for keeping the pressure on to do this with billing addresses, as well.

danyao:
@samuelweiler I incorporated your comments:

- Added a note in Section 10 to discourage the use of the boolean and encourage the use of the new `requestShippingAddress()` method. Also noting potential future deprecation of the boolean.
- Added similar note to Section 18.4.2
- Added the detailed algorithm to Section 18.4.2 to highlight now the incremental request interact with the concept of redactList.

I didn't add a note to the Security Considerations in Section 20.7 because that section currently focuses on non-normative recommendations to payment handler implementer to not over share user information (especially billing address) via `PaymentMethodChangeEvent`. I considered expanding it to highlight the user agent's responsibility to not over share shipping address in `PaymentRequestUpdateEvent`, but I felt it's redundant with the normative requirements encoded in the algorithm in 18.4.2. I also considered adding a non-normative note to compel the payee to use `requestShippingAddress()`, but that also seems redundant with the other notes in Section 10 and 18.4.2.

Would you mind taking another look? @marcoscaceres @ianbjacobs @aestes PTAL as well.

Also, as I was writing up the algorithm, I realize that the UX implementation may be tricky and would like to hear what others think. Say a payee initially requests "country", then request "region", should the user agent prompt the user for permission each time? This could be annoying. Should the user agent ask for a blanket permission upfront, e.g. "your shipping information will be shared"? But this is a bit misleading because user perceives more information is being shared than what the payee is actually asking for.

Currently this is only a problem for user agents with built-in payment methods (e.g. ApplePay in Safari and basic-card in Chrome). Soon, payment handlers will face this problem as well when they start to support delegated shipping and contact information (https://github.com/w3c/payment-handler/issues/337).

adrianhopebailie:
@danyao thank you for this work, sorry you couldn't be at TPAC where we discussed the issue at length.

My read of the room was that there is wide support for this but that the group does not want to hold up the process by making this a normative requirement in v1.0 nor remove the existing APIs which are already widely used.

@marcoscaceres @aestes @rsolomakhin @ianbjacobs is there a way this can be marked as an "optional" feature for v1.0 of the spec?

If so, I propose we make this change and I will do a call for consensus from the WG to proceed with v1.0 including this PR and that amendment.

index.html
```diff
@@ -3981,6 +4080,111 @@ <h3>
...
+              all up front using the {{PaymentOptions.requestShipping}}
+              boolean. For example, if flat-rate shipping is available in
+              country A but only variable per-region shipping is available in
+              country B, a developer should request only "country" first and
+              decide if "region" is needed.
+            </p>
+            <p>
+              A future version of this specification may remove the all-in-one
+              {{PaymentOptions.requestShipping}} boolean.
+            </p>
```
ianbjacobs: @danyao, I think we should delete this paragraph. It is too unknown at this point.

index.html
```diff
@@ -2137,6 +2222,19 @@ <h2>
...
+              A developer may not need the full shipping address of a user, for
+              example, just having "country" may be sufficient to calculate
+              shipping cost. To be sensitive to user's privacy, developers should
+              prefer to use <a>PaymentRequestUpdateEvent.requestShippingAddress()
+              </a> to incrementally request only the address parts they need.
+            </p>
+            <p>
+              A future version of this specification may remove the all-in-one
+              {{PaymentOptions.requestShipping}} boolean.
+            </p>
```
ianbjacobs, line 2265: See other comment; I think we can remove this.

danyao:
Hi all,

I significantly reworked this pull request to now cover both billing and shipping addresses, and clarified how the new fields affect both addresses returned in intermediate events and the final response. I also added specific language to delegate the responsibility of returning an appropriate version of the billing address to the payment handler. The Payment Handler API and Basic Card spec need to be updated to implement the other side of this stub, but I think the Payment Request API side is reasonably complete and stands on its own.

PTAL.

index.html
```diff
@@ -1137,7 +1139,9 @@ <h2>
           <li>
             <p>
               Pass the <a data-lt="converted to an IDL value">converted</a>
-              second element in the |paymentMethod| tuple and |modifiers|.
+              second element in the |paymentMethod| tuple, |modifiers|,
+              |request|.<a>[[\options]]</a>.<a data-lt="PaymentOptions.requestBillingAddress">requestBillingAddress</a> and
```
```suggestion
              |request|.<a>[[\options]]</a>.{{PaymentOptions/requestBillingAddress}} and
```

index.html
```diff
@@ -1137,7 +1139,9 @@ <h2>
...
+              |request|.<a>[[\options]]</a>.<a data-lt="PaymentOptions.requestBillingAddressParts">requestBillingAddressParts</a>.
```
```suggestion
              |request|.<a>[[\options]]</a>.{{PaymentOptions/requestBillingAddressParts}}.
```

index.html
```diff
@@ -1490,6 +1494,20 @@ <h2>
               The <a>PaymentOptions</a> supplied to the constructor.
             </td>
           </tr>
+          <tr>
+            <td>
+              <dfn>[[\requestedShippingAddressParts]]</dfn>
+            </td>
+            <td>
+              The current <a>sequence</a>&lt;<a>AddressParts</a>&gt; specifying
```
```suggestion
              The current <a>sequence</a>&lt;{{AddressParts}}&gt; specifying
```

index.html
```diff
@@ -1490,6 +1494,20 @@ <h2>
...
+              the payee-requested shipping address parts, initially supplied to
+              the constructor via {{PaymentOptions}}'s
+              <a data-lt="PaymentOptions.requestShippingAddressParts">
```
```suggestion
```

index.html
```diff
@@ -1490,6 +1494,20 @@ <h2>
...
+              requestShippingAddressParts</a> and then updated with calls to
```
```suggestion
              {{PaymentOptions/requestShippingAddressParts}} and then updated with calls to
```

index.html
```diff
@@ -1490,6 +1494,20 @@ <h2>
...
+              <a data-lt="PaymentRequestUpdateEvent.requestShippingAddress">
```
```suggestion
```

index.html
```diff
@@ -1490,6 +1494,20 @@ <h2>
...
+              requestShippingAddress()</a>.
```
```suggestion
              {{PaymentRequestUpdateEvent/requestShippingAddress()}}.
```

index.html
```diff
@@ -2031,6 +2049,91 @@ <h2>
...
+          "recipient",
+          "region",
+          "sortingCode"
+        };
+      </pre>
+      <p>
+        The {{AddressParts}} enum is used to represent individual parts of a
+        <a>physical address</a>.
+      </p>
+      <dl data-dfn-for="AddressFields">
```
```suggestion
      <dl>
```

index.html
```diff
@@ -2094,12 +2199,33 @@ <h2>
...
           and return the billing address associated with a <a>payment
           method</a> (e.g., the billing address associated with a credit card).
-          Typically, the user agent will return the billing address as part of
-          the {{PaymentMethodChangeEvent}}'s <a>methodDetails</a>. A merchant
-          can use this information to, for example, calculate tax in certain
-          jurisdictions and update the displayed total. See below for privacy
-          considerations regarding <a href="#user-info">exposing user
-          information</a>.
+          The <a>user agent</a> returns the billing address as part of the
+          {{PaymentMethodChangeEvent}}'s <a>methodDetails</a> attribute and the
```
```suggestion
          {{PaymentMethodChangeEvent}}'s {{PaymentMethodChangeEvent/methodDetails}} attribute and the
```

index.html
```diff
@@ -2094,12 +2199,33 @@ <h2>
...
+          {{PaymentResponse}}'s <a>details</a> attribute. A payee can use this
```
```suggestion
          {{PaymentResponse}}'s {{PaymentResponse/details}} attribute. A payee can use this
```

index.html
```diff
@@ -2094,12 +2199,33 @@ <h2>
...
+          information to, for example, calculate tax in certain jurisdictions
+          and update the displayed total. See below for privacy considerations
+          regarding <a href="#user-info">exposing user information</a>.
```
```suggestion
          regarding [[[#user-info]]].
```

You might need to reword this one after it expands out. However, avoid `<a href=>` and let ReSpec deal with linking :)

index.html
```diff
@@ -2094,12 +2199,33 @@ <h2>
...
+        </dd>
+        <dt>
+          <dfn>requestBillingAddressParts</dfn> member
+        </dt>
+        <dd>
+          A sequence of <a>AddressParts</a> that specifies which fine-grained
```
```suggestion
          A sequence of {{AddressParts}} that specifies which fine-grained
```

index.html
```diff
@@ -2094,12 +2199,33 @@ <h2>
...
+          parts of the billing address that the payee wishes to receive when the
+          <a>requestBillingAddress</a> boolean is true. An empty list means no
```
```suggestion
          <a>requestBillingAddress</a> member is true. An empty list means no
```

index.html
```diff
@@ -2094,12 +2199,33 @@ <h2>
...
+          restriction, i.e. the payee requests the full billing address. This
+          field has no effect if the <a>requestBillingAddress</a> boolean is
```
```suggestion
          field has no effect if the <a>requestBillingAddress</a> member is
```

index.html
```diff
@@ -2094,12 +2199,33 @@ <h2>
...
+          false.
+          <aside class="note">
+            The billing address is part of the payment-method specific response.
+            The <a>user agent</a> simply passes the payee-specified
+            <a>requestBillingAddressParts</a> to the payment handler. It is the
+            responsibility of the payment handler to ensure no more than the
+            requested pieces is returned. A payment handler may choose to not
+            return some requested pieces (e.g., as a result of applying a redact
```
marcoscaceres: Nit: `<a>` around "redact list"

index.html
```diff
@@ -2094,12 +2199,33 @@ <h2>
...
+            list). The payee can request additional pieces incrementally using
+            <a>PaymentMethodChangeEvent.requestBillingAddress()</a>.
```
```suggestion
            {{PaymentMethodChangeEvent}}'s {{PaymentMethodChangeEvent/requestBillingAddres()}}.
```

index.html
```diff
@@ -2133,10 +2259,37 @@ <h2>
...
         <dd>
           A boolean that indicates whether the <a>user agent</a> SHOULD collect
-          and return a shipping address as part of the payment request. For
-          example, this would be set to true when physical goods need to be
-          shipped by the merchant to the user. This would be set to false for
-          the purchase of digital goods.
+          and return a shipping address as part of the payment request (e.g.,
+          when physical goods need to be shipped by the payee to the user). The
+          <a>user agent</a> returns the shipping address in the
+          {{PaymentRequest}}'s <a data-lt="PaymentRequest.shippingAddress">
```
```suggestion
          {{PaymentRequest}}'s {{PaymentRequest/shippingAddress}}
```

index.html
```diff
@@ -2133,10 +2259,37 @@ <h2>
...
+          shippingAddress</a> attribute and the {{PaymentResponse}}'s
```
```suggestion
          attribute and the {{PaymentResponse}}'s
```

index.html
```diff
@@ -2133,10 +2259,37 @@ <h2>
...
+          <a data-lt="PaymentResponse.shippingAddress">shippingAddress</a>
```
```suggestion
          {{PaymentResponse/shippingAddress}}
```

index.html
```diff
@@ -2133,10 +2259,37 @@ <h2>
...
+            </p>
+          </aside>
+        </dd>
+        <dt>
+          <dfn>requestShippingAddressParts</dfn> member
+        </dt>
+        <dd>
+          A sequence of <a>AddressParts</a> that specifies which fine-grained
+          parts of the shipping address that the payee wishes to receive when
+          the <a>requestShipping</a> boolean is true. An empty list means no
```
```suggestion
          the <a>requestShipping</a> member is true. An empty list means no
```

index.html
```diff
@@ -2133,10 +2259,37 @@ <h2>
...
+          restriction, i.e. the <a>user agent</a> will return the full shipping
+          address, subject to any applicable redact list. This field has no
+          effect if the <a>requestShipping</a> boolean is false. The payee can
```
```suggestion
          effect if the <a>requestShipping</a> member is false. The payee can
```

index.html
```diff
@@ -2133,10 +2259,37 @@ <h2>
...
+          request additional pieces incrementally using
+          <a>PaymentRequestUpdateEvent.requestShippingAddress()</a>.
```
```suggestion
          {{PaymentRequestUpdateEvent}}'s {{PaymentRequestUpdateEvent/requestShippingAddress()}}.
```

index.html
```diff
@@ -2937,9 +3098,9 @@ <h2>
               </li>
             </ol>
           </li>
-          <li>If "sortingCode" is not in |redactList|, set
-          |details|["<a>sortingCode</a>"] to the user-provided sorting code, or
-          to the empty string if none was provided.
+          <li>If "sortingCode" is n |requestedParts| and is not in |redactList|,
```
```suggestion
          <li>If "sortingCode" is in |requestedParts| and is not in |redactList|,
```

index.html
```diff
@@ -3915,6 +4077,72 @@ <h2>
             more information.
           </p>
         </section>
+        <section>
+          <h2>
+            <dfn data-lt="requestBillingAddress(addressParts)">requestBillingAddress()</dfn> method
+          </h2>
+          <aside class="note">
+            <p>
+              <a>PaymentMethodChangeEvent.requestBillingAddress()</a> allows the
```
```suggestion
              The {{PaymentMethodChangeEvent/requestBillingAddress()}} method allows the
```

index.html
```diff
@@ -3915,6 +4077,72 @@ <h2>
...
+              developer to incrementally request additional pieces of the
+              billing address that were not initially requested in
+              {{ PaymentOptions.requestBillingAddressParts }}.
+            </p>
+            <pre class="example js" title="how to use `requestBillingAddress()` correct.">
```
```suggestion
            <pre class="example js" title="how to use `requestBillingAddress()`.">
```

index.html
```diff
@@ -3915,6 +4077,72 @@ <h2>
...
+          </aside>
+          <p>
+            The <a>requestBillingAddress(|addressParts|)</a> method MUST act as
+            follows:
+          </p>
+          <ol class="algorithm">
+            <li>Let |event:PaymentMethodChangeEvent| be this
+            {{PaymentMethodChangeEvent}} instance.
+            </li>
+            <li>If |event|'s {{ Event.isTrusted }} attribute is false, then
```
```suggestion
            <li>If |event|'s {{ Event/isTrusted }} attribute is false, then
```

index.html
```diff
@@ -3915,6 +4077,72 @@ <h2>
...
+            return <a>a promise rejected with</a> an {{"InvalidStateError"}}
+            {{DOMException}}.
+            </li>
+            <li>Let |request:PaymentRequest| be the value of |event|'s
+            [=Event/target=].
+            </li>
+            <li>Assert: |request| is an instance of {{PaymentRequest}}.
+            </li>
+            <li>If |request|.<a>[[\state]]</a> is not "<a>interactive</a>",
```
marcoscaceres: It's not clear what happens if it's called multiple times:

```JS
ev.requestBillingAddress(["country"]);
ev.requestBillingAddress(["postcode"]);
```

Or called in a different order:
```JS
ev.updateWith(somePromise); // [[waitForUpdate]] is true now... 
ev.requestBillingAddress(["country"]);
```

And there might also be:

```JS
await ev.updateWith(somePromise);
ev.requestBillingAddress(["country"]); // invalid state I guess
```

So we might need some kind of internal state to prevent some of the above...  requestBillingAddress() can only happen before updateWith() and requestBillingAddress() can only be called once.

index.html
```diff
@@ -3981,6 +4210,100 @@ <h3>
...
+          </aside>
+          <p>
+            The <a>requestShippingAddress(|addressParts|)</a> method MUST act as
+            follows:
+          </p>
+          <ol class="algorithm">
+            <li>Let |event:PaymentRequestUpdateEvent| be this
+            {{PaymentRequestUpdateEvent}} instance.
+            </li>
+            <li>If |event|'s {{ Event.type }} attribute is not <a>
```
```suggestion
            <li>If |event|'s {{ Event/type }} attribute is not <a>
```

index.html
```diff
@@ -3981,6 +4210,100 @@ <h3>
...
+            shippingaddresschange</a>, then return <a>a promise rejected with
+            </a> a {{TypeError}}.
+            <li>Let |request:PaymentRequest| be the value of |event|'s
+            [=Event/target=].
+            </li>
+            <li>Assert: |request| is an instance of {{PaymentRequest}}.
+            </li>
+            <li>If |event|'s {{ Event.isTrusted }} attribute is false, then
```
```suggestion
            <li>If |event|'s {{ Event/isTrusted }} attribute is false, then
```

marcoscaceres (changes requested):
Thanks for continuing to work on this... found a couple of issues. The main think is around the state machine of the event that needs a few clarifications. Having multiple ongoing asynchronous  operations (.updateWidth() + .requestBillingAddress()) is going to be quite challenging to implement... but I guess as long as the order is clear (i.e., await `requestBillingAddress()` first, then do `updateWidth()`, then it might be ok 😅). 

I need to chew on this a bit more... the example makes me wonder if it could be simplified.

rsolomakhin:
>  @marcoscaceres requested a review from @rsolomakhin 4 hours ago

Anything in particular you want me to take a look at or just the whole thing in general?

marcoscaceres:
> Anything in particular you want me to take a look at or just the whole thing in general?

There was some suggested changes to the state machine of the event (i.e., what happens if you call things out of order, what happens if one promise resolves before the other, etc... what happens if you requestAddress, it rejects, and but updateWith resolves? etc). Can you check if that makes sense and if we need it?

rsolomakhin:
> There was some suggested changes to the state machine of the event (i.e., what happens if you call things out of order, what happens if one promise resolves before the other, etc... what happens if you requestAddress, it rejects, and but updateWith resolves? etc). Can you check if that makes sense and if we need it?

I would prefer that `requestShippingAddress()` and `requestBillingAddress()` can be called only inside of the `event.updateWith()` method. Calling the two in parallel seems to introduce a lot of complexity. Is the parallel calling designed to satisfy a requirement that I'm missing?

```javascript
// GOOD
paymentRequest.addEventListener('shippingaddressupdate', (event) => {
  event.updateWith(new Promise((updateAddressResolver) => {
    const country = paymentRequest.shippingAddress.country;
    if (country === 'US') {
      const address = await event.requestShippingAddress(['regionCode', 'country']);
      updateAddressResolver(calculatePriceForUsState(address.regionCode));
    } else {
      updateAddressResolver(calculatePriceForCountry(country));
    }
  }));
});

// BAD
paymentRequest.addEventListener('shippingaddressupdate', (event) => {
  event.updateWith(new Promise((updateAddressResolver) => {
    const country = paymentRequest.shippingAddress.country;
    updateAddressResolver(calculatePriceForCountry(country));
  }));
  const address = await event.requestShippingAddress([]);  // Too late!
});
```

index.html
danyao: Thanks @marcoscaceres for helping to think this through. I agree that it doesn't make sense to call `requestBillingAddress()` after `updateWith()` because the request may not be interactive by then.

I think multiple calls should be allowed if they form a promise chain. For example, maybe a merchant wants to iteratively calculate taxes based on country then region.

Either way, @rsolomakhin's suggestion to scope the two new methods to within `updateWith()` makes sense. I can't think of a use case that would necessitate calling the request*Address() method before and outside updateWith().

I adjusted the examples and algorithms. PTAL.

index.html
danyao: Done.

rsolomakhin (approved):
Great! Thank you so much!

marcoscaceres:
Blocking on tests and implementation commitments.

index.html
```diff
@@ -890,6 +890,8 @@ <h2>
           </li>
           <li>Set |request|.<a>[[\options]]</a> to |options|.
           </li>
+          <li>Set |request|.<a>[[\requestedShippingAddressParts]]</a> to
+          |options|.<a data-lt="PaymentOptions.requestShippingAddressParts">requestShippingAddressParts</a>.
```
```suggestion
          |options|.{{PaymentOptions/requestShippingAddressParts}}.
```

index.html
```diff
@@ -2094,12 +2197,34 @@ <h2>
...
           and return the billing address associated with a <a>payment
           method</a> (e.g., the billing address associated with a credit card).
-          Typically, the user agent will return the billing address as part of
-          the {{PaymentMethodChangeEvent}}'s <a>methodDetails</a>. A merchant
-          can use this information to, for example, calculate tax in certain
-          jurisdictions and update the displayed total. See below for privacy
-          considerations regarding <a href="#user-info">exposing user
-          information</a>.
+          The <a>user agent</a> returns the billing address as part of the
+          {{PaymentMethodChangeEvent}}'s {{PaymentMethodChangeEvent/methodDetails}} attribute and the
```
marcoscaceres: Nit: hmmm, does the document need a run of tidy?

index.html
```diff
@@ -4286,6 +4641,43 @@ <h2>
           </li>
         </ol>
       </section>
+      <section>
+        <h2>
+          <dfn data-lt="shipping-redact-list">
+            Get redact list for shipping address changed algorithm
```
marcoscaceres: I don't think we need an algorithm for this (as also implied by how "shipping-redact-list" is used in the rest of the pull request)... the redact list is always the same. Let's just define:
```
The <dfn>shipping redact list</dfn> is a <a>redact list</a> that includes  « "organization", "phone", "recipient", "addressLine" ».
``` 

We can probably also then get rid of the duplicate privacy note?

index.html
```diff
@@ -2133,10 +2258,40 @@ <h2>
...
-          shipped by the merchant to the user. This would be set to false for
-          the purchase of digital goods.
+          and return a shipping address as part of the payment request (e.g.,
+          when physical goods need to be shipped by the payee to the user). The
+          <a>user agent</a> returns the shipping address in the
+          {{PaymentRequest}}'s {{PaymentRequest/shippingAddress}}
+          attribute and the {{PaymentResponse}}'s
+          {{PaymentResponse/shippingAddress}}
+          attribute. The former is subject to a
+          <a data-lt="shipping-redact-list">redact list</a> per the
```
marcoscaceres: See comment below about this list... 

```suggestion
          <a>shipping redact list</a> per the
```

index.html
```diff
@@ -2133,10 +2258,40 @@ <h2>
...
+              restrict the initial set of address parts returned by the
+              <a>user agent</a>.
+            </p>
+          </aside>
+        </dd>
+        <dt>
+          <dfn>requestShippingAddressParts</dfn> member
+        </dt>
+        <dd>
+          A sequence of <a>AddressParts</a> that specifies which fine-grained
```
```suggestion
          A sequence of {{AddressParts}} that specifies which fine-grained
```

index.html
```diff
@@ -2094,12 +2197,34 @@ <h2>
...
+          regarding [[[#user-info]]].
+        </dd>
+        <dt>
+          <dfn>requestBillingAddressParts</dfn> member
+        </dt>
+        <dd>
+          A sequence of {{AddressParts}} that specifies which fine-grained
+          parts of the billing address that the payee wishes to receive when the
+          <a>requestBillingAddress</a> member is true. An empty list means no
+          restriction, i.e. the payee requests the full billing address. This
```
```suggestion
          restriction, i.e. the payee requests the full billing address. The <a>requestBillingAddressParts</a>
```

index.html
```diff
@@ -2094,12 +2197,34 @@ <h2>
...
+          field has no effect if the <a>requestBillingAddress</a> member is
```
```suggestion
          member has no effect if the <a>requestBillingAddress</a> member is
```

index.html
```diff
@@ -2133,10 +2258,40 @@ <h2>
...
+        <dt>
+          <dfn>requestShippingAddressParts</dfn> member
+        </dt>
+        <dd>
+          A sequence of <a>AddressParts</a> that specifies which fine-grained
+          parts of the shipping address that the payee wishes to receive when
+          the <a>requestShipping</a> member is true. An empty list means no
+          restriction, i.e. the <a>user agent</a> will return the full shipping
+          address, subject to any applicable
+          <a data-lt="shipping-redact-list">redact list</a>. This field has
```
```suggestion
          <a>shipping redact list</a>. This field has
```

index.html
```diff
@@ -2133,10 +2258,40 @@ <h2>
...
+        </dd>
+        <dt>
+          <dfn>requestShippingAddressParts</dfn> member
+        </dt>
+        <dd>
+          A sequence of <a>AddressParts</a> that specifies which fine-grained
+          parts of the shipping address that the payee wishes to receive when
+          the <a>requestShipping</a> member is true. An empty list means no
+          restriction, i.e. the <a>user agent</a> will return the full shipping
+          address, subject to any applicable
```
```suggestion
          address, subject to the
```

index.html
```diff
@@ -2031,6 +2047,91 @@ <h2>
...
+          "recipient",
+          "region",
+          "sortingCode"
+        };
+      </pre>
+      <p>
+        The {{AddressParts}} enum is used to represent individual parts of a
+        <a>physical address</a>.
+      </p>
+      <dl>
```
marcoscaceres, line 2060: I have a suspicion that some of these might be linking to the IDL definitions instead of the concepts... if they are: 

```suggestion
      <dl data-link-for="">
```

index.html
```diff
@@ -3915,6 +4079,78 @@ <h2>
...
+            </dfn> method
+          </h2>
+          <aside class="note">
+            <p>
+              The {{PaymentMethodChangeEvent/requestBillingAddress()}} method allows the
+              developer to incrementally request additional pieces of the
+              billing address that were not initially requested in
+              {{ PaymentOptions.requestBillingAddressParts }}.
+            </p>
+            <pre class="example js" title="how to use `requestBillingAddress()` correctly.">
```
marcoscaceres: The nesting makes this a little bit hard to read... suggestion:

```JS
request.onpaymentmethodchange = ev =&gt; {
  // ✅ Good: request finer-grained address parts inside updateWith()
  ev.updateWith(
    requestParts(ev, ["country", "postalCode"])
  );
};

async function requestParts(ev, parts) {
  const methodDetails = await ev.requestBillingAddress(parts);
  const newTotal = await getNewTotal(methodDetails);
  return newTotal;
}
```

index.html
marcoscaceres: > I think multiple calls should be allowed if they form a promise chain. For example, maybe a merchant wants to iteratively calculate taxes based on country then region.

Right, yes, calling ev.requestBillingAddress() per `.updateWith()`, with multiple `.updateWith()` calls makes sense.. right now it's not clear what one would do in the UI with:

```JS
ev.updateWith((async () => {
   const p1 = ev.requestBillingAddress(["country"]);
   await new Promise(r => setTimeout(r, 2000));
   const p2 = ev.requestBillingAddress(["postcode"]);
   await new Promise(r => setTimeout(r, 2000));
   const p3 = ev.requestBillingAddress(["city", "state"]);
   await Promise.all([p1, p2, p3]);
   return someNewTotal();
}));
```

Should calling `ev.requestBillingAddress()` more than once inside `ev.updateWith()` reject? Or should the browser UI just deal with all those?

index.html
```diff
@@ -3981,6 +4218,124 @@ <h3>
...
+                ev.updateWith(getPromiseForNewDetails(ev.target.shippingAddress));
+              };
+            </pre>
+          </aside>
+          <p>
+            The <a>requestShippingAddress(|addressParts|)</a> method MUST act as
+            follows:
+          </p>
+          <ol class="algorithm">
+            <li>Let |event:PaymentRequestUpdateEvent| be this
```
marcoscaceres, line 4312: the `requestShippingAddress()` and `requestBillingAddress()` steps are almost identical. I wonder if we should generalize them to a "request address" algorithm and pass in a /type/ and then "if type is 'billing, do x. If it's 'shipping', do y"?

marcoscaceres (changes requested):
Couple more suggestions... getting there tho! 💪

index.html
danyao: Is there a standard formatter that we use for W3C specs? I tried [prettier](https://prettier.io) on this file but it modifies a lot of existing content. So I was planning to land this patch, and then run `prettier` once to just adjust the format. I'd be happy to help document the steps to keep a spec style-compliant.

index.html
danyao: I checked and they all link to Section 12 Physical Address. I think this is what we expect, right?

index.html
marcoscaceres: Yes, sorry. I thought it was listed in one of the guides. We use:

https://github.com/htacg/tidy-html5

If you are on a Mac:
```
brew install tidy-html5
```

And to run it:

```
tidy -config tidyconfig.txt -o index.html index.html
```

index.html
marcoscaceres: Ok, cool. As long as they are fine, then no problem :)

index.html
danyao: Thanks! I tried the command, but `tidy` complains about not recognizing &lt;section&gt;. Do you know why?
```
line 94 column 5 - Error: <section> is not recognized!
line 94 column 5 - Warning: discarding unexpected <section>
[omitted similar lines]
645 warnings, 148 errors were found! Not all warnings/errors were shown.

This document has errors that must be fixed before
using HTML Tidy to generate a tidied up version.
```

marcoscaceres: oh, it must not be finding the right tidy version.... check:

```
tidy --version
```

Should say "HTML Tidy for Apple macOS version 5.6.0". Otherwise, it didn't install correctly. 

No worries if we can't tidy it right now tho... we can always do that last.

index.html
danyao: I considered that, but unfortunately they are just different enough I think a combined algorithm will have a lot of "if "billing", do x, else if "shipping", do y", which I think is less readable.

After Shipping Delegation [1] ships, I think we will be able to refactor all shipping / contact information algorithms to align with billing address, as all information will come from the payment handler. Can I defer this refactor until then?

[1] https://github.com/w3c/payment-handler/pull/349

index.html
danyao: > Should calling `ev.requestBillingAddress()` more than once inside `ev.updateWith()` reject? Or should the browser UI just deal with all those?

Multiple calls that are chained as a thenable should be allowed, but parallel calls should not. I introduced a new Internal Slot for both event types to enforce this.

As after discussing @rsolomakhin, I don't think there's actually a way to check that `requestBillingAddress()` is only called "inside" `updateWith()`, because by "inside", we actually mean that `requestBillingAddress()` be called inside the Promise that is passed to `updateWith()`. From an expression evaluation order perspective, these two are identical:

```
// requestBillingAddress() called "inside" updateWith():
ev.updateWith(async () => {
  const methodDetails = await ev.requestBillingAddress(["country"]);
  const total = await getNewTotal(methodDetails);
  return total;
});

// which is identical to:
let p = ev.requestBillingAddress(["country"])
  .then(methodDetails => {
    return getNewTotal(methodDetails);
  });
ev.updateWith(p);
```

The best solution I can come up with is to require `requestBillingAddress()` be called *before* `updateWith()` (by checking `ev.[[waitForUpdate]]` is not true in `requestBillingAddress()` algorithm). The only case this doesn't cover is if `ev.requestBillingAddress(parts)` is called without calling `ev.updateWith()`. This is essentially a fire-and-forget case, which I think is OK. The payment handler should update its internal states and return a billing address that includes `parts` in the final response.

UI is a bit tricky. I think at a minimum `requestBillingAddress()` should only be callable when the payment handler UI is showing so that the merchant is only able to request address as part of an active transaction with a payment handler. The current algorithm enforces this by checking `request.[[state]]` is "interactive". I'm not sure if it makes sense to also spec a user prompt that either the browser of the payment handler needs to show for every `requestBillingAddress()` call. It seems to expose too much implementation details to the user.

> Right, yes, calling ev.requestBillingAddress() per `.updateWith()`, with multiple `.updateWith()` calls makes sense.. right now it's not clear what one would do in the UI with:
> 
> ```js
> ev.updateWith((async () => {
>    const p1 = ev.requestBillingAddress(["country"]);
>    await new Promise(r => setTimeout(r, 2000));
>    const p2 = ev.requestBillingAddress(["postcode"]);
>    await new Promise(r => setTimeout(r, 2000));
>    const p3 = ev.requestBillingAddress(["city", "state"]);
>    await Promise.all([p1, p2, p3]);
>    return someNewTotal();
> }));
> ```

Yeah I think this case should not be allowed because the three calls are "in parallel". I introduced a new internal slot on both event types to check this and also expanded the code examples to cover all known good and bad patterns.

danyao:
Hi @marcoscaceres! Thanks for your patience with this pull request. I think I've addressed all your comments. Would you mind taking another pass?

marcoscaceres:
Sure! will take a look soon!

index.html
```diff
@@ -2133,10 +2258,38 @@ <h2>
...
+        <dt>
+          <dfn>requestShippingAddressParts</dfn> member
+        </dt>
+        <dd>
+          A sequence of {{AddressParts}} that specifies which fine-grained
+          parts of the shipping address that the payee wishes to receive when
+          the <a>requestShipping</a> member is true. An empty list means no
+          restriction, i.e. the <a>user agent</a> will return the full shipping
+          address, subject to the
+          <a>shipping redact list</a>. This field has no effect if the
```
marcoscaceres: We generally try to avoid the word "field", as it's ambiguous in the context of JS/WebIDL. Properties on dictionaries are formally referred to as "members".

```suggestion
          <a>shipping redact list</a>. The {{PaymentOptions/requestShippingAddressParts}} member has no effect if the
```

index.html
```diff
@@ -2133,10 +2258,38 @@ <h2>
...
+          <a>requestShipping</a> member is false. The payee can request
```
```suggestion
          {{PaymentOptions/requestShipping}} member is false. The payee can request
```

index.html
```diff
@@ -2830,7 +2983,8 @@ <h2>
         <p>
           The steps to <dfn>create a `PaymentAddress` from user-provided
           input</dfn> are given by the following algorithm. The algorithm takes
-          a <a>list</a> |redactList|.
+          a <a>list</a> |redactList| and a
+          <a>sequence</a>&lt;<a>AddressParts</a>&gt; |requestedParts|.
```
```suggestion
          <a>sequence</a>&lt;{{AddressParts}}&gt; |requestedParts|.
```

index.html
```diff
@@ -3915,6 +4098,119 @@ <h2>
...
+            <dfn data-lt="requestBillingAddress(addressParts)">
+              requestBillingAddress()
+            </dfn> method
+          </h2>
+          <aside class="note">
+            <p>
+              The {{PaymentMethodChangeEvent/requestBillingAddress()}} method
+              allows the developer to incrementally request additional pieces of
+              the billing address that were not initially requested in
+              {{ PaymentOptions.requestBillingAddressParts }}.
```
marcoscaceres: Just noting that the usage of {{X.Y}} in other parts of the spec is wrong... I need to fix those ☺️ The {{X.Y}} means "access **static** method/attribute Y on X" - while {{X / Y}} means "access the **instance** method/attribute Y of X"

```suggestion
              {{ PaymentOptions/requestBillingAddressParts }}.
```

marcoscaceres:
As I was just reviewing this... I... um... had a thought...🤔 what if: 

```JS
typedef (sequence<AddressParts> or boolean) AddressRequirement;

dictionary PaymentOptions {
    AddressRequirement requestBillingAddress = false;
    AddressRequirement requestShippingAddress = false; 
};
```
That would:

 * retain backwards compatibility with 1.0 (sequence just gets coerced to `true` in 1.0 implementations).
 * it avoids the `{ requestShippingAddress: false, requestShippingAddressParts: ["country"] }` developer pitfall. 
 * avoid adding adding extra members

WDYT? 

(`AddressRequirement` is a terrible name... we can change that.)

index.html
marcoscaceres: I'm ok to defer.

index.html
marcoscaceres: I need to chew on this a bit more too...

danyao:
> ```js
> typedef (sequence<AddressParts> or boolean) AddressRequirement;
> 
> dictionary PaymentOptions {
>     AddressRequirement requestBillingAddress = false;
>     AddressRequirement requestShippingAddress = false; 
> };
> ```
> 
> That would:
> 
> * retain backwards compatibility with 1.0 (sequence just gets coerced to `true` in 1.0 implementations).
> * it avoids the `{ requestShippingAddress: false, requestShippingAddressParts: ["country"] }` developer pitfall.
> * avoid adding adding extra members
> 
> WDYT?

I thought of two issues, but they seem solvable. First one is (old code + new browser) combination. If a website sets `PaymentOptions.requestShippingAddress = true`, a new browser needs to know how to translate that into a list. Does this mean we won't be able to leverage IDL type checking?

Second is how to differentiate between "address not requested" and "all address parts requested" if there is only a list. Currently with the boolean flag, we are treating {flag=true, list=[]} as "all". If there's only one list, I can think of three ways:

1. `list=null` means "address not requested" and `list=[]` means "all parts requested". But because both `null` and `[]` are falsey values, it seems an easy pitfall.
1. Introduce an "all" sentinel value to `AddressParts` to mean "all parts requested". All algorithms that take this list need to first translate "all" to a list of all parts. This feels a bit cumbersome to describe in spec language, but is probably easy enough in actual code.
1. Make developer explicitly list all parts when they want all. This can be a bit annoying when introducing new address parts because every developer needs to update their code to explicitly request the new part. But maybe this is a privacy feature? Also address formats change slowly so this may not be a big deal in reality.

marcoscaceres:
> I thought of two issues, but they seem solvable. First one is (old code + new browser) combination. If a website sets PaymentOptions.requestShippingAddress = true, a new browser needs to know how to translate that into a list. Does this mean we won't be able to leverage IDL type checking?

According to WebIDL, we should be able to distinguish between them. So if a boolean is passed, and it's `true`, then we can substitute it with some kind of default list. 

> If there's only one list, I can think of three ways: 

Oooh! I **really like** where you are going with 3!: 

 * It never over-shares if we add new address parts, 
 * it forces the developer to have good think about what they need, 
 * and it gives us a nice way to deprecate the boolean over the next few years, if we want.

marcoscaceres:
No more payment address with #955

PR #875, opened 2019-07-18
Remove requirement to reject after document is inactive

### LINKED ISSUE #872: Reject promise if document becomes not fully active
There are multiple places in the spec today that requires any outstanding promises to be rejected if the document becomes not fully active, e.g. [show()], [retry()], [complete()].

[show()]: https://w3c.github.io/payment-request/#show-method
[retry()]: https://w3c.github.io/payment-request/#retry-method
[complete()]: https://w3c.github.io/payment-request/#complete-method

In Blink, this turns out to be difficult to implement because there is no hook to inform the `blink::PaymentRequest` object that its associated context is about to be destroy. The only hook is after the context is destroyed. This is too late, as any rejection queue then will not actually be executed. So from web developer's perspective, the promise gets into a limbo state, and never rejects nor resolves (http://crbug.com/941271).

Where this causes a real problem for web developers is when a `show()` promise is created from a `PaymentRequest` in one context (e.g. an iframe) and passed to a second context (e.g. a parent frame), and the first context navigates away. If the parent frame relies on the promise to resolve or reject to control its UI, it will hang.

@domenic you suggested the following on the [platform-architecture-dev thread] regarding this issue:

> make resolve/reject a no-op if the promise's relevant Realm's global object is a Window, and that Window's associated Document is not fully-active. And then, I guess, remove all uses of rejected promises to signal failures due to document inactivity, like the one opening this thread.

[platform-architecture-dev thread]: https://groups.google.com/a/chromium.org/d/msg/platform-architecture-dev/BszBZqcPcuk/SgTinm_KAQAJ

I'm not sure I fully follow what you mean. Can you clarify how this solves the problem?

In the end, I think there are two options out of this problem:

1. Change Payment Request API spec to not require outstanding promises be rejected when the document becomes not fully active. I'm not sure how to address the parent frame in limbo question.
2. Figure out a way to solve this problem in Blink. Based on a discussion I had with @aestes, WebKit doesn't have this problem because there is a hook just before the context is destroyed. So perhaps Blink can introduce a similar hook.

Any thoughts? @marcoscaceres @domenic @aestes @rsolomakhin @ianbjacobs

danyao:
closes #872

 * [x] Modified Web platform tests (https://github.com/web-platform-tests/wpt/pull/17920)
 
Implementation commitment:

 * [ ] Safari (link to issue)
 * [x] Chrome (http://crbug.com/941271)
 * [ ] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1567641)
 * [ ] Edge (public signal)

Optional, impact on Payment Handler spec? N/A

***
<a href="https://pr-preview.s3.amazonaws.com/danyao/payment-request/pull/875.html" title="Last updated on Jul 18, 2019, 11:19 PM UTC (f2f8ba4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/875/2f35ff0...danyao:f2f8ba4.html" title="Last updated on Jul 18, 2019, 11:19 PM UTC (f2f8ba4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1234,9 +1234,6 @@
               <li>Set |request|'s <a>payment-relevant browsing context</a>'s
               <a>payment request is showing</a> boolean to false.
               </li>
-              <li>Reject |acceptPromise| with an "<a>AbortError</a>"
-              <a>DOMException</a>.
-              </li>
             </ol>
           </li>
         </ol>
@@ -3172,19 +3169,18 @@
           affords the end user the ability to retry accepting the payment
           request.
           </li>
-          <li data-tests=
-          "payment-response/rejects_if_not_active-manual.https.html">If
-          |document| stops being <a>fully active</a> while the user interface
-          is being shown, or no longer is by the time this step is reached,
-          then:
+          <li>
+            <span data-tests=
+            "payment-response/rejects_if_not_active-manual.https.html">If
+            |document| stops being <a>fully active</a> while the user interface
+            is being shown, or no longer is by the time this step is reached,
+            then:
+            </span>
             <ol>
               <li>Close down the user interface.
               </li>
               <li>Set |request|'s <a>payment-relevant browsing context</a>'s
               <a>payment request is showing</a> boolean to false.
-              </li>
-              <li>Reject |retryPromise| with an "<a>AbortError</a>"
-              <a>DOMException</a>.
               </li>
             </ol>
           </li>
@@ -3489,9 +3485,6 @@
               </li>
               <li>Set |request|'s <a>payment-relevant browsing context</a>'s
               <a>payment request is showing</a> boolean to false.
-              </li>
-              <li>Reject |promise| with an "<a>AbortError</a>"
-              <a>DOMException</a>.
               </li>
             </ol>
           </li>


domenic (approved):

marcoscaceres (approved):
👋😢

marcoscaceres:
Just noting that we will be keeping the previous behavior in Firefox. I don’t imagine it will cause any interop issues (hopefully!).

domenic:
How does that work? Does Firefox violate the HTML spec and run the event loop even in non-active documents?

marcoscaceres:
> How does that work? 

We monitor if the document becomes non-active and reject the relevant promises. 

> Does Firefox violate the HTML spec and run the event loop even in non-active documents?

No (maybe). We just don’t leave the promises unsettled.

domenic:
Well, unless you run the event loop in non-active documents, you can never observe the promise becoming settled, because you have to run the event loop to call any fulfillment/rejection handlers... That's why I was mentioning in https://github.com/w3c/payment-request/issues/872#issuecomment-512026384 that introducing this requirement was a mess-up.

marcoscaceres:
Ok, I’ll have another read of the architecture discussion. I didn’t read it very closely as I thought it was a limitation in Blink. If it applies to Gecko too, I can remove the code.

domenic:
Pro-tip: just read my posts in that thread, since they come from someone not as familiar with the Blink architecture, but instead with the spec ^_^.

marcoscaceres:
Ok, I see. I think it should be fine to change Gecko too. I’ll file a bug and do some investigation next week.

marcoscaceres:
https://bugzilla.mozilla.org/show_bug.cgi?id=1567641

PR #885, opened 2019-12-06
Editorial: use new user activation model

### LINKED ISSUE #883: Rephrase "triggered by user activation" to match the new user activation model
We have User Activation v2 model in the HTML spec now.  "Triggered by user activation" should now be changed accordingly.  See the HTML issue mentioned below.

marcoscaceres:
closes #883 

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/885.html" title="Last updated on Dec 10, 2019, 2:27 PM UTC (ca9f2b0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/885/d013715...ca9f2b0.html" title="Last updated on Dec 10, 2019, 2:27 PM UTC (ca9f2b0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -956,13 +956,11 @@
           follows:
         </p>
         <ol class="algorithm">
-          <li>Let |window:Window| be the [=relevant global object=] of the
-          [=environment settings object/responsible document=].
-          </li>
           <li data-tests=
           "payment-request-show-method.https.html, show-method-postmessage-manual.https.html">
-          If |window| does not have [=transient activation=], return [=a
-          promise rejected with=] with a {{"SecurityError"}} {{DOMException}}.
+          If the [=relevant global object=] does not have [=transient
+          activation=], return [=a promise rejected with=] with a
+          {{"SecurityError"}} {{DOMException}}.
           </li>
           <li>Let |request:PaymentRequest| be the <a>context object</a>.
           </li>
@@ -5131,8 +5129,8 @@
         <p>
           To help ensure that users do not inadvertently share sensitive
           credentials with an origin, this API requires that PaymentRequest's
-          <a>show()</a> method be triggered by [=transient activation=] (e.g.,
-          via a click or press).
+          <a>show()</a> method be invoked while the relevant {{Window}} has
+          [=transient activation=] (e.g., via a click or press).
         </p>
         <p>
           To avoid a confusing user experience, this specification limits the
--- a/index.html
+++ b/index.html
@@ -961,9 +961,8 @@
           </li>
           <li data-tests=
           "payment-request-show-method.https.html, show-method-postmessage-manual.https.html">
-          If the method was not triggered by [=transient activation=], or the
-          |window|'s [=transient activation=] has expired, return [=a promise
-          rejected with=] with a {{"SecurityError"}} {{DOMException}}.
+          If |window| does not have [=transient activation=], return [=a
+          promise rejected with=] with a {{"SecurityError"}} {{DOMException}}.
           </li>
           <li>Let |request:PaymentRequest| be the <a>context object</a>.
           </li>
--- a/index.html
+++ b/index.html
@@ -958,8 +958,8 @@
         <ol class="algorithm">
           <li data-tests=
           "payment-request-show-method.https.html, show-method-postmessage-manual.https.html">
-          If the [=relevant global object=] does not have [=transient
-          activation=], return [=a promise rejected with=] with a
+          If the [=relevant global object=] of [=this=] does not have
+          [=transient activation=], return [=a promise rejected with=] with a
           {{"SecurityError"}} {{DOMException}}.
           </li>
           <li>Let |request:PaymentRequest| be the <a>context object</a>.
--- a/index.html
+++ b/index.html
@@ -961,9 +961,9 @@
           </li>
           <li data-tests=
           "payment-request-show-method.https.html, show-method-postmessage-manual.https.html">
-          If the method was not triggered by <a>transient activation</a>, or
-          the |window|'s <a>transient activation</a> has expired, return [=a
-          promise rejected with=] with a {{"SecurityError"}} {{DOMException}}.
+          If the method was not triggered by [=transient activation=], or the
+          |window|'s [=transient activation=] has expired, return [=a promise
+          rejected with=] with a {{"SecurityError"}} {{DOMException}}.
           </li>
           <li>Let |request:PaymentRequest| be the <a>context object</a>.
           </li>
@@ -5132,8 +5132,8 @@
         <p>
           To help ensure that users do not inadvertently share sensitive
           credentials with an origin, this API requires that PaymentRequest's
-          <a>show()</a> method be triggered by <a>transient activation</a>
-          (e.g., via a click or press).
+          <a>show()</a> method be triggered by [=transient activation=] (e.g.,
+          via a click or press).
         </p>
         <p>
           To avoid a confusing user experience, this specification limits the
--- a/index.html
+++ b/index.html
@@ -956,11 +956,14 @@
           follows:
         </p>
         <ol class="algorithm">
+          <li>Let |window:Window| be the [=relevant global object=] of the
+          [=environment settings object/responsible document=].
+          </li>
           <li data-tests=
           "payment-request-show-method.https.html, show-method-postmessage-manual.https.html">
-          If the method was not <a>triggered by user activation</a>, return <a>
-            a promise rejected with</a> with a {{"SecurityError"}}
-            {{DOMException}}.
+          If the method was not triggered by <a>transient activation</a>, or
+          the |window|'s <a>transient activation</a> has expired, return [=a
+          promise rejected with=] with a {{"SecurityError"}} {{DOMException}}.
           </li>
           <li>Let |request:PaymentRequest| be the <a>context object</a>.
           </li>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@mustaqahmed, @domenic, ok.... first shot at this. Let me know how it looks. See also my other comments in the relevant HTML bug.

ianbjacobs:
Thanks @marcoscaceres

index.html
```diff
@@ -5129,7 +5131,7 @@ <h2>
         <p>
           To help ensure that users do not inadvertently share sensitive
           credentials with an origin, this API requires that PaymentRequest's
-          <a>show()</a> method be <a>triggered by user activation</a> (e.g.,
+          <a>show()</a> method be triggered by [=transient activation=] (e.g.,
```
mustaqahmed: To fully eliminate old model's token passing idea, I suggest dropping "triggered by" completely.  What about this: "... `show()` method be invoked while the relevant `Window` has transient activation"?

mustaqahmed (changes requested):

index.html
```diff
@@ -956,11 +956,13 @@ <h2>
           follows:
         </p>
         <ol class="algorithm">
+          <li>Let |window:Window| be the [=relevant global object=] of the
+          [=environment settings object/responsible document=].
```
domenic: Which environment settings object's responsible document is this?

You should use the relevant global object of [this](https://heycam.github.io/webidl/#this) instead. See the general advice in https://html.spec.whatwg.org/multipage/webappapis.html#realms-settings-objects-global-objects

domenic (commented):
Mostly seems good.

Should this consume user activation, or just observe it? That's probably a question for implementers and/or the working group?

index.html
marcoscaceres: It's not clear to me if the "incumbent settings object" (or one of the others) is the one that holds the "transient activation" (I'm seeing it as a flag, because that's how Gecko exposes this in C++... sorta).  I think HTML needs to make this more clear :( Or least how to reason about the realms in relation to each "activation" type. 

(I'm struggling with this because these things don't map super cleanly to [how we do this in Gecko](https://searchfox.org/mozilla-central/source/dom/base/UserActivation.cpp#36) - although I've no doubt we are doing the right thing...  but it's a bit confusing for me at the same time). cc @bzbarsky, as it's maybe of interest... we may be able to also provide some guidance over in Gecko's UserActivation.cpp for people implementing specs.

marcoscaceres:
> Should this consume user activation, or just observe it? That's probably a question for implementers and/or the working group?

I think because this predates the new user activation model in HTML, it's currently "transient". It's definitely transient in Gecko right now - but I think because of the modal nature the payment sheet, we should consume user activation. 

@aestes, @danyao, @rsolomakhin... the problem we are solving is:

```JS
window.onclick = () => {
   // set up payment stuff...
   const pay = request.show();
   // Does this succeed or reject? i.e., did .show() "consume" the user activation.
   document.body.requestFullscreen();
}

```

index.html
domenic: To be clear, the choice of global is not a specific problem for user activation, so any guidance does not belong there. It's the same as any other time you choose a global, which is why I directed you to the section that gives guidance on choosing a global.

Was the paragraph saying "you almost always want the relevant global" not clear enough? What could we do to make it clearer?

marcoscaceres: > Was the paragraph saying "you almost always want the relevant global" not clear enough? 

See below about relocating that text, as it's way too easy to miss!  And way too far down from where a reader needs it.

> What could we do to make it clearer?

The warning about not using "Incumbent" and "entry" needs to be moved to where "Incumbent" and "entry" is defined... it's currently sandwiched between two examples, and looks like it was a spec issue and not **really important guidance**. I totally just skipped over that thinking it was non-normative stuff and some random spec issue/discussion. 

Also, imagine, from my perspective, I just want to get in and out of the HTML spec as quickly as possible. I don't want to dive into the realms etc. unless I absolutely need to. I already understood how "triggered by user activation" works, and all I care about is "will it consume it?"... rest is interesting but 🤷‍♂️. 

The majority of cases for checking the activation are to prevent usage without activation (and throw and error or whatever) - so clear(er) setup steps are still needed (get the global, do the check). Get in, get out, happy spec author (why I also suggested just adding an IDL extended attribute). 

Ok, so even after this discussion, and rereading stuff, I'm still not sure what I'm supposed to change this too? I'm really interested to see how other spec authors going to phrase this (I'm hopefully not the only thick/confused spec Editor here:)).

domenic: I don't think there are really any shortcuts here. The "get in and out as quickly as possible" attitude is not tenable, and I'm not OK with making edits to HTML to encourage it. As a spec editor, dealing with global objects is one of the most fundamental things one needs to understand. One needs to read and grok that section fully.

In particular, the latest patch is missing a statement of which object's relevant global you are looking at. This is emphasized three times in the section I pointed you at:

> When writing algorithms, **the most prominent platform object whose relevant Realm might be important is the this value of the currently-running function object.** In some cases, there can be other important relevant Realms, such as those of any arguments.

and

> Note how the entry, incumbent, and current concepts are usable without qualification, whereas the relevant concept must be applied to a particular platform object.

and 

> In general, web platform specifications should use the relevant concept, applied to the object being operated on (**usually the this value of the current method**).

I'm not really sure how to deal with situations where we have three separate sentences attempting to make something clear, but spec editors are attempting to skim and thus missing them... Suggestions appreciated.

I think in the end it may be offputting that you're coming to this understanding via something that used to be simple-but-wrong (user activation), and being forced to grasp the concept of different global objects in order to fix user activation. But fundamentally different global object usage pervades all specs, including this one; we've spent several issue threads previously working on this (e.g. https://github.com/w3c/payment-request/pull/811).

marcoscaceres: >  Suggestions appreciated

I think we should add a diagram showing how all the things relate - there is just a lot of concepts there, which go down deep rabbit holes... It be great if someone could do a conference talk (or similar) explaining all the relationships in a more human way, tbh.

marcoscaceres: > You should use the relevant global object of **this** instead.

Just want to apologize - I thought that by linking to "this" initially you were pointing me to "this section of WebIDL" - not to a `this` value. I'd forgotten that there was a **this** value available via WebIDL 😅. 

It's the old, "I'd make a JS joke about what `this` is, but I just don't know what `this` is." 😬... 🥺. 

Anyway, to finish this off, should I pull the relevant global from `this` or from the "payment-relevant browsing context"?

domenic: > Anyway, to finish this off, should I pull the relevant global from this or from the "payment-relevant browsing context"?

That's a great question, and really gets at some of the subleties here.

Per https://html.spec.whatwg.org/#activation-notification:

- Whenever an iframed Window is activated, its ancestor (payment-relevant browsing context's Window) will be activated too
- Whenever a payment-relevant browsing context is activated, its descendant will get activation, if its descendant is same-origin.

So the question for the spec editors is, do you want to allow clicking on the top-level Window to activate a PaymentRequest object that comes from a same-origin iframe? (You will never be able to allow clicking on the top-level Window to activate a PaymentRequest object that comes from a different-origin iframe, within the current framework.)

If you do, then checking the payment-relevant browsing context's WindowProxy's [[Window]] makes sense. If you don't, then checking the relevant global of this makes sense.

Personally I'd lean toward the more conservative option, but I'm unsure what folks have implemented, or what kind of workflows the PaymentRequest API hopes to support in this fashion.

marcoscaceres: > You will never be able to allow clicking on the top-level Window to activate a PaymentRequest object that comes from a different-origin iframe, within the current framework.

Oh, um, ok... we might need to check how Stripe and Shopify work (why we have the “allowpaymentrequest” on iframes). That might be a showstopper if they currently depend on this behavior. 

Did the above used to work for other things in old v1 activation model? I seem to recall “triggered by user activation” needing to survive a postMessage() to work with some payment services.

domenic: I suspect they work by clicking within the iframe (but I don't know).

marcoscaceres: Ok, Stripe depends on clicking within the iframe - so we are good there. 

Screenshot 2019-12-11 12 48 17

> So the question for the spec editors is, do you want to allow clicking on the top-level Window to activate a PaymentRequest object that comes from a same-origin iframe? 

I don't think we do. The activated window should be the one where `request.show()` is called.

For the clarity of those watching at home:

```HTML 
  <button>pay</button>
  <iframe src="same-origin.html" hidden>
    <!-- this iframe won't activate the payment request. --> 
  </iframe> 
```

This will never ever work, which is fine:
```HTML
  <button>pay</button>
  <iframe allowpaymentrequest src="https://cross.origin/" hidden>
  </iframe>
```

This will work: 
```HTML
  <iframe src="same-origin.html">
    <!-- something in iframe is clicked/activated, payment request works --> 
  </iframe>
  <iframe allowpaymentrequest src="https://cross.origin/" hidden>
    <!-- something in iframe is clicked/activated, payment request works --> 
  </iframe>
``` 

> Personally I'd lean toward the more conservative option, but I'm unsure what folks have implemented, or what kind of workflows the PaymentRequest API hopes to support in this fashion.

I agree. Let's go with that...  or **this** 😂.

domenic (approved):

marcoscaceres:
I’ll add tests once we deal with consume.

PR #896, opened 2020-01-24
Remove recommendation to localize sheet based on body element

### LINKED ISSUE #895: the user interface's language should not be derived from `the body element`
From the payment request spec, 
https://github.com/w3c/payment-request/blob/fe208dba28724d0a316bf706fbf39cc4414bd8f2/index.html#L1094-L1097
[The body element](https://html.spec.whatwg.org/multipage/dom.html#the-body-element-2)'s definition states "The body element of a document is the first of the html element's children that is either a body element or a frameset element, or null if there is no such element."

Thus,
* `<body>` is a node can can have a language
* `<frameset>` is a node
* `null` is not a node and cannot have a language.

In [this PR](https://github.com/whatwg/html/pull/5228), @annevk states
> I'm not sure I support the use case. Other APIs, e.g., the notification API, have no such defaulting and have explicit support for language annotation. And especially for the frameset case it seems that it's highly likely it might not match the language of the application anyway.

Thus, the user interface's language should not be derived from "the body element". The spec could probably include explicit support for language annotation the way the [notification API does it](https://notifications.spec.whatwg.org/#language).

cc @marcoscaceres

marcoscaceres:
closes #895 

None.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/896.html" title="Last updated on Feb 11, 2020, 7:54 AM UTC (66b77a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/896/fe208db...66b77a3.html" title="Last updated on Feb 11, 2020, 7:54 AM UTC (66b77a3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1089,12 +1089,23 @@
               </li>
             </ol>
           </li>
-          <li>Present a user interface that will allow the user to interact
-          with the |handlers|. The user agent SHOULD prioritize the preference
-          of the user when presenting payment methods. It is RECOMMENDED that
-          the language of the user interface match the [=node/language=] of the
-          <a data-cite="HTML/dom.html#the-body-element-2">the body element</a>
-          element.
+          <li>
+            <p>
+              Present a user interface that will allow the user to interact
+              with the |handlers|. The user agent SHOULD prioritize the
+              preference of the user when presenting payment methods.
+            </p>
+            <aside class="note" title="Localization of the payment sheet">
+              <p>
+                The API does not provide a means for developers to specify the
+                language in which the payment sheet is presented to end
+                users. As such, user agents will generally present the payment
+                sheet using the user agent's default language. The working
+                group is contemplating adding the ability for developers to
+                specify the language the payment sheet, and of specific
+                {{PaymentItems}}, in a future version of this API.
+              </p>
+            </aside>
           </li>
           <li data-tests=
           "show-method-optional-promise-rejects-manual.https.html, show-method-optional-promise-resolves-manual.https.html">
--- a/index.html
+++ b/index.html
@@ -1093,16 +1093,16 @@
             <p>
               Present a user interface that will allow the user to interact
               with the |handlers|. The user agent SHOULD prioritize the
-              preference of the user when presenting payment methods.
+              user's preference when presenting payment methods.
             </p>
             <aside class="note" title="Localization of the payment sheet">
               <p>
-                The API does not provide a means for developers to specify the
+                The API does not provide a way for developers to specify the
                 language in which the payment sheet is presented to end
                 users. As such, user agents will generally present the payment
                 sheet using the user agent's default language. The working
                 group is contemplating adding the ability for developers to
-                specify the language the payment sheet, and of specific
+                specify the language of the payment sheet, and of specific
                 {{PaymentItems}}, in a future version of this API.
               </p>
             </aside>


marcoscaceres:
@janiceshiu, when you can, could you check this to see if it makes sense (and for typos/grammar)? Thanks! 🙏

index.html
```diff
@@ -1089,12 +1089,23 @@ <h2>
...
-          with the |handlers|. The user agent SHOULD prioritize the preference
-          of the user when presenting payment methods. It is RECOMMENDED that
-          the language of the user interface match the [=node/language=] of the
-          <a data-cite="HTML/dom.html#the-body-element-2">the body element</a>
-          element.
+          <li>
+            <p>
+              Present a user interface that will allow the user to interact
+              with the |handlers|. The user agent SHOULD prioritize the
+              preference of the user when presenting payment methods.
```
janiceshiu:
```suggestion
              user's preference when presenting payment methods.
```
* clearer and shorter wording 
* also - it is not very clear to me what `preferences` users are allowed to have. could you clarify this?

index.html
```diff
@@ -1089,12 +1089,23 @@ <h2>
...
+            </p>
+            <aside class="note" title="Localization of the payment sheet">
+              <p>
+                The API does not provide a means for developers to specify the
```
janiceshiu:
```suggestion
                The API does not provide a way for developers to specify the
```
imho `way` is less jargon-y than `means`

index.html
```diff
@@ -1089,12 +1089,23 @@ <h2>
...
+                language in which the payment sheet is presented to end
+                users. As such, user agents will generally present the payment
```
janiceshiu, line 1102: read through the [show method](https://www.w3.org/TR/payment-request/#show-method). after doing a quick search in that section, I noticed that with one exception, the term `user` is always used. that exception is for `end-user` and it appears here.
https://github.com/w3c/payment-request/blob/fe208dba28724d0a316bf706fbf39cc4414bd8f2/index.html#L1119-L1120
elsewhere in the spec, `end user` is occasionally used.

it is not clear to me what is the difference between a `user` and an `end user`. I did not see linked definitions for `user`, `end-user`, or `end user`. there is, however, a definition for `user agent`.

perhaps we could clarify the difference between a `user` and an `end user`, and decide whether to use `end user` or `end-user`?

index.html
```diff
@@ -1089,12 +1089,23 @@ <h2>
...
+                sheet using the user agent's default language. The working
+                group is contemplating adding the ability for developers to
+                specify the language the payment sheet, and of specific
+                {{PaymentItems}}, in a future version of this API.
```
janiceshiu:
```suggestion
The working group is thinking of adding the ability for developers to specify the language for the payment sheet and for specific {{PaymentItems}}. If added, these abilities will be in a future version of this API.
```
* Overall, this is a bit hard to read and understand. I had to read it multiple times. Thus, to make the meaning clearer, I split it into 2 sentences.
* `contemplating adding` is a rather hard to read and understand. probably due to the two successive `-ing` verbs.

index.html
janiceshiu, line 1102:
```suggestion
                users. As such, user agents will generally use the user agent's default language to present the payment sheet.
```
* Sentence was rather hard to read. Have suggested a rewording.
* `generally` doesn't seem to have a particular meaning unlike `SHOULD` and `RECOMMENDED`. In this case, it is not clear how strictly the spec wants user agents to follow the `generally` recommendation. Perhaps we could clarify this?

index.html
```diff
@@ -1089,12 +1089,23 @@ <h2>
...
+              preference of the user when presenting payment methods.
+            </p>
+            <aside class="note" title="Localization of the payment sheet">
+              <p>
+                The API does not provide a means for developers to specify the
+                language in which the payment sheet is presented to end
+                users. As such, user agents will generally present the payment
+                sheet using the user agent's default language. The working
+                group is contemplating adding the ability for developers to
+                specify the language the payment sheet, and of specific
```
janiceshiu:
```suggestion
                specify the language of the payment sheet, and of specific
```
Missing a word. ^^ I've also suggested a rewording of the entire sentence in another comment.

janiceshiu (changes requested):
would be lovely to get some things clarified. also suggested some rewordings to make the text clearer ^^

index.html
marcoscaceres: end user, I think

marcoscaceres: Can fix in a follow up.

marcoscaceres:
@ianbjacobs, over to you for a final read.

ianbjacobs (approved):

Issue #684, opened 2018-02-12
Define handling of multiple applicable modifiers


mnoorenberghe:
1) From what I can tell, the spec doesn't currently define which `additionalDisplayItems` should be shown if a UA decides to show any.
2) More generally, handling of multiple applicable modifiers isn't clear. It looks like it's possible to have multiple `methodData` and `modifiers` with the same `supportedMethods` including the same `data`.

Consider the following example snippet:
```js
// Credit card incurs a $3.00 processing fee.
const creditCardFee = {
  label: "Credit card processing fee",
  amount: { currency: "USD", value: "3.00" },
};

// Visa card incurs a $1.00 processing fee.
const visaFee = {
  label: "Visa processing fee",
  amount: { currency: "USD", value: "1.00" },
};

// Modifiers apply when the user chooses to pay with
// a credit card.
const modifiers = [
  {
    additionalDisplayItems: [creditCardFee],
    supportedMethods: "basic-card",
    total: {
      label: "Total due",
      amount: { currency: "USD", value: "53.00" },
    },
    data: {
      supportedTypes: ["credit"],
    },
  },
  {
    additionalDisplayItems: [visaFee],
    supportedMethods: "basic-card",
    total: {
      label: "Total due",
      amount: { currency: "USD", value: "54.00" },
    },
    data: {
      supportedNetworks: ["visa"]
    },
  },
];
Object.assign(details, { modifiers });
```
If multiple modifiers match the selected payment method (e.g. if the user selects a Visa card):
* Should all applicable additional display items be shown? The first matching one? Last matching one? Most specific one (would need to be defined)
* Which modifier's `total` should be used?

This seems like something that should be tested as well.

ianbjacobs:
@mnoorenberghe, here's my recollection:

1) This is issue 309:
    https://github.com/w3c/payment-request/issues/309

 2) We discussed in March 2017:
     https://www.w3.org/2017/03/23-wpwg-minutes

 3) I believe the "first match wins" approach was adopted
     https://github.com/w3c/payment-request/commit/2972fbb86df49ef7a4f4b9bdfc8b5e3c1c1aee31

@marcoscaceres, there was text in that pull request to cover this question. I don't see it in the latest draft of the spec. Is it implicit in one of the algorithms now?

Thank you,
Ian

marcoscaceres:
I was also sure there was spec text covering this case. Will have a look tomorrow.

marcoscaceres:
[...]

PR #581, opened 2017-08-10
Validate PMIs, ignore duplicates

marcoscaceres:
* Validate supportedMethods (closes #464)
* Defines behavior for duplicate PMIs in constructor and updateWith() (closes #309)

* [x]  ctor PMI check tests: https://github.com/w3c/web-platform-tests/pull/6816
* [x]  updateWith() PMI validity tests: https://github.com/w3c/web-platform-tests/pull/7021

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/payment-request/validate_and_ignore.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/payment-request/76bd629...efde8c3.html)

PR #908, opened 2020-04-27
fix: disallow duplicate PMIs when creating request

### LINKED ISSUE #905: Disallow ambiguous methodData declarations?
In my Masters' thesis, I found a set of issues with the Web Payment APIs (see #903 for further reference). This is one of the mentioned issues.

The current specification allows for methodData declarations that have the following form:

```
const methodData = [
  {
    supportedMethods: "https://example.com/bobpay",
    data: {
      merchantIdentifier: "XXXX",
      bobPaySpecificField: true,
    },
  },
  {
    supportedMethods: "https://example.com/bobpay",
    data: {
      merchantIdentifier: "YYYY",
      bobPaySpecificField: true,
    },
  },
];
```
Such an ambiguous methodData declaration (two different entries for a single payment method identifier), is not further discussed by the spec. 
If a payment handler and/or the trusted payment UI are inconsistent regarding which entry they use, this might have serious implications (such as diferent merchants receiving the payment).
This especially holds true for potential future named payment methods such as basic-card.

In case of the declaration of modifiers, the spec offers a "last-one-wins" approach.
I advocate that such ambiguous methodData declarations should not be allowed at all, at the very least they should be mentioned to raise awareness. 
In my opinion they should lead to an erroneous case on payment request creation.

marcoscaceres:
closes #905

 * [X] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/commit/8bdfc7451d7ad917ebabefc3bedab5f49e0eee5d)
 * [X] Modified MDN Docs - edge case, let's not bother. 
 * [X] Has undergone security/privacy review - part of #905
 
Implementation commitment:

 * [x] [Safari](https://bugs.webkit.org/show_bug.cgi?id=220824)
 * [x] [Chrome](https://crbug.com/1085712)
 * [x] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1633344)
 * [ ] Edge (public signal)

None.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/908.html" title="Last updated on May 22, 2020, 9:41 AM UTC (af76701)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/908/f868c93...af76701.html" title="Last updated on May 22, 2020, 9:41 AM UTC (af76701)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -705,6 +705,8 @@
               [=exception/throw=] a {{TypeError}}, optionally informing the
               developer that at least one <a>payment method</a> is required.
               </li>
+              <li>Let |seenPMIs:Set| be the empty [=set=].
+              </li>
               <li>For each |paymentMethod| of |methodData|:
                 <ol>
                   <li data-tests=
@@ -714,6 +716,21 @@
                   returns false, then throw a {{RangeError}} exception.
                   Optionally, inform the developer that the payment method
                   identifier is invalid.
+                  </li>
+                  <li>Let |pmi| be the result of parsing
+                  |paymentMethod|.{{PaymentMethodData/supportedMethods}} with
+                  [=basic URL parser=]:
+                    <ol>
+                      <li>If failure, set |pmi| to
+                      |paymentMethod|.{{PaymentMethodData/supportedMethods}}.
+                      </li>
+                    </ol>
+                  </li>
+                  <li>If |seenPMIs| [=set/contain|contains=] |pmi| throw a
+                  {{RangeError}} {{DOMException}} optionally letting the
+                  developer this [=payment method identifier=] is a duplicate.
+                  </li>
+                  <li>[=set/Append=] |pmi| to |seenPMIs|.
                   </li>
                   <li>If the {{PaymentMethodData/data}} member of
                   |paymentMethod| is missing, let |serializedData| be null.


marcoscaceres:
We have a patch ready to go on the Firefox side... @edenchuang was also wondering what we should do about duplicate payment method modifiers for the same PMI... we have a last one wins algo there, the maybe we should start throwing for those too (we can do that as a followup).

danyao:
> We have a patch ready to go on the Firefox side... @edenchuang was also wondering what we should do about duplicate payment method modifiers for the same PMI... we have a last one wins algo there, the maybe we should start throwing for those too (we can do that as a followup).

I went to look for why we decided on "last one wins" before and found these two threads:
https://github.com/w3c/payment-request/issues/684
https://github.com/w3c/payment-request/pull/581

IIUC, the main thrust for supporting multiple modifiers for the same PMI is because of `basic-card`. A merchant may want to specify a different modifier for different `supportedNetworks` of `basic-card`.

While I think it'll be conceptually cleaner if the spec treats method data and method modifier consistently, the `basic-card` use case seems legitimate. So I think we should keep the current "last one wins" algorithm and not throw.

marcoscaceres:
This is now implemented in Gecko and we added a Web Platform test for it: https://github.com/web-platform-tests/wpt/commit/8bdfc7451d7ad917ebabefc3bedab5f49e0eee5d 

@danyao or @rsolomakhin, would you mind providing a tracking chrome bug? I think then we are good to merge this.

marcoscaceres:
> While I think it'll be conceptually cleaner if the spec treats method data and method modifier consistently, the basic-card use case seems legitimate. So I think we should keep the current "last one wins" algorithm and not throw.

Agree, with respect to modifiers.

rsolomakhin:
https://crbug.com/1085712

marcoscaceres:
Awesome! @danyao, do you want to give this a last review? I need a ✅ to merge.

danyao (approved):
LGTM!

marcoscaceres:
Was fixed in WebKit by https://bugs.webkit.org/show_bug.cgi?id=220824

ianbjacobs:
Thanks!!

PR #930, opened 2020-09-15
Remove hasEnrolledInstrument() from version 1.0

marcoscaceres:
closes #926 

The following tasks have been completed:

 * [X] Confirmed there are no ReSpec errors/warnings.
 * [X] [Modified Web platform tests] - tests in payment-request for `hasEnrolledInstrument` are all tentative.  
 * [ ] Modified MDN Docs (link)
 * [ ] Has undergone security/privacy review (link)
 
Implementation commitment:

 * [x] Safari - Not implemented. 
 * [ ] Chrome 
 * [x] Firefox - Not implemented. 

Optional, impact on Payment Handler spec?
None

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/930.html" title="Last updated on Oct 5, 2020, 6:42 AM UTC (cb008d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/930/8337fee...cb008d7.html" title="Last updated on Oct 5, 2020, 6:42 AM UTC (cb008d7)">Diff</a>

PR #931, opened 2020-10-07
Add hasEnrolledInstrument() again

marcoscaceres:
Revert "Remove hasEnrolledInstrument() (#930)"

This reverts commit f697360ed9a34fc6974117a66a6a653bf3f2ecd5.

closes #???

Implementation commitment:

 * [ ] Safari (link to issue)
 * [ ] Chrome (link to issue)
 * [ ] Firefox (link to issue)
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/931.html" title="Last updated on Oct 7, 2020, 6:34 AM UTC (bf05cc9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/931/f697360...bf05cc9.html" title="Last updated on Oct 7, 2020, 6:34 AM UTC (bf05cc9)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/931.html" title="Last updated on Feb 26, 2024, 10:06 PM UTC (bf05cc9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/931/f697360...bf05cc9.html" title="Last updated on Feb 26, 2024, 10:06 PM UTC (bf05cc9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -632,6 +632,8 @@
           Promise&lt;undefined&gt; abort();
           [NewObject]
           Promise&lt;boolean&gt; canMakePayment();
+          [NewObject]
+          Promise&lt;boolean&gt; hasEnrolledInstrument();
 
           readonly attribute DOMString id;
           readonly attribute PaymentAddress? shippingAddress;
@@ -1365,7 +1367,7 @@
         <h2>
           <dfn>canMakePayment()</dfn> method
         </h2>
-        <div class="note" title="canMakePayment()">
+        <div class="note" title="canMakePayment() vs hasEnrolledInstrument()">
           <p>
             The {{PaymentRequest/canMakePayment()}} method can be used by the
             developer to determine if the <a>user agent</a> has support for one
@@ -1375,11 +1377,30 @@
           <p>
             A true result from {{PaymentRequest/canMakePayment()}} does not
             imply that the user has a provisioned instrument ready for payment.
+            For that, use {{PaymentRequest/hasEnrolledInstrument()}} instead.
           </p>
         </div>
         <p data-tests="payment-request-canmakepayment-method.https.html">
           The {{PaymentRequest/canMakePayment()}} method MUST run the <a>can
-          make payment algorithm</a>.
+          make payment algorithm</a> with |checkForInstruments| set to false.
+        </p>
+      </section>
+      <section data-dfn-for="PaymentRequest">
+        <h2>
+          <dfn>hasEnrolledInstrument()</dfn> method
+        </h2>
+        <p class="note">
+          The {{PaymentRequest/hasEnrolledInstrument()}} method can be used by
+          the developer to determine if the <a>user agent</a> has support for
+          one of the desired <a>payment methods</a> and if a <a>payment
+          handler</a> has an instrument ready for payment. See
+          [[[#canmakepayment-protections]]].
+        </p>
+        <p data-tests=
+        "payment-request-hasenrolledinstrument-method.https.html">
+          The {{PaymentRequest/hasEnrolledInstrument()}} method MUST run the
+          <a>can make payment algorithm</a> with |checkForInstruments| set to
+          true.
         </p>
       </section>
       <section data-dfn-for="PaymentRequest">
@@ -3924,7 +3945,10 @@
         <p>
           The <dfn>can make payment algorithm</dfn> checks if the <a>user
           agent</a> supports making payment with the <a>payment methods</a>
-          with which the {{PaymentRequest}} was constructed.
+          with which the {{PaymentRequest}} was constructed. It takes a boolean
+          argument, |checkForInstruments|, that specifies whether the algorithm
+          checks for existence of enrolled instruments in addition to
+          supporting a <a>payment method</a>.
         </p>
         <ol class="algorithm">
           <li>Let |request:PaymentRequest| be the {{PaymentRequest}} object on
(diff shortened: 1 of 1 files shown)


dcrousso:
WebKit is interested in adding something along these lines.

For Apple Pay specifically, we require that a `unsigned long version` is provided as part of the `object data` in `PaymentMethodData` that corresponds to the desired [supported Apple Pay version](https://developer.apple.com/documentation/apple_pay_on_the_web/apple_pay_on_the_web_version_history).  Currently, we suggest that developers use [`ApplePaySession.supportsVersion`](https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession/1778014-supportsversion) to check if the desired feature (more specifically the related first version to include that feature) is supported _before_ attempting to do anything with payments (including showing UI).  We'd love for there to be a standardized approach for this :)

Along these lines, would `hasEnrolledInstrument` just result in `false` or reject with an explanatory error if the value provided for `object data` in `PaymentMethodData` (or a `PaymentDetailsModifier`) has issues?  As an  example, if a developer specified a `total` with a negative `value` in its `amount` and then called `hasEnrolledInstrument`, what would happen?  What about if, using the Apple Pay example above, a developer specified `version: 999`?  I ask because it would be nice to have some way of distinguishing between the data provided being invalid (e.g. the `Promise` is rejected with something like "`total` cannot be negative") and the data provided being unsupported (e.g. the `Promise` resolves with `false`).

rsolomakhin:
Hi @dcrousso! A negative `total` will throw an exception in `PaymentRequest` constructor, I believe. By the way, can the version be checked in `canMakePayment()` instead of `hasEnrolledInstrument()`? What is the advantage of `hasEnrolledInstrument()` over `canMakePayment()` for WebKit's use case, if any?

dcrousso:
> A negative `total` will throw an exception in `PaymentRequest` constructor, I believe.

Ah then that was a bad example.  Hopefully you get the idea I was going for :)

> By the way, can the version be checked in `canMakePayment()` instead of `hasEnrolledInstrument()`? What is the advantage of `hasEnrolledInstrument()` over `canMakePayment()` for WebKit's use case, if any?

That would also work!  AFAICT though this is not allowed by the [can make payment algorithm](https://www.w3.org/TR/payment-request/#dfn-can-make-payment-algorithm), as it only looks at the _identifier_ of the _paymentMethod_.

rsolomakhin:
Thank you for the explanation, @dcrousso . That use case makes sense. If the user agent knows the version of the payment app and the `minVersion` parameter to `PaymentRequest` is somehow standardized, then the user agent should be able to check the version of the payment app without the payment app's knowledge, thus not leaking any information to it. Is `minversion` the only piece of information that's necessary for your use case, or were thinking about more data?

dcrousso:
Ideally there'd be a way of checking "is all of the data I've provided to the `PaymentRequest` (specifically the `object data` inside `PaymentMethodData` and `PaymentDetailsModifier`) valid?".  So no, it'd be more than just `version`/`minVersion`.  It'd really be the whole thing.

rsolomakhin:
Are there any privacy concerns about passing an `object data` between origins without explicit user consent? (Probably applies more to platforms where a user can install a 3p payment app.)

dcrousso:
Can you elaborate when/how that'd happen with `PaymentRequest`?

rsolomakhin:
> Can you elaborate when/how that'd happen with PaymentRequest?

1. Install a [payment handler](https://w3c.github.io/payment-handler/) or an [Android payment app](https://web.dev/android-payment-apps-developers-guide/) from `https://payment-app.example`.
2.  Visit `https://shop.example`.
3. `https://shop.example` calls `new PaymentRequest([{supportedMethods: 'https://payment-app.example', data}]).hasEnrolledInstrument()`.
4. User agent silently passes data from `https://shop.example` to the `https://payment-app.example` payment app.

Passing data between origins silently is not something desirable from privacy perspective. Would you agree?

That's why my preference had been to remove the `hasEnrolledInstrument()` method, the `"canmakepayment"` event in Payment Handler API, and the `IS_READY_TO_PAY` intent filter from Android payment apps.

However, I can see how the use case of validating the `object data` is desirable. Do you have any ideas how that can be done in a privacy preserving way?

For example, the spec could define a "read-only mode" for the payment app, so it cannot write user identifying information to disk or have any network access when the user agent passes `object data` into it. That _may be feasible_ for `"canmakepayment"` event, but I am not aware of how this can be guaranteed for native Android or IOS apps, aside from operating system support. Does iOS have this ability?

dcrousso:
Ah I see.  We definitely don't want data to be sent without user knowledge/approval/etc..

A "readonly" (or maybe "validation") mode would be ideal.  All I'm really after is a way to perform steps 16.3-16.4 of the [`show() method`](https://www.w3.org/TR/payment-request/#show-method) without having to actually call `show`.

marcoscaceres:
@dcrousso, I wonder if this is already covered by the [constructor](https://www.w3.org/TR/payment-request/#constructor)'s step 4.3.6.2. As the notes states, "This step assures that any IDL type conversion errors are caught as early as possible."

When doing the conversion there, it would throw if the required `unsigned long version` is missing. 

However, we could clarify in the constructor steps that custom validation could also be performed there.

marcoscaceres:
Spun off a new issue: https://github.com/w3c/payment-request/issues/975

dcrousso:
> @dcrousso, I wonder if this is already covered by the [constructor](https://www.w3.org/TR/payment-request/#constructor)'s step 4.3.6.2. As the notes states, "This step assures that any IDL type conversion errors are caught as early as possible."
> 
> When doing the conversion there, it would throw if the required `unsigned long version` is missing.

Oh wow yeah that would work!  I'll this in WebKit :)

dcrousso:
@marcoscaceres  err, actually, reading _very_ literally, I don't think it would as currently written.  It only says that we can 

> [Convert](https://heycam.github.io/webidl/#dfn-convert-ecmascript-to-idl-value) _object_ to an IDL value of the type specified by the specification that defines the _paymentMethod_

That doesn't mention anything about actually looking at the values _inside_ the IDL type (in this case `ApplePayRequest`) to see if they're valid.  Based on that, I think #976 is desirable/necessary.

marcoscaceres:
@dcrousso agree. Let’s go with #976. If by chance you could share a WebKit bug number, that would be really helpful (tho not required… just helps us track implementation commitment/status).

dcrousso:
@marcoscaceres 
> If by chance you could share a WebKit bug number, that would be really helpful (tho not required… just helps us track implementation commitment/status).

I created \<https://webkit.org/b/233292> ([Payment Request] Validate payment method data on construction) 🙂

marcoscaceres:
Thanks @dcrousso. I've linked it from #976.

Doro3781 (approved):

vinniespina (approved):

vinniespina (commented):
reopen please

PR #942, opened 2021-03-16
Editorial: use unicode name/code for currencies in note

marcoscaceres:
Begins to address https://github.com/w3c/i18n-activity/issues/1044

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/942.html" title="Last updated on Mar 18, 2021, 12:36 AM UTC (2357b5b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/942/62e3ad0...2357b5b.html" title="Last updated on Mar 18, 2021, 12:36 AM UTC (2357b5b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1677,9 +1677,19 @@
             alphabetic code (i.e., the numeric codes are not supported). Their
             canonical form is upper case. However, the set of combinations of
             currency code for which localized currency symbols are available is
-            implementation dependent. Where a localized currency symbol is not
-            available, a user agent SHOULD use U+00A4 (¤) for formatting. User
-            agents MAY format the display of the
+            implementation dependent.
+          </p>
+          <p>
+            Currency symbols can be ambiguous due to use across a number of
+            different currencies (e.g., "$" could mean any of USD, AUD, NZD,
+            CAD, and so on.). When displaying a monetary value, it is
+            RECOMMENDED that user agents display the currency code.
+          </p>
+          <p>
+            Further, when displaying a monetary value, it is OPTIONAL for user
+            agents to display currency symbols. User agent that display
+            currency symbols MAY use U+00A4 (¤) when a currency symbol is not
+            available. User agents MAY format the display of the
             {{PaymentCurrencyAmount/currency}} member to adhere to OS
             conventions (e.g., for localization purposes).
           </p>
--- a/index.html
+++ b/index.html
@@ -1682,7 +1682,7 @@
           <p>
             When displaying a monetary value, it is RECOMMENDED that user
             agents display the currency code, but it's OPTIONAL for user agents
-            to display a currency symbol. Note that currency symbols can be
+            to display a currency symbol. This is because currency symbols can be
             ambiguous due to use across a number of different currencies (e.g.,
             "$" could mean any of USD, AUD, NZD, CAD, and so on.). User agent
             that display currency symbols MAY use U+00A4 (¤) when a currency
--- a/index.html
+++ b/index.html
@@ -1703,7 +1703,7 @@
               Dollar Sign ($), "GBP" is U+00A3 Pound Sign (£), and the
               non-standard "XBT" could be shown as U+0243 Latin Capital Letter
               B with Stroke (Ƀ)). When a code cannot be matched, the
-              specification recommends browsers show a U+0024 Currency Sign
+              specification recommends browsers show a U+00A4 Currency Sign
               (¤).
             </p>
             <p>
--- a/index.html
+++ b/index.html
@@ -1687,7 +1687,8 @@
           </p>
           <p>
             Further, when displaying a monetary value, it is OPTIONAL for user
-            agents to display currency symbols. User agent that display
+            agents to display currency symbols (see note below for rationale,
+            currency symbols can be ambiguous). User agent that display
             currency symbols MAY use U+00A4 (¤) when a currency symbol is not
             available. User agents MAY format the display of the
             {{PaymentCurrencyAmount/currency}} member to adhere to OS
--- a/index.html
+++ b/index.html
@@ -1713,7 +1713,7 @@
               Dollar Sign ($), "GBP" is U+00A3 Pound Sign (£), and the
               non-standard "XBT" could be shown as U+0243 Latin Capital Letter
               B with Stroke (Ƀ)). When a code cannot be matched, the
-              specification recommends browsers show a U+00A4 Currency Sign
+              specification recommends showing a U+00A4 Currency Sign
               (¤).
             </p>
             <p>
--- a/index.html
+++ b/index.html
@@ -1699,10 +1699,12 @@
               [[ISO4217]] list (e.g., XBT, XRP, etc.). If the provided code is
               a currency that the browser knows how to display, then an
               implementation will generally display the appropriate currency
-              symbol in the user interface (e.g., "USD" is shown as "$", "GBP"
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1677,9 +1677,19 @@ <h2>
...
-            implementation dependent. Where a localized currency symbol is not
-            available, a user agent SHOULD use U+00A4 (¤) for formatting. User
-            agents MAY format the display of the
+            implementation dependent.
+          </p>
+          <p>
+            Currency symbols can be ambiguous due to use across a number of
+            different currencies (e.g., "$" could mean any of USD, AUD, NZD,
+            CAD, and so on.). When displaying a monetary value, it is
+            RECOMMENDED that user agents display the currency code.
```
ianbjacobs, line 1686: Maybe "display the currency code rather than the currency symbol alone."

ianbjacobs (commented):
Hi @marcoscaceres,

It feels to me like we should review and harmonize the guidance in
this section. As written I think there are some contradictions between
the ISO4217 and non-ISO sections.

Here's an overall flow (with some questions inline). Let me know what you think.

INPUT:

  1) An [[ISO4217]] <a data-cite=
    "ecma-402#sec-iswellformedcurrencycode">well-formed</a> 3-letter
    alphabetic code (i.e., the numeric codes are not supported).

  2) Current implementations also allow the use of well-formed
     currency codes that are not part of the official [[ISO4217]] list
     (e.g., XBT, XRP, etc.).

  3) In more detail: User agents implementing this specification
     enforce [[ISO4217]]'s 3-letter codes format via ECMAScript’s <a
     data-cite=
     "ecma-402#sec-iswellformedcurrencycode">isWellFormedCurrencyCode</a>
     abstract operation, which is invoked as part of the <a>check and
     canonicalize amount</a> algorithm. When a code does not adhere to
     the [[ISO4217]] defined format, a {{RangeError}} is thrown.

DISPLAY:

 1) Browsers that display the 3-letter currency codes generally do so
    using uppercase.

 2) User agents MAY format the currency to adhere to OS
    conventions (e.g., for localization purposes).

    OBSERVATION: This one overrides the points below so I am suggesting
    to put it first.

 3) Currency symbols


    - Browsers MAY display a currency symbol instead of a currency code.
     (e.g., "USD" is shown as U+0024 Dollar Sign ($), "GBP" is U+00A3
      Pound Sign (£), and the non-standard "XBT" could be
      shown as U+0243 Latin Capital Letter B with Stroke (Ƀ))

    - However, the meaning of currency symbols may be ambiguous, so
      user agents SHOULD display currency codes.

      QUESTION: Should they only do that when there is ambiguity?
      QUESTION: Should they not display the currency symbol at all,
                or should they display the currency code along with the
		symbol?

    - The set of combinations of currency code for which localized
      currency symbols are available is implementation dependent.
    - When a currency symbol is not available, user agent MAY
      MAY use U+00A4 (¤).

ADDITIONAL NOTES:

 * Efforts are underway at ISO to account for digital currencies,
   which may result in an update to the [[ISO4217]] registry or an
   entirely new registry. The community expects this will resolve
   ambiguities that have crept in through the use of non-standard
   3-letter codes; for example, does "BTC" refer to Bitcoin or to a
   future Bhutan currency? At the time of publication, it remains
   unclear what form this evolution will take, or even the time frame
   in which the work will be completed. The W3C Web Payments Working
   Group is liaising with ISO so that, in the future, revisions to
   this specification remain compatible with relevant ISO registries.

marcoscaceres:
> Browsers MAY display a currency symbol instead of a currency code.

We probably don't want to recommend this at all. 

> QUESTION: Should they only do that when there is ambiguity?

No, they SHOULD always do it - otherwise it would get really confusing. 

> QUESTION: Should they not display the currency symbol at all,
or should they display the currency code along with the
symbol?

That's a call for UX team - and it may be locale dependent. Thus, it's not our (the WG's/Spec's) call. We can only point out what the pitfalls are (the note). 

The additional note seems good. I'll update.

marcoscaceres:
> The additional note seems good. I'll update.

Never mind :) it's the same.

marcoscaceres:
> Browsers that display the 3-letter currency codes generally do so using uppercase.

Note that we only say that the canonical form as it relates to the currency member. We don't say anything about how it should be shown: that's a UA/OS decision, not a spec decision.

marcoscaceres:
@ianbjacobs, rereading you comments, and questions aside, I think the updated prose addresses your comments... or at least, can you help me understand more specifically where it doesn't. 

I think there might be a misunderstanding in that there are no "ISO4217 and non-ISO sections." Let me clarify: any three letter combination is valid irrespective of ISO4217, as isWellFormedCurrencyCode doesn't care if a currency code is ISO4217 or not.  Thus, saying `"{"currency": "XXX"}"` or `"{"currency": "XBT"}"` for that matter  is totally valid, even though it's not part of ISO4217.

index.html
```diff
@@ -1677,9 +1677,19 @@ <h2>
             alphabetic code (i.e., the numeric codes are not supported). Their
             canonical form is upper case. However, the set of combinations of
             currency code for which localized currency symbols are available is
-            implementation dependent. Where a localized currency symbol is not
-            available, a user agent SHOULD use U+00A4 (¤) for formatting. User
-            agents MAY format the display of the
+            implementation dependent.
+          </p>
+          <p>
```
ianbjacobs, line 1682: Edited down and re-ordered slightly: 

 <p>When displaying a monetary value, it is
            RECOMMENDED that user agents display the currency code and OPTIONAL for user agents to display a currency symbol. Note that            currency symbols can be ambiguous due to use across a number of
            different currencies (e.g., "$" could mean any of USD, AUD, NZD,
            CAD, and so on.). User agent that display currency symbols MAY use U+00A4 (¤) when a currency symbol is not  available.
</p>
<p>User agents MAY format the display of the
            {{PaymentCurrencyAmount/currency}} member to adhere to OS
            conventions (e.g., for localization purposes).
          </p>

marcoscaceres: Ok, looks good! let me integrate that. The "and OPTIONAL" is a bit awkward, so I might rephrase that a tiny bit. But apart from that, all seems good.

ianbjacobs: Cool! If you want to update the PR I can wrap up the review rapidly. Thanks again!

marcoscaceres: on it!

ianbjacobs (approved):

ianbjacobs:
Hi @marcoscaceres,

At the 15 April 2021 WPWG call it was suggested to add an example of a 2-character currency symbol (just to make the point that not all currency symbols are single Unicode characters). One suggestion was to use PLN / zł.

A concrete editorial proposal is to change

 (e.g., "USD" is shown as U+0024 Dollar Sign ($), "GBP" is U+00A3 Pound Sign (£), and the non-standard "XBT" could be shown as U+0243 Latin Capital Letter B with Stroke (Ƀ)).

to

 (e.g., "USD" is shown as U+0024 Dollar Sign ($), "GBP" is shown as U+00A3 Pound Sign (£), "PLN" is shown as U+007A U+0142 Złoty (zł), and the non-standard "XBT" could be shown as U+0243 Latin Capital Letter B with Stroke (Ƀ)).

cc @aphillips, @r12a

marcoscaceres:
Updated spec to include Złoty.

PR #944, opened 2021-03-23
Recommend Payment UI matches doc's language

marcoscaceres:
Maybe addresses https://github.com/w3c/i18n-activity/issues/1043 

Implementation commitment:

 * [ ] Safari
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge

Although RECOMMENDED, no one implements this at the moment.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/944.html" title="Last updated on Apr 21, 2021, 6:28 AM UTC (e2e899b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/944/df6f802...e2e899b.html" title="Last updated on Apr 21, 2021, 6:28 AM UTC (e2e899b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1136,10 +1136,13 @@
             </ol>
           </li>
           <li>
-            <p>
+            <p data-cite="html">
               Present a user interface that will allow the user to interact
               with the |handlers|. The user agent SHOULD prioritize the user's
-              preference when presenting payment methods.
+              preference when presenting payment methods. It is RECOMMENDED
+              that the language of the user interface match |document|'s
+              [=document element|document element's=] [=Node/language=], if
+              any.
             </p>
             <aside class="note" title="Localization of the payment sheet">
               <p>
--- a/index.html
+++ b/index.html
@@ -1139,10 +1139,11 @@
             <p data-cite="html">
               Present a user interface that will allow the user to interact
               with the |handlers|. The user agent SHOULD prioritize the user's
-              preference when presenting payment methods. It is RECOMMENDED
-              that the language of the user interface match |document|'s
-              [=document element|document element's=] [=Node/language=], if
-              any.
+              preference when presenting payment methods. The user interface
+              SHOULD be presented using the language and locale-based
+              formatting that matches the |document|'s [=document
+              element|document element's=] [=Node/language=], if any, or an
+              appropriate fallback locale if that is not available.
             </p>
             <aside class="note" title="Localization of the payment sheet">
               <p>


ianbjacobs (approved):

index.html
```diff
@@ -1136,10 +1136,13 @@ <h2>
...
           </li>
           <li>
-            <p>
+            <p data-cite="html">
               Present a user interface that will allow the user to interact
               with the |handlers|. The user agent SHOULD prioritize the user's
-              preference when presenting payment methods.
+              preference when presenting payment methods. It is RECOMMENDED
+              that the language of the user interface match |document|'s
+              [=document element|document element's=] [=Node/language=], if
```
ianbjacobs: During the 15 April 2021 WPWG call, it was suggested to delete "if any".

marcoscaceres: It's not true that a document will always have a language. it will have a language attribute, that language attribute might return the empty string.

ianbjacobs:
@aphillips,

At the 15 April 2021 WPWG meeting you suggested dropping "if any" from this pull request; please see the comment above from @marcoscaceres which indicates that the language attribute may return the empty string. At the meeting, @xfq pointed us to the DOM spec [1] which also suggests in some cases there may not be a document language.

Could you confirm that understanding and let us know if this pull request closes #650 and #952. Thank you,

Ian

[1] https://html.spec.whatwg.org/multipage/dom.html#language

index.html
marcoscaceres: https://jsfiddle.net/54f8mLxz/7/

marcoscaceres:
blocked waiting on resolution from i18n.

aphillips:
I'm nervous about saying this addresses #650 and #952 because it is "merely" a recommendation that the language of the native interface match that of the document's node. I think this could be improved by being slightly more explicit. 

I18N has recommendations (and definitions) in our document [LTLI](https://www.w3.org/TR/ltli/#i18n-terminology), particularly around [here](https://www.w3.org/TR/ltli/#ltli-format-like-doc-language). Perhaps the text would be sufficiently stronger if worded something like this:

> The user interface SHOULD be presented using the language and locale-based formatting that matches the |document|'s [=document element|document element's=] [=Node/language=], if any, or an appropriate fallback locale if that is not available.

Note that I didn't define what "an appropriate fallback locale" would be, leaving that up to the implementation, including allowing the user agent to fall back to its own localization/runtime locale.

I am "okay" with the text as is, but my disquiet is that it can be easily overlooked. 

Thoughts?

ianbjacobs:
@aphillips,

Thank you for the concrete suggestion. +1 from me. @marcoscaceres, thoughts?

Ian

marcoscaceres:
+1 from me too. Thanks @aphillips, I'll update the PR.

marcoscaceres:
I've left #650 to be addressed in the future.

stephenmcgruer:
Grah, I typed up a reply to this yesterday but never clicked the green 'Comment' button >_<. So apologies for the post-landing drive-by:

From the Chrome perspective; I believe we agree conceptually that page language is the correct one to use, but it is technically difficult for us to do currently (https://crbug.com/1079289#c18 has some context). At the moment I believe we translate the payment-handler selection UX and basic-card UX into the browser-level language. So we are supportive but are unlikely to be able to move beyond matching the browser-level selected language in the intermediate future :(

PR #955, opened 2021-04-21
Drop PaymentAddress, shipping + billing address support

### LINKED ISSUE #389: PaymentRequest does not handle the case of store pickup
As specified in the spec, if requestShipping is true, then shippingType can be "shipping",  "delivery", or "pickup". But, in case of store pickup, shipping is not done by the merchant instead customer has to pickup from merchant store. How this scenario will be handled in PaymentRequest?

Example: Some restaurant gives some discount if ordered online, we can pickup the order if we are visiting the restaurant.

### LINKED ISSUE #480: Add payerGivenName + payerFamilyName to PaymentAddress
It would be good to then clarify that payerName would be the user’s preferred rendering of their name.

### LINKED ISSUE #537: Shipping Address Validation - Suggest Alternate Address
The scenario I'm trying to address is when a customer enters a shipping address, which fails validation with a shipping service such as FedEx or UPS, but we are able to present the user with a suggested address based on what was provided.

Looking over the API, I only see a way to output an error message to the user explaining that the address was invalid, is there some way we can prompt the user to check whether our alternative address is acceptable to the customer?

### LINKED ISSUE #548: Adding custom fields next after shipping options have been added.
Some ecommerce sites require extra fields with the shipping details other than the address they are sending to. E.G;
- Type of residence
- Instructions to the delivery driver
- Notes to the merchant 

Ideally having an empty text field, which could be populated by the user would work in most cases. Or adding a drop down. 

The reason for doing this in the PR API rather than on the merchants site, is the UX flow would seem broken to partially ask for some delivery details on the merchants site and others in the PR API. 

The other alternative is to ask for all delivery details on the Merchant side, but this begs the question of why using PR API in the first place when all you are asking is credit card details. 

This was originally filed as a [chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=727709#c13), moved here for discussion

### LINKED ISSUE #614: Pre-filter shipping addresses, e.g. by country
If a merchant only ships to certain regions, it would be useful to pre-filter the shipping addresses the user can select from. The case of no eligible address would be quite frequent - the user could see instead an explanation provided by the merchant, along with an affordance for adding an eligible address.

Example use cases:

* Merchant doesn't ship to non-continental US, and the user has two addresses, one in Hawaii and one in California. In this case, only the California address would be shown.

### LINKED ISSUE #615: Pre-calculate shipping costs for available addresses
Use case: the user has 2 or more address they can ship to. The merchant would like to pre-calculate the shipping costs and ideally pre-sort the addresses by cost.

To prevent accessing PII without the user's knowledge, the UA could provide only partial addresses (e.g. Country + zip/postal code) upfront, based on which the merchant could still calculate costs.

### LINKED ISSUE #653: Merchant-specified default shipping address
At TPAC we've talked about merchant defining the default shipping address as a requirement for some implementers or merchants. I'm in favor of this feature. Should we also allow specification of other fields? For example, name, email, phone number. Even basic-card related data is conceivable here, but I'm not sure whether any merchant would use that.

cc: @marcoscaceres @aestes @michelle-stripe @jenan-stripe

### LINKED ISSUE #663: ContactAddress 'region' inconsistencies

### LINKED ISSUE #677: Add latitude and longitude to PaymentAddress interface?
Would it make sense to add optional latitude and longitude fields to the PaymentAddress construct? I see at least two potential use cases:

(1) Providing a location hint in cases where a navigation system can yield ambiguous results (this happens to me all the time at home - delivery drivers will end up 20 miles away at a street with the same name in the next county).

(2) Providing an accurate location in parts of the world where things like street names aren't used (e.g., https://hackernoon.com/online-shopping-in-africa-doesnt-work-because-of-this-web-form-e4974a8d165a describes this kind of scenario).

### LINKED ISSUE #837: Add phonetic name as one of `PaymentAddress` properties
Apple recently decided to [add support](https://webkit.org/blog/8607/release-notes-for-safari-technology-preview-76/) for phonetic name in its Payment Request API's `PaymentAddress` object.

In Japan, people use 3 different characters - Kanji, Hiragana and Katakana. Texts are usually written in mixture of those 3 characters. While Kanji is written for content words, Hiragana and Katakana (they are equivalent different shapes) are written for phonetics. Names written in Kanji needs accompanying Hiragana or Katakana in order to be pronounced properly. That's where phonetic name field helps. They exists commonly in checkout forms in Japanese ecommerce services.

Therefore, I would like to propose to add DOMString `phoneticName` as a member of `PaymentAddress` object.

### LINKED ISSUE #842: Richer negotiation re: address redaction?

### LINKED ISSUE #888: Ship Mode Update for PO BOX
Here's an issue I am running into with Payment Request in regards to PO Boxes and Ship Modes. 

We have 3 Standard Ship Modes that user can select from: 
**Overnight - $30**
**Express - $18**
**Standard - Free**

We can calculate shipping modes based on State and update accordingly. 

User in this case selects Standard Ship Mode. 

For users with a PO Box as shipping address, we only offer one ship mode which is unique to PO Box. 

**Unique Shipping - $25**

The change here would affect total cost of product with shipping and would need to notify user that ship mode has been updated to reflect PO Box address. 

We have limited ship modes for users with PO BOX as shipping address. But, since addressLine is not exposed until user commits to payment, how would Payment Request handle this case?

Would we use the retry method and throw an error and advise user to select the updated ship mode? 
https://www.w3.org/TR/payment-request/#retry-method

cc @rsolomakhin

### LINKED ISSUE #902: Proposal to add optional "description" string field to PaymentShippingOption dictionary
Payment apps supporting shipping delegation and/or use agents can use the merchant provided description field to show more details about each shipping option upon user interaction (e.g. when user hovers over the shipping option label)

### LINKED ISSUE #932: Matching PaymentAddress fields to autofill attributes
Hi,

There is no clear mapping between the fields of [`PaymentAddress`](https://w3c.github.io/payment-request/#physical-addresses) and the values accepted by the [`autocomplete`](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill) attribute for shipping and billing addresses: `PaymentAddress` uses descriptive semantics while `autocomplete` leaves the semantics open to regional variations. Is there a suggested mapping between the two?

Thank you.

### LINKED ISSUE #939: Split shipping address from Payment Request API 1.0
Payment Request API currently specifies a [standardized interface](https://w3c.github.io/payment-request/#dom-paymentoptions) for merchants to request shipping and contact information from payment methods. However, [Basic-Card](https://w3c.github.io/payment-method-basic-card/) is the only payment method that supports this interface. The two other payment methods in production, [Apple Pay](https://developer.apple.com/documentation/apple_pay_on_the_web/applepaypaymentrequest) and [Google Pay](https://developers.google.com/pay/api/web/reference/request-objects#PaymentDataRequest), both implement shipping and contact information collection via [method-specific data](https://w3c.github.io/payment-request/#dom-paymentmethoddata). Other payment app developers have also [requested](https://github.com/w3c/payment-handler/issues/337) shifting the collection of such information from the browser to the payment apps.

On the other hand, shipping address is a sensitive piece of personally identifying information of the user. It has been pointed out (e.g. #842) that any interface for collecting shipping address should support richer negotiation to avoid unnecessary collection of user information.

Given the lack of developer interest for the browser to provide shipping address, and the significant development effort involved to fully implement #842, I'd like to propose that we split out the shipping address related behaviors from Payment Request API 1.0 and to be considered in a future iteration of Payment Request API.

ianbjacobs:
closes w3c/payment-request#842, 
closes w3c/payment-request#939, 

as well as issues raised to add family name, given name, and phonetic name:

closes w3c/contact-picker#69
closes w3c/contact-picker#70 

This edit also reduces the number of strings in the specification, which means there are fewer (but still some) strings requiring better lang/dir support.

Also closes the follow address related issues: 

closes w3c/payment-request#902
closes w3c/contact-picker#71 
closes w3c/payment-request#888
closes w3c/contact-picker#72
closes w3c/contact-picker#65 
closes w3c/payment-request#653
closes w3c/payment-request#615
closes w3c/payment-request#614 
closes w3c/payment-request#548
closes w3c/payment-request#537 
closes w3c/payment-request#389

 * [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/28830)
 
Implementation commitment:

 * [ ] [Safari](https://bugs.webkit.org/show_bug.cgi?id=225432)
 * [ ] [Chrome](https://crbug.com/1206261)
 * [ ] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1709761)
 * [ ] Edge (public signal)

If these changes are adopted, then we will also want to make corresponding changes in Payment Handler API to remove API-level data (instead of payment method level data).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/955.html" title="Last updated on May 6, 2021, 12:46 PM UTC (abd9d43)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/955/4bb62e1...abd9d43.html" title="Last updated on May 6, 2021, 12:46 PM UTC (abd9d43)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -302,6 +302,10 @@
                 amount: { currency: "GBP", value: "55.00" },
               },
               {
+                label: "Value-Added Tax (VAT)",
+                amount: { currency: "GBP", value: "5.00" },
+              },
+              {
                 label: "Standard shipping",
                 amount: { currency: "GBP", value: "5.00" },
               },
@@ -309,7 +313,7 @@
             total: {
               label: "Total due",
               // The total is GBP£65.00 here because we need to
-              // add shipping, which costs GBP£5.00.
+              // add tax and shipping.
               amount: { currency: "GBP", value: "65.00" },
             },
           };
--- a/index.html
+++ b/index.html
@@ -1218,16 +1218,7 @@
               {{PaymentRequest/[[serializedModifierData]]}} internal slot.
             </td>
           </tr>
-          <tr>
-            <td>
-              <dfn data-export="" data-dfn-for=
-              "PaymentRequest">[[\options]]</dfn>
-            </td>
-            <td>
-              The {{PaymentOptions}} supplied to the constructor.
-            </td>
-          </tr>
-          <tr>
+	  <tr>
             <td>
               <dfn>[[\state]]</dfn>
             </td>
--- a/index.html
+++ b/index.html
@@ -112,53 +112,14 @@
           Changes since last publication
         </h3>
         <p>
-          Substantive changes to the Payment Request API since the 9 July 2018
-          version are as follows. The complete list of changes, including all
-          editorial changes, is viewable in the <a href=
+          This version of the specification removes data features from the API,
+          essentially pushing data details to payment method descriptions. The
+          complete list of changes, including all editorial changes, is
+          viewable in the <a href=
           "https://github.com/w3c/payment-request/commits/gh-pages">commit
           history</a>. Key set of changes are viewable in the <a href=
           "#changelog">Changelog</a>.
         </p>
-        <ul>
-          <li>Added support for notification when the user selects a payment
-          handler, but before confirming payment. This allows merchant to
-          update totals, validate acceptance, etc.
-          </li>
-          <li>Added support to notify site of billing address selection. This
-          allows a merchant to update a total (e.g., for VAT in Europe). To
-          enhance privacy, only some billing address data is returned to the
-          merchant as long as the user has not confirmed payment.
-          </li>
-          <li>Added support for {{PaymentResponse/retry()}} and fine-grain
-          error reporting to the user.
-          </li>
-          <li>Clearer definition of {{PaymentRequest/canMakePayment()}} and
-          worked to align implementations. {{PaymentRequest/canMakePayment()}}
-          does not reveal whether payment handler is primed to pay.
-          </li>
-          <li>Removed `languageCode` and `regionCode` from {{PaymentAddress}}.
-          </li>
-          <li>Removed `currencySystem`.
-          </li>
-          <li>"<a>Payment request is showing</a>" boolean now attached to
-          top-level browsing context. Previously, only a single payment UI was
-          allowed to be shown at a time across the whole browser. This now
-          allows multiple browser tabs to show a payment UI at the same time
-          (for those payment handlers able to support it).
-          </li>
-          <li>Integrated with [[[permissions-policy]]].
-          </li>
-          <li>Defined handling of multiple applicable modifiers.
-          </li>
-          <li>Removed support for merchant validation because of lack of
-          multi-implementer support.
-          </li>
-          <li>Deprecated `allowpaymentrequest` attribute.
-          </li>
-          <li>Calling {{PaymentRequest/show()}} now consumes the user
-          activation.
-          </li>
-        </ul>
       </section>
     </section>
     <section class='informative'>
@@ -277,15 +238,10 @@
         </li>
         <li>The |details|: The details of the transaction, as a
         <a>PaymentDetailsInit</a> dictionary. This includes total cost, and
-        optionally a list of goods or services being purchased, for physical
-        goods, and shipping options. Additionally, it can optionally include
-        "modifiers" to how payments are made. For example, "if you pay with a
-        card belonging to network X, it incurs a US$3.00 processing fee".
-        </li>
-        <li>The |options|: Optionally, a list of things as {{PaymentOptions}}
-        that the site needs to deliver the good or service (e.g., for physical
-        goods, the merchant will typically need a physical address to ship to.
-        For digital goods, an email will usually suffice).
+        optionally a list of goods or services being purchased. Additionally,
+        it can optionally include "modifiers" to how payments are made. For
+        example, "if you pay with a card belonging to network X, it incurs a
+        US$3.00 processing fee".
         </li>
       </ul>
       <p>
@@ -353,38 +309,10 @@
             total: {
               label: "Total due",
               // The total is GBP£65.00 here because we need to
-              // add shipping (below). The selected shipping
-              // costs GBP£5.00.
+              // add shipping, which costs GBP£5.00.
               amount: { currency: "GBP", value: "65.00" },
             },
           };
-        </pre>
-      </section>
-      <section>
-        <h3>
-          Adding shipping options
-        </h3>
-        <p>
-          Here we see an example of how to add two shipping options to the
-          |details|.
-        </p>
-        <pre class="example js" title="Adding shipping options">
-          const shippingOptions = [
-            {
-              id: "standard",
-              // Shipping by truck, 2 days
-              label: "🚛  Envío por camión (2 dias)",
-              amount: { currency: "EUR", value: "5.00" },
-              selected: true,
-            },
-            {
-              id: "drone",
-              // Drone shipping, 2 hours
-              label: "🚀 Drone Express (2 horas)",
-              amount: { currency: "EUR", value: "25.00" }
-            },
-          ];
-          Object.assign(details, { shippingOptions });
         </pre>
       </section>
       <section>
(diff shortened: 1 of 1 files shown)


ianbjacobs:
This pull request removes a number of features from the specification around addresses and contact information. I have marked it as blocked because it will require some review before we proceed.

marcoscaceres (approved):
Gave this a print and run through... changes seem fine.

ianbjacobs:
Thanks @marcoscaceres. I’d like to run this by the group.

marcoscaceres:
Sure, take your time!

index.html
```diff
@@ -353,40 +309,12 @@ <h3>
             total: {
```
danyao, line 314: Maybe this example can be made more clear by adding a new display item for shipping? e.g.
```
displayItems: [
  { // sub-total },
  { // tax },
  {
    label: "Standard shipping",
    amount: { currency: "GBP", value: "5.00" },
  },
]
```

danyao (commented):
LGTM!

ianbjacobs:
@marcoscaceres,

We discussed this pull request at today's WPWG meeting:
 https://www.w3.org/2021/04/29-wpwg-minutes#t01

There were no objections to it. 

We have a couple of things to do before a CfC to the WG to proceed (back) to CR
with this edit. But I'll unblock this PR.

Ian

ianbjacobs:
@samuelweiler, can you let us know if removing address features through this pull request would address your concern (raised in w3c/payment-request#842)? Thank you.

marcoscaceres:
@ianbjacobs, when you merge, please make sure the merge message is clear (e.g., "Remove PaymentAddress, shipping + billing address support" or something similar).

marcoscaceres:
Test removals: https://github.com/web-platform-tests/wpt/pull/28830

marcoscaceres:
Filed bugs on Firefox and Safari (links at the top). @rsolomakhin, could you please do the same for Blink?

rsolomakhin:
Filed https://crbug.com/1206261 for Blink.

marcoscaceres:
👋😭 so long `PaymentAddress`, we hardly knew ye.

jcayzac:
> Optional, impact on Payment Handler spec?

Note that [the Contact Picker API also has a dependency on `PaymentAddress`](https://wicg.github.io/contact-api/spec/#idl-index) (although I don't understand the reason for that).

cc @rayankans

rayankans:
@jcayzac @ianbjacobs 

The reason was to not have multiple equivalent address definitions spec'd. The end goal was to move the address to its own spec at some point and have each spec that needs it use it directly.

Is something equivalent to PaymentAddress going to be added to this spec? If not I can just move the definition over to unblock this PR.

ianbjacobs:
@rayankans,
At this point it doesn't seem like we will add back the PaymentAddress. 
@marcoscaceres should we move it to a Note? 
Ian

sideshowbarker:
FYI, I’ve raised PRs to make corresponding updates to MDN and BCD —

* https://github.com/mdn/content/pull/4850
* https://github.com/mdn/browser-compat-data/pull/10413

marcoscaceres:
@rayankans, wrote:

> Is something equivalent to PaymentAddress going to be added to this spec? If not I can just move the definition over to unblock this PR.

Yeah, let's move it over. We can also give it a better name as it's not longer tied to payments.

jcayzac:
@rayankans could you kindly mention me on your issue for moving it? We have some input related to w3c/contact-picker#71 that is not relevant to this repo anymore. Other `PaymentAddress`-related issues mentioned above may also now apply to `ContactAddress`.

Issue #951, opened 2021-03-31
Localizability of errors


aphillips:
6.3 PaymentDetailsUpdate dictionary 
https://www.w3.org/TR/2020/CR-payment-request-20201203/#paymentdetailsupdate-dictionary

>  error member
>    A human-readable string that explains why goods cannot be shipped to the chosen shipping address, or any other reason why no shipping options are available. When the payment request is updated using updateWith(), the PaymentDetailsUpdate can contain a message in the error member that will be displayed to the user if the PaymentDetailsUpdate indicates that there are no valid shippingOptions (and the PaymentRequest was constructed with the requestShipping option set to true). 

There appears to be no language/direction metadata associated with this error message, nor a way to localize the message.

(also found in various places in the spec wherever DOMString `error` appears, for example [here](https://www.w3.org/TR/2020/CR-payment-request-20201203/#paymentvalidationerrors-dictionary))

This is a specific example of (and thus related to) #946 and #948

marcoscaceres:
Based on https://github.com/w3c/payment-request/pull/956, leaving this open to address in a future version of the spec. 

@aphillips, for our records, could you please indicate if i18n is satisfied with this course of action? - and if so, remove the "i18n-needs-resolution" label. 

Please add any other labels the i18n WG needs to track this going forward.

r12a:
(Just a process point: We don't ever remove the -needs-resolution labels.  The transition looks for tracker issues in the i18n WG repo that are still open and have -needs-resolution labels.  So the normal process just involves closing the issue (both here and in the i18n tracking repo) when resolved.  This btw enables us to easily see which issues were raised by the i18n WG, even after closure. We also apply labels to our tracker db (and not usually yours), as we close an item, to indicate that we need to ressurect the comment for the next version of the spec. hth )

marcoscaceres:
Thanks @r12a for clarifying. We were a bit unsure.

r12a:
@marcoscaceres here's the place to find guidance on that: https://www.w3.org/Guide/documentreview/#working_with_horizontal_review_labels

That page is linked from The Guide under 'Wide Review', and is the place you should start when you want any information about how wide/horizontal reviews are done.  We tried to consolidate all the WG-related guidance we could find into that one page, or link to it from that page.

marcoscaceres:
Thanks @r12a! I wasn't aware of that document. That's super helpful.

ianbjacobs:
[...]

Issue #948, opened 2021-03-31
Localization and string metadata for descriptions


aphillips:
2.2 Describing what is being paid for 
https://www.w3.org/TR/2020/CR-payment-request-20201203/#describing-what-is-being-paid-for

Example 2 contains this:
```
const details = {
  id: "super-store-order-123-12312",
  displayItems: [
    {
      label: "Sub-total",
      amount: { currency: "USD", value: "55.00" },
    },
    {
      label: "Sales Tax",
      amount: { currency: "USD", value: "5.00" },
      type: "tax"
    },
  ],
  total: {
    label: "Total due",
    // The total is USD$65.00 here because we need to
    // add shipping (below). The selected shipping
    // costs USD$5.00.
    amount: { currency: "USD", value: "65.00" },
  },
};
```

The value `label` contains natural language text with no language metadata or base direction information. The same usage is found in other example sections that follow. I18N recommends that there be a means of providing this metadata so that the document's processor can display the value correctly.

This comment is probably a duplicate of #327. It is also related to #946

aphillips:
I'm going to add w3c/i18n-activity#1359 and w3c/i18n-activity#1357 as additional examples to this issue instead of cluttering your repo with additional specific issues. I've replicated the text of these issues in this comment.

----
14.1.2 PayerErrors dictionary 
https://www.w3.org/TR/2020/CR-payment-request-20201203/#payererrors-dictionary

```
 const payer = {
  email: "The domain is invalid.",
  phone: "Unknown country code.",
  name: "Not in database.",
};
```

Example 13 shows the payer errors dictionary being used to convey unlabeled natural language string values. As noted in other comments, there is no language negotiation scheme in place. There is also no metadata here to help with display.

----

PaymentShippingOption dictionary 
https://www.w3.org/TR/2020/CR-payment-request-20201203/#paymentshippingoption-dictionary

> >  label member
>    A human-readable string description of the item. The user agent SHOULD use this string to display the shipping option to the user. 

The `label` member has no language or direction metadata. Further, it cannot be localized by providing reference to a multilingual construct of some sort. There doesn't appear to be a language negotiation mechanism so that a language can be requested.

The SHOULD might be problematic vs. allowing the user agent to wrap the provided display string with a localized value, i.e. it could say something approximately like:

> The user agent SHOULD use this string (or a localized version available to the user agent locally) to display the shipping option to the user.

marcoscaceres:
[...]

Issue #946, opened 2021-03-31
Missing language/locale negotiation mechanism


aphillips:
(throughout the spec, some examples follow)
https://www.w3.org/TR/2020/CR-payment-request-20201203/#payererrors-dictionary
https://www.w3.org/TR/2020/CR-payment-request-20201203/#paymentshippingoption-dictionary
https://www.w3.org/TR/2020/CR-payment-request-20201203/#show-method

Throughout the payment-request spec we note that natural language text values are returned. In our teleconference of 2021-03-25 we decided to group a number of pending issues into three buckets:

* The need for language and direction metadata on natural language items (already tracked in payment-request #327)
* The need for localizability of natural language fields (e.g. the ability to have more than one language present)
* The need for a language negotiation mechanism so that the the API caller can receive natural language messages such as errors or labels in their preferred language/locale (if available)

This issue tracks the last of these. It is possible this is a duplicate of #650 

For example, at \#show-method in the NOTE under item # 18:

>    Note: Localization of the payment sheet

>    The API does not provide a way for developers to specify the language in which the payment sheet is presented to end users. As such, user agents will generally present the payment sheet using the user agent's default language. The working group is contemplating adding the ability for developers to specify the language of the payment sheet, and of specific PaymentItems, in a future version of this API.

The user agent's default language and the language of the rendering site or host page might not match. Usually items shown as part of the payment-initiating customer experience should match the language and locale of that experience. Defaulting to the user agent's locale would could thus produce a mixed language experience, which is undesirable.

Why *not* this version?

marcoscaceres:
Based on https://github.com/w3c/payment-request/pull/956, leaving this open to address in a future version of the spec. 

@aphillips, for our records, could you please indicate if i18n is satisfied with this course of action? - and if so, remove the "i18n-needs-resolution" label. 

Please add any other labels the i18n WG needs to track this going forward.

aphillips:
**I am satisfied by the current resolution.**

Please do not remove the "i18n-needs-resolution" label. Closing the issue is sufficient (we then close our matching issue to indicate satisfaction). If you want to keep this open for "future", that's okay too. Basically the label informs PLH's tracking tool which issues were from HR and it lets us keep track of our issues. Any open issues will actually be reviewed and are not a blocker. Alternatively, you have my permission to change the label from "i18n-needs-resolution" to "i18n-tracker".

ianbjacobs:
Sounds like the simplest is to close the issue here; thank you @aphillips

PR #956, opened 2021-05-05
Derive direction from document element too

### LINKED ISSUE #327: Add dir/lang member on dictionaries whose content is displayed in browser UI

marcoscaceres:
Related to, but does not close #327, #946, #948, #951

Implementation commitment:

 * [ ] Safari (link to issue)
 * [ ] Chrome (link to issue)
 * [ ] Firefox (link to issue)
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/956.html" title="Last updated on May 20, 2021, 2:40 AM UTC (d894b79)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/956/79e23d4...d894b79.html" title="Last updated on May 20, 2021, 2:40 AM UTC (d894b79)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1145,15 +1145,17 @@
               {{Document/dir}} attribute, if any, or an appropriate fallback
               locale if that is not available.
             </p>
-            <aside class="note" title="Localization of the payment sheet">
+            <aside class="note" title=
+            "Localization of the payments user interface">
               <p>
                 The API does not currently provide a way for developers to
                 specify the language and base direction in which the payment
-                sheet is presented to end users. As such, user agents will
-                generally present the payment sheet using the user agent's
-                default language. Future versions of this specification may add
-                support for language and direction metadata via mechanism
-                specified in [[STRING-META]] and [[INTERNATIONAL-SPECS]].
+                sheet is presented to end users. Instead, the API relies on
+                localization information inherited from the document. The
+                working group is considering support for fine-grained
+                localization of the user interface and of individual
+                {{PaymentItems}} and user-facing error messages in a future
+                version of this API.
               </p>
             </aside>
           </li>
--- a/index.html
+++ b/index.html
@@ -1139,20 +1139,21 @@
               Present a user interface that will allow the user to interact
               with the |handlers|. The user agent SHOULD prioritize the user's
               preference when presenting payment methods. The user interface
-              SHOULD be presented using the language and locale-based
-              formatting that matches the |document|'s [=document
-              element|document element's=] [=Node/language=], if any, or an
-              appropriate fallback locale if that is not available.
+              SHOULD be presented using the language, base direction, and
+              locale-based formatting that matches the |document|'s [=document
+              element|document element's=] [=Node/language=] and
+              {{Document/dir}} attribute, if any, or an appropriate fallback
+              locale if that is not available.
             </p>
             <aside class="note" title="Localization of the payment sheet">
               <p>
-                The API does not provide a way for developers to specify the
-                language in which the payment sheet is presented to end users.
-                As such, user agents will generally present the payment sheet
-                using the user agent's default language. The working group is
-                contemplating adding the ability for developers to specify the
-                language of the payment sheet, and of specific
-                {{PaymentItems}}, in a future version of this API.
+                The API does not currently provide a way for developers to
+                specify the language and base direction in which the payment
+                sheet is presented to end users. As such, user agents will
+                generally present the payment sheet using the user agent's
+                default language. Future versions of this specification may add
+                support for language and direction metadata via mechanism
+                specified in [[STRING-META]] and [[INTERNATIONAL-SPECS]].
               </p>
             </aside>
           </li>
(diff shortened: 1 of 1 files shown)


ianbjacobs:
Hi @marcoscaceres,

Thanks for starting this. Some thoughts:

1) The phrase "payment sheet" is not used elsewhere in the document. I think we should either define it (e.g., by adding "(called the 'payment sheet')" after "Present a user interface" at the beginning of item 18) or remove it. If we remove it, then the note title could become: "Localization of the user interface to interact with handlers". Then you could just s/payment sheet/user interface/ in the note body.

2) Point 18 says "SHOULD use info provided in content." It's not clear to me that the note needs to say "As such, user agents will generally present the payment sheet using the user agent's default language." That to me suggests that despite the recommendation we just made, nobody is doing it.

What about a Note that says:

 "This API relies on inherited localization information. The working group is considering support for fine-grained localization of the user interface and of individual PaymentItems in a future version of this API."

ianbjacobs (changes requested):
(See my comments above)

marcoscaceres:
> That to me suggests that despite the recommendation we just made, nobody is doing it.

That's the truth tho :) 

Will update based on your suggestions.

ianbjacobs (approved):

ianbjacobs:
@aphillips, would you mind having a look at this proposal to address the lack of fine-grain localization in V1 of the API? Thank you!

index.html
```diff
@@ -1139,20 +1139,23 @@ <h2>
...
               with the |handlers|. The user agent SHOULD prioritize the user's
               preference when presenting payment methods. The user interface
-              SHOULD be presented using the language and locale-based
-              formatting that matches the |document|'s [=document
-              element|document element's=] [=Node/language=], if any, or an
-              appropriate fallback locale if that is not available.
+              SHOULD be presented using the language, base direction, and
+              locale-based formatting that matches the |document|'s [=document
+              element|document element's=] [=Node/language=] and
+              {{Document/dir}} attribute, if any, or an appropriate fallback
```
aphillips: The direction should also be at the Node level, no?

index.html
```diff
@@ -1139,20 +1139,23 @@ <h2>
...
+              locale if that is not available.
```
aphillips: Omit `locale` since this is both locale and direction now?

index.html
```diff
@@ -1139,20 +1139,23 @@ <h2>
...
-                using the user agent's default language. The working group is
-                contemplating adding the ability for developers to specify the
-                language of the payment sheet, and of specific
-                {{PaymentItems}}, in a future version of this API.
+                The API does not currently provide a way for developers to
+                specify the language and base direction in which the payment
+                sheet is presented to end users. Instead, the API relies on
+                localization information inherited from the document. The
+                working group is considering support for fine-grained
+                localization of the user interface and of individual
```
aphillips: I'd probably change "fine-grained" to "control of" (too strong?) or perhaps "requesting  specific"

aphillips (changes requested):
Generally looks good. Small tweaks proposed.

index.html
marcoscaceres: It should, but there is no such thing on Node unfortunately:
https://dom.spec.whatwg.org/#node
 
The only "dir"s in the platform are: 
https://respec.org/xref/?term=dir

marcoscaceres: Oh, wait... we might be able to use https://html.spec.whatwg.org/multipage/dom.html#the-directionality ... but I'll need to get it "exported" so we can use it in our spec.

marcoscaceres: Sent https://github.com/whatwg/html/pull/6676 ...

aphillips: @marcoscaceres Thanks. I squinted at `{{HTMLElement/dir}}` and it wasn't clear to me if it included inherited direction. `[^html-global/dir^]` is the definition of the `dir` attribute and includes inheritance, but doesn't seem quite to match our needs here. Exporting `directionality` is a better solution--not just here but generally! I may take whatwg/html#6676 to expedite ;-).

aphillips: LOL. I thought 6676 was an issue, but it's a PR...

marcoscaceres: @aphillips, take a look at @annevk's comments at https://github.com/whatwg/html/pull/6676 ... I think they make a lot of sense. 

My feeling is that we punt on the direction for now, and solve this with "localizable" in a future version of the spec.

ianbjacobs: @aphillips, would that work for you?

aphillips: LGTM.

@marcoscaceres I'm pondering a reply to @annevk: I think he is right, although I18N does want to promote using the document's context when it is available and is appropriate.

aphillips (approved):

PR #971, opened 2021-10-05
Add internationalization support for human readable labels

ianbjacobs:
The Internationalization Working Group has raised concerns about lack of Internationalization support for two strings in the specification (see issues #948 and #951). 

I agree that the specification should support internationalization.

I believe the following is deployment reality:

* In most cases, language context from the document will suffice for the proper representation of the two strings. That is: I believe that explicit metadata for these strings will rarely be needed.
* These strings are either rendered in the browser "sheet" or in "first party" payment apps (e.g., Google Pay and Apple Pay). These strings are not passed to Web-based payment handlers (through Payment Handler API) and, I believe, not passed to "third party" Android native payment handlers.  (There are no third party payment apps with Safari.)
* I also believe that most real-world uses of payment request API involve "skipping" the sheet. That is: calls to Payment Request API involve a single payment method and a single payment app matches, so the browser launches the payment app directly without displaying the sheet. Thus, in practice, rendering these strings in the sheet is rare. 
* In Chrome the sheet has been used as part of built-in support for Basic Card. However, Chrome has signaled that they plan to deprecate their built-in support for Basic Card, so that use of the sheet is likely to go away. The Web Payments Working Group has also retired the Basic Card payment method specification.

Summary: In practice these strings are used almost exclusively used by the Google Pay and Apple Pay native apps. 

Other observations:

* The Web community is working on a platform-wide solution ("Localizable").
* I believe that browser vendors will consider it a low priority to add support for features that will be rarely necessary and for strings that are not directly handled in the browser (but are passed to a small number of native applications).

Given this context, the current pull request seeks a balance:

* Recognize more visibly the need for internationalization support in this specification. Indicate via Candidate Additions [1] that two dictionaries lack "language" and "direction" members. We can continue to work with Google and Apple on support for these features in Google Pay and Apple Pay.
* Indicate that the Working Group is monitoring the Localizable discussion. It may turn out that a platform-wide solution is preferable to a one-off solution; we don't know yet. We have to consider both backward and forward compatibility in determining how to modify the spec in the future.
* Because these features are expected to be used rarely in the near term, allow the specification to advance to Recommendation with the Candidate Additions. 

[Added after posting: if people disagree with the above characterization, please weigh in on that. Thanks!]

I look forward to feedback from @aphillips, @r12a, and the Internationalization Working Group, as well as @marcoscaceres, @stephenmcgruer, @rsolomakhin, and @nickjshearer.

[1] https://www.w3.org/2020/Process-20200915/#candidate-addition

 * [✅] Confirmed there are no ReSpec errors/warnings.
 
Implementation commitment:

 * [ ] Safari (link to issue)
 * [ ] Chrome (link to issue)
 * [ ] Firefox (link to issue)
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/971.html" title="Last updated on Oct 27, 2021, 8:35 PM UTC (8282cae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/971/ee3bd03...8282cae.html" title="Last updated on Oct 27, 2021, 8:35 PM UTC (8282cae)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -915,9 +915,7 @@
                 The API does not currently provide a way for developers to
                 specify the language and base direction in which the payment
                 sheet is presented to end users. Instead, the API relies on
-                localization information inherited from the document. The
-                working group is considering support for requesting specific
-                localization of the user interface.
+                localization information inherited from the document.
               </p>
             </aside>
           </li>
--- a/index.html
+++ b/index.html
@@ -111,6 +111,11 @@
         There has been no change in dependencies on other workings groups
         during the development of this specification.
       </p>
+      <div class="correction">
+        This specification includes <a href=
+        "https://www.w3.org/Consortium/Process/#candidate-correction">Candidate
+        corrections</a>, which are marked in the document.
+      </div>
       <section>
         <h3>
           Changes since last publication
@@ -1706,6 +1711,16 @@
       <h2>
         <dfn>PaymentItem</dfn> dictionary
       </h2>
+      <div class="correction" id="c1">
+        <span class="marker">Candidate Correction 1:</span> This dictionary
+        should include definitions for optional members "language" and
+        "direction" to support use cases where document context is insufficient
+        for the proper internationalization of "label". Values for these
+        members would be the same as for "lang" and "dir" in HTML. The Working
+        Group is monitoring progress on <a href=
+        "https://github.com/w3c/string-meta/issues/54">Localizable in
+        WebIDL</a>.
+      </div>
       <pre class="idl">
         dictionary PaymentItem {
           required DOMString label;
@@ -1945,6 +1960,16 @@
           <h3>
             <dfn>PaymentValidationErrors</dfn> dictionary
           </h3>
+          <div class="correction" id="c2">
+            <span class="marker">Candidate Correction 2:</span> This dictionary
+            should include definitions for optional members "language" and
+            "direction" to support use cases where document context is
+            insufficient for the proper internationalization of "error". Values
+            for these members would be the same as for "lang" and "dir" in
+            HTML. The Working Group is monitoring progress on <a href=
+            "https://github.com/w3c/string-meta/issues/54">Localizable in
+            WebIDL</a>.
+          </div>
           <pre class="idl">
             dictionary PaymentValidationErrors {
               DOMString error;
--- a/index.html
+++ b/index.html
@@ -111,11 +111,6 @@
         There has been no change in dependencies on other workings groups
         during the development of this specification.
       </p>
-      <p>
-        This specification includes <a href=
-        "https://www.w3.org/2020/Process-20200915/#candidate-addition">Candidate
-        additions</a>, which are marked in the document.
-      </p>
       <section>
         <h3>
           Changes since last publication
@@ -922,9 +917,7 @@
                 sheet is presented to end users. Instead, the API relies on
                 localization information inherited from the document. The
                 working group is considering support for requesting specific
-                localization of the user interface and of individual
-                {{PaymentItems}} and user-facing error messages in a future
-                version of this API.
+                localization of the user interface.
               </p>
             </aside>
           </li>
@@ -1711,16 +1704,6 @@
       <h2>
         <dfn>PaymentItem</dfn> dictionary
       </h2>
-      <aside class="addition" id="a1">
-        <p>
-        <span class="marker">Candidate Addition 1:</span>
-         This dictionary might include definitions for optional members `lang` and
-        `dir` to indicate the language and base direction of the {{PaymentItem/label}} member. 
-        The Working Group is monitoring progress on <a href=
-        "https://github.com/whatwg/webidl/issues/1025">
-        WebIDL issue 1025</a>. 
-        </p>
-      </aside>
       <pre class="idl">
         dictionary PaymentItem {
           required DOMString label;
@@ -1740,6 +1723,15 @@
         <dd>
           A human-readable description of the item. The <a>user agent</a> may
           display this to the user.
+          <aside class="note" title="Internationalization of the label">
+            <p>
+              The language and direction of the {{PaymentItem/label}} may often
+              be determined from information inherited from the document.
+              However, this approach may not suffice for some use cases. The
+              working group intends to fix this in a future version by aligning
+              with general approaches for the Web that are in development.
+            </p>
+          </aside>
         </dd>
         <dt>
           <dfn>amount</dfn> member
@@ -1960,16 +1952,6 @@
           <h3>
             <dfn>PaymentValidationErrors</dfn> dictionary
           </h3>
-
-          <aside class="addition" id="a2">
-           <p><span class="marker">Candidate Addition 1:</span>
-           This dictionary might include definitions for optional members `lang` and
-           `dir` to indicate the language and base direction of the {{PaymentValidationErrors/error}} member. 
-           The Working Group is monitoring progress on <a href=
-            "https://github.com/whatwg/webidl/issues/1025">
-             WebIDL issue 1025</a>. 
-           </p>
-          </aside>
           <pre class="idl">
             dictionary PaymentValidationErrors {
               DOMString error;
@@ -1988,6 +1970,16 @@
               general overview of validation issues, or it can be passed in
               combination with other members of the {{PaymentValidationErrors}}
               dictionary.
+              <aside class="note" title="Internationalization of the error">
+                <p>
+                  The language and direction of the
+                  {{PaymentValidationErrors/error}} may often be determined
+                  from information inherited from the document. However, this
+                  approach may not suffice for some use cases. The working
+                  group intends to fix this in a future version by aligning
+                  with general approaches for the Web that are in development.
+                </p>
+              </aside>
             </dd>
             <dt>
               <dfn>paymentMethod</dfn> member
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1706,6 +1711,16 @@ <h2>
...
+      <div class="correction" id="c1">
+        <span class="marker">Candidate Correction 1:</span> This dictionary
+        should include definitions for optional members "language" and
+        "direction" to support use cases where document context is insufficient
+        for the proper internationalization of "label". Values for these
+        members would be the same as for "lang" and "dir" in HTML. The Working
+        Group is monitoring progress on <a href=
+        "https://github.com/w3c/string-meta/issues/54">Localizable in
+        WebIDL</a>.
+      </div>
```
```suggestion
      <aside class="correction" id="c1">
        <p>
        <span class="marker">Candidate Correction 1:</span>
         This dictionary might include definitions for optional members `lang` and
        `dir` to indicate the language and text-direction override of {{PaymentItem/label}}. 
        The Working Group is monitoring progress on <a href=
        "https://github.com/whatwg/webidl/issues/1025">
        WebIDL issue 1025</a>. 
        </p>
      </aside>
```

index.html
```diff
@@ -1945,6 +1960,16 @@ <h2>
...
+          <div class="correction" id="c2">
+            <span class="marker">Candidate Correction 2:</span> This dictionary
+            should include definitions for optional members "language" and
+            "direction" to support use cases where document context is
+            insufficient for the proper internationalization of "error". Values
+            for these members would be the same as for "lang" and "dir" in
+            HTML. The Working Group is monitoring progress on <a href=
+            "https://github.com/w3c/string-meta/issues/54">Localizable in
+            WebIDL</a>.
+          </div>
```
```suggestion

          <aside class="correction" id="c2">
           <p><span class="marker">Candidate Correction 1:</span>
           This dictionary might include definitions for optional members `lang` and
           `dir` to indicate the language and text-direction override of {{PaymentValidationErrors/error}}. 
           The Working Group is monitoring progress on <a href=
            "https://github.com/whatwg/webidl/issues/1025">
             WebIDL issue 1025</a>. 
           </p>
          </aside>
```

index.html
```diff
@@ -111,6 +111,11 @@ <h1 id="title">
         There has been no change in dependencies on other workings groups
         during the development of this specification.
       </p>
+      <div class="correction">
+        This specification includes <a href=
+        "https://www.w3.org/Consortium/Process/#candidate-correction">Candidate
+        corrections</a>, which are marked in the document.
+      </div>
```
marcoscaceres: My reading of the process document is that these would be classified as "candidate additions", not corrections. 

```suggestion
      <p>
        This specification includes <a href=
        "https://www.w3.org/2020/Process-20200915/#candidate-addition">Candidate
        additions</a>, which are marked in the document.
      </p>
```

marcoscaceres (changes requested):
I don't think it's sufficient to just state that `dir` and `lang` will, be added... we need to actually specify them properly, like saying if `dir` will be an enum, and if an invalid language tag will throw (I guess that's part of https://github.com/whatwg/webidl/issues/1025). 

We can keep building on this pull request tho.

index.html
```diff
@@ -1706,6 +1711,16 @@ <h2>
       <h2>
         <dfn>PaymentItem</dfn> dictionary
       </h2>
+      <aside class="correction" id="c1">
+        <p>
+        <span class="marker">Candidate Correction 1:</span>
```
```suggestion
        <span class="marker">Candidate Addition 1:</span>
```

index.html
```diff
@@ -1945,6 +1960,16 @@ <h2>
           <h3>
             <dfn>PaymentValidationErrors</dfn> dictionary
           </h3>
+
+          <aside class="correction" id="c2">
+           <p><span class="marker">Candidate Correction 1:</span>
```
```suggestion
           <p><span class="marker">Candidate Addition 1:</span>
```

marcoscaceres:
Here is a concrete plan:
 
 * First: get agreement on platform wide `dir` and `lang` definition in https://github.com/whatwg/webidl/issues/1025
 * Based on https://github.com/whatwg/webidl/issues/1025, update this spec to do whatever we decide there.
 * If it's adding `dir` and `lang`, add them to the spec here.
 * File bugs on implementations to see if we can get implementation commitment (I can jump into WebKit and/or Chromium if needed).

ianbjacobs:
@marcoscaceres, quick clarification regarding your plan. Are these steps you recommend after publication of the Rec with this pull request integrated?

marcoscaceres:
After Rec, as these are additions for a Recommendation.

stephenmcgruer (approved):
I'm ok with this approach. I don't think I agree with [Marcos' comment](https://github.com/w3c/payment-request/pull/971#pullrequestreview-772142951) that we need to include how `lang` and `dir` would work in detail at this point, but I might be persuaded if you can explain your reasoning @marcoscaceres  :)

(I also don't mind if we **do** include details, if Ian wants to do the extra work.)

marcoscaceres:
> I might be persuaded if you can explain your reasoning @marcoscaceres :)

So, we wouldn't merge this until **after** REC, and we would only merge this if our group's criteria is met: we would fully spec this out, add tests, and get implementation commitment (same as any feature we've added to the spec in the past).

index.html
```diff
@@ -1706,6 +1711,16 @@ <h2>
       <h2>
         <dfn>PaymentItem</dfn> dictionary
       </h2>
+      <aside class="addition" id="a1">
+        <p>
+        <span class="marker">Candidate Addition 1:</span>
+         This dictionary might include definitions for optional members `lang` and
+        `dir` to indicate the language and text-direction override of {{PaymentItem/label}}.
```
aphillips: It wouldn't be "text-direction override". It would be better to say "base direction" or "base text-direction".

index.html
```diff
@@ -1945,6 +1960,16 @@ <h2>
           <h3>
             <dfn>PaymentValidationErrors</dfn> dictionary
           </h3>
+
+          <aside class="addition" id="a2">
+           <p><span class="marker">Candidate Addition 1:</span>
+           This dictionary might include definitions for optional members `lang` and
+           `dir` to indicate the language and text-direction override of {{PaymentValidationErrors/error}}.
```
aphillips: same comment as above.

index.html
```diff
@@ -1706,6 +1711,16 @@ <h2>
       <h2>
         <dfn>PaymentItem</dfn> dictionary
       </h2>
+      <aside class="addition" id="a1">
+        <p>
+        <span class="marker">Candidate Addition 1:</span>
+         This dictionary might include definitions for optional members `lang` and
```
aphillips: we generally prefer `language` and `direction` for data values and `lang` and `dir` for attribute values, but changing that would be optional, as there is a certain harmony to using the same names as the HTML attributes. I'd just leave it for now.

aphillips (approved):
I think this covers the need to call out a potential change while we figure out whether to make it and what form it should take.

index.html
ianbjacobs: Here is my mental model:

* The document (e.g., in the body element) specifies language and base direction for text in the document.
* If the API is called without any dir/lang metadata, the base direction and language come from the document.
* If the API is called with the dir/lang metadata, they override the base direction and language.

I think that's where the "override" text comes from. If that is not the right mental model and we should just say base direction, that will be fine.

index.html
ianbjacobs: (see above request for clarification)

index.html
```suggestion
        `dir` to indicate the language and base direction of the {{PaymentItem/label}} member. 
```

index.html
```suggestion
           `dir` to indicate the language and base direction of the {{PaymentValidationErrors/error}} member. 
```

index.html
marcoscaceres: @aphillips is right... it's the base direction of the label itself.

ianbjacobs:
Based on joint I18N/WPWG discussion today [1] I anticipate that we will not add dir/lang to Payment Request. I believe there is now support for publishing the Recommendation with informative hooks that indicate that internationalization for these two strings needs to be fixed and that we are part of a Web community discussion about how to do that with a platform-wide solution. I took an action to work with @r12a on text for the specification. 

[1] https://www.w3.org/2021/10/26-wpwg-minutes.html#t03

ianbjacobs:
@r12a and @aphillips, I have updated this pull request based on the 26 Oct joint discussion. Per @r12a's comments during the meeting, I have characterized this as (1) a fix and (2) pending progress on the platform wide solution. I welcome your review of this and any suggestions.

marcoscaceres (approved):
I’m approving these, but I’d like to go on the record that I disagree with the additions as they might  set false expectations for readers.

I am broadly supportive of i18n solutions, but if OS-level payment handlers don’t ever end up supporting dir/lang (and there is no implementation backing), then there is not much the W3C or working group can do about that (we can “recommend” things, but we can’t force anyone to implement them). 

I hope that changes in the future through real-world demand. I would have preferred we continue to advocate for the better i18n support as we’ve done with other features in this specification: through close collaboration with implementers and sites/services that are deploying the API at scale.

aphillips (approved):
The I18N WG reviewed this PR in our teleconference of 2021-11-04 and I drew an [action item](https://www.w3.org/International/track/actions/1092) to approve it.

ianbjacobs:
@aphillips, many thanks to you and the I18N Working Group for working with us on this and other issues. We really do appreciate the engagement.

PR #976, opened 2021-11-14
Proposed addition: validate .data on construction

marcoscaceres:
closes #975 

The following tasks have been completed:

 * [ ] <del>Modified Web platform tests</del> - not testable, as it's for proprietary/arbitrary method data types.   
 
Implementation commitment:

 * [X] [WebKit](https://webkit.org/b/233292) - [patch](https://trac.webkit.org/changeset/286452/webkit)
 * [ ] Chromium (commitment not required - it's for arbitrary method data types)
 * [ ] Gecko (commitment not required - it's for arbitrary method data types)

Optional, impact on Payment Handler spec?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/976.html" title="Last updated on Dec 2, 2021, 10:27 PM UTC (76dc5d3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/976/b4a10f6...76dc5d3.html" title="Last updated on Dec 2, 2021, 10:27 PM UTC (76dc5d3)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/976.html" title="Last updated on Feb 9, 2022, 8:59 AM UTC (76dc5d3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/976/b4a10f6...76dc5d3.html" title="Last updated on Feb 9, 2022, 8:59 AM UTC (76dc5d3)">Diff</a>

PR #977, opened 2021-11-19
Define concepts for converting and validating `.data`

stephenmcgruer:
This PR introduces two new definitions (`dfn`) for both the IDL type that
`PaymentMethodData/data` should be converted into, and for the validation
steps that a payment method will perform on it. This allows payment method
specs to define these in a way that is cross-linkable to the PaymentRequest
specification.

This change is mostly editorial, **except** that it allows for arbitrary error types
to be thrown during validation (previously only `TypeError` would be thrown).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/977.html" title="Last updated on Jul 6, 2022, 1:30 PM UTC (05f8d67)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/977/b6b0630...05f8d67.html" title="Last updated on Jul 6, 2022, 1:30 PM UTC (05f8d67)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -149,6 +149,30 @@
         </li>
       </ul>
       <p>
+        A payment method defines:
+      </p>
+      <dl>
+        <dt>
+          <dfn>Payment Method additional data type</dfn>
+        </dt>
+        <dd>
+          An IDL type that defines the data the <a>payment method</a> expects
+          to receive in the {{PaymentMethodData/data}} member of the
+          {{PaymentMethodData}}. If not specified for a given payment method,
+          no conversion to IDL is done and the payment method will receive
+          {{PaymentMethodData/data}} as JSON.
+        </dd>
+        <dt>
+          <dfn>Steps to validate payment method data</dfn>
+        </dt>
+        <dd>
+          How a <a>payment method</a> validates the {{PaymentMethodData/data}}
+          member of the {{PaymentMethodData}}, after it is converted to the
+          <a>payment method additional data type</a>. If not specified for a
+          given payment method, no validation is done.
+        </dd>
+      </dl>
+      <p>
         The details of how to fulfill a payment request for a given <a>payment
         method</a> is an implementation detail of a <dfn data-export="">payment
         handler</dfn>, which is an application or service that handles requests
@@ -573,27 +597,28 @@
                   |paymentMethod|.{{PaymentMethodData/data}} into a string.
                   Rethrow any exceptions.
                   </li>
-                  <li>If |serializedData| is not null, and if required by the
-                  specification that defines the
-                  |paymentMethod|.{{PaymentMethodData/supportedMethods}}:
+                  <li>If |serializedData| is not null, and if the specification
+                  that defines the
+                  |paymentMethod|.{{PaymentMethodData/supportedMethods}}
+                  specifies a <a>payment method additional data type</a>:
                     <ol>
                       <li>Let |object| be the result of <a data-cite=
                       "ECMASCRIPT#sec-json.parse">JSON-parsing</a>
                       |serializedData|.
                       </li>
                       <li>
-                        <p>
-                          Let |idl| be the result of [=converted to an IDL
+                        <p>Let |idl| be the result of [=converted to an IDL
                           value|converting=] |object| to an IDL value of the
-                          type specified by the specification that defines the
-                          |paymentMethod|.{{PaymentMethodData/supportedMethods}}.
-                          Rethrow any exceptions.
+                          type specified by the <a>payment method additional
+                          data type</a>. Rethrow any exceptions.
                         </p>
+                      </li>
+                      <li>
                         <p>
-                          If required by the specification that defines the
-                          |paymentMethod|.{{PaymentMethodData/supportedMethods}},
-                          validate the members of |idl|. If a member's value is
-                          invalid, throw a {{TypeError}}.
+                          Run the <a>steps to validate payment method data</a>,
+                          if any, from the specification that defines the
+                          |paymentMethod|.{{PaymentMethodData/supportedMethods}}
+                          on |object|. Rethrow any exceptions.
                         </p>
                         <p class="note">
                           These step assures that any IDL type conversion and
@@ -855,10 +880,11 @@
               "ECMASCRIPT#sec-json.parse">JSON-parsing</a> the second element
               in the |paymentMethod| tuple.
               </li>
-              <li>If required by the specification that defines the
-              |identifier|, then [=converted to an IDL value|convert=] |data|
-              to an IDL value of the type specified there. Otherwise,
-              [=converted to an IDL value|convert=] to {{object}}.
+              <li>If the specification that defines the |identifier| specifies a
+              <a>payment method additional data type</a>, then [=converted to an
+              IDL value|convert=] |data| to an IDL value of that type.
+              Otherwise, [=converted to an IDL value|convert=] |data| to
+              {{object}}.
               </li>
               <li>If conversion results in an <a>exception</a> |error|:
                 <ol>
--- a/index.html
+++ b/index.html
@@ -153,23 +153,24 @@
       </p>
       <dl>
         <dt>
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
@marcoscaceres - fyi this is my suggestion for adding a `dfn` for https://github.com/w3c/payment-request/pull/976, to make it easier for other specifications to define their behaviour for it.

index.html
```diff
@@ -573,27 +582,22 @@ <h2>
...
-                        </p>
-                        <p>
-                          If required by the specification that defines the
-                          |paymentMethod|.{{PaymentMethodData/supportedMethods}},
-                          validate the members of |idl|. If a member's value is
-                          invalid, throw a {{TypeError}}.
+                          Run the <a>steps to convert and validate payment
+                          method data</a> from the specification that defines the
+                          |paymentMethod|.{{PaymentMethodData/supportedMethods}}
+                          on |object|. Rethrow any exceptions.
```
marcoscaceres: Nit... Maybe rewritten as the following, to be a bit consistent with the stuff up on line 585: 
```suggestion
                          If provided by the specification that defines the
                          |paymentMethod|.{{PaymentMethodData/supportedMethods}}, 
                          run the <a>steps to convert and validate payment
                          method data</a> on |object|. Rethrow any exceptions.
```

index.html
```diff
@@ -155,6 +155,15 @@ <h2>
         for payment. Concretely, a payment handler defines:
       </p>
       <dl>
+        <dt>
+          <dfn>Steps to convert and validate payment method data</dfn>
+        </dt>
+        <dd>
+          How a <a>payment method</a> [=converted to an IDL value|converts=] and
```
marcoscaceres: Hmm... I think conversion and validation need to be separate things. Conversion doesn't need its own steps: it's clearly defined by the WebIDL step (i.e., it's redundant and shouldn't be part of this). Validation, on the other hand should be specified by the other specification.

marcoscaceres (changes requested):
Validation should be separate from conversion...

index.html
stephenmcgruer: I'm sympathetic to that argument. I wrote it like this for two reasons:

1. It was awkward to find the words to say "if a spec says to convert it, then do, and also if a spec has `steps to validate the converted data`, do those too."
2. For SPC, I was actually hoping to normalize the data as part of 'conversion' (see what is currently step 5 of https://w3c.github.io/secure-payment-confirmation/#sctn-steps-to-check-if-a-payment-can-be-made), which means that it takes on a meaning beyond the basic WebIDL meaning.

I'm happy to poke at the wording here myself, but interested in your thoughts on 2 :). (A reasonable counter-argument is to point out that we're likely to move away from the PR API for SPC anyway, and so it shouldn't be bent to make SPC work better!)

marcoscaceres: Hmm...  yeah, that algorithm (2) seems to overstep a bit... it's doing checking, but it's also mutating an object. Editorially, I'd recommend splitting that (personally, that's come back to bite me quite a few times). 

Nevertheless, I get what you are trying to do now... but I'd still urge your to split those into two if possible.

stephenmcgruer: Ack, I agree. Coming back to this, I've split conversion and validation back out, and also made them idempotent (i.e., no normalization). This means that SPC will have to continue to normalize in step 5 of https://w3c.github.io/secure-payment-confirmation/#sctn-steps-to-check-if-a-payment-can-be-made, but I think that's fine. (And I very much agree with your concerns about overstepping).

Please take another look :)

stephenmcgruer:
cc @ianbjacobs as landing this blocks https://github.com/w3c/secure-payment-confirmation/issues/194

The build is currently failing because the upstream of this PR (https://github.com/w3c/payment-request/pull/976) is behind HEAD and needs updated to include https://github.com/w3c/payment-request/commit/34c206fdb8f88728ba1e9198d6bceb08860ed29f

index.html
```diff
@@ -148,6 +148,30 @@ <h2>
...
+        </dt>
+        <dd>
+          An IDL type that defines the data the <a>payment method</a> expects
+          to receive in the {{PaymentMethodData/data}} member of the
+          {{PaymentMethodData}}. If not specified for a given payment method,
+          no conversion to IDL is done and the payment method will receive
+          {{PaymentMethodData/data}} as JSON.
+        </dd>
+        <dt>
+          <dfn>Steps to validate payment method data</dfn>
```
marcoscaceres, line 166: Hmmm... this doesn't actually define any steps? Maybe we don't need this definition?

index.html
```diff
@@ -148,6 +148,30 @@ <h2>
...
+      <dl>
+        <dt>
+          <dfn>Payment Method additional data type</dfn>
+        </dt>
+        <dd>
+          An IDL type that defines the data the <a>payment method</a> expects
+          to receive in the {{PaymentMethodData/data}} member of the
+          {{PaymentMethodData}}. If not specified for a given payment method,
+          no conversion to IDL is done and the payment method will receive
+          {{PaymentMethodData/data}} as JSON.
```
```suggestion
          Optionally, an IDL type that the <a>payment method</a> expects
          to receive as the {{PaymentMethodData}}'s {{PaymentMethodData/data}} member. 
          If not specified for a given payment method,
          no conversion to IDL is done and the payment method will receive
          {{PaymentMethodData/data}} as JSON.
```

index.html
```diff
@@ -148,6 +148,30 @@ <h2>
...
+        </dd>
+        <dt>
+          <dfn>Steps to validate payment method data</dfn>
+        </dt>
+        <dd>
+          How a <a>payment method</a> validates the {{PaymentMethodData/data}}
+          member of the {{PaymentMethodData}}, after it is converted to the
+          <a>payment method additional data type</a>. If not specified for a
+          given payment method, no validation is done.
```
```suggestion
          Algorithmic steps that specify how a <a>payment method</a> validates the {{PaymentMethodData/data}}
          member of the {{PaymentMethodData}}, after it is converted to the
          <a>payment method additional data type</a>. If not specified for a
          given payment method, no validation is done.
```

index.html
```diff
@@ -855,10 +880,11 @@ <h2>
               "ECMASCRIPT#sec-json.parse">JSON-parsing</a> the second element
               in the |paymentMethod| tuple.
               </li>
-              <li>If required by the specification that defines the
-              |identifier|, then [=converted to an IDL value|convert=] |data|
-              to an IDL value of the type specified there. Otherwise,
-              [=converted to an IDL value|convert=] to {{object}}.
+              <li>If the specification that defines the |identifier| specifies a
+              <a>payment method additional data type</a>, then [=converted to an
```
```suggestion
              [=payment method additional data type=], then [=converted to an
```

index.html
```diff
@@ -148,6 +148,30 @@ <h2>
         method.
         </li>
       </ul>
+      <p>
+        A payment method defines:
+      </p>
+      <dl>
+        <dt>
+          <dfn>Payment Method additional data type</dfn>
```
marcoscaceres: I think this should be defined as: 

```suggestion
          <dfn data-dfn-for="Payment Method">additional data type</dfn>
```

And then cited as:

```
[=Payment Method/additional data type=]
```

marcoscaceres (dismissed):
Noting this makes a couple of normative changes... in particular, it no longer exclusively throws a `TypeError` during the processing steps.

index.html
stephenmcgruer: Done.

index.html
stephenmcgruer: Done.

index.html
stephenmcgruer: Done.

index.html
stephenmcgruer: Done.

stephenmcgruer:
> Noting this makes a couple of normative changes... in particular, it no longer exclusively throws a TypeError during the processing steps.

Good catch - thanks for noticing this! I've updated the PR description and title (along with addressing your other comments) to reflect that new errors may be thrown.

marcoscaceres (approved):

PR #978, opened 2021-11-19
Editorial: Changes in response to AC review

ianbjacobs:
This is a set of changes proposed to address a formal objection raised during AC review of the specification. My expectation is first to determine whether there is consensus within the WG to adopt these changes in the Recommendation.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/978.html" title="Last updated on Dec 3, 2021, 9:49 PM UTC (a30e7ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/978/10e7a6d...a30e7ac.html" title="Last updated on Dec 3, 2021, 9:49 PM UTC (a30e7ac)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -870,9 +870,12 @@
                 </ol>
               </li>
               <li>Let |registeredHandlers| be a <a>list</a> of registered
-              payment handlers for the payment method |identifier|. Payment
-              handler registration merchanisms are outside the scope of this
-              specification. For one mechanism, see [[[payment-handler]]].
+              payment handlers for the payment method |identifier|.
+                <aside class="note" title="Payment Handler registration">
+                  Merchanisms to register a payment handler are outside the
+                  scope of this specification. For one mechanism, see
+                  [[[payment-handler]]].
+                </aside>
               </li>
               <li>For each |handler| in |registeredHandlers|:
                 <ol>
--- a/index.html
+++ b/index.html
@@ -984,7 +984,7 @@
             </p>
             <p>
               Decisions on behalf of the user (e.g., steps 6 and 12) MUST NOT
-              limit competition.
+              restrict competition.
             </p>
             <aside class="example" title=
             "Handling of multiple applicable modifiers">
--- a/index.html
+++ b/index.html
@@ -982,10 +982,6 @@
               say, an item at the end of the list always takes precedence over
               any item at the beginning of the list (see example below).
             </p>
-            <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -870,7 +870,9 @@ <h2>
                 </ol>
               </li>
               <li>Let |registeredHandlers| be a <a>list</a> of registered
-              payment handlers for the payment method |identifier|.
+              payment handlers for the payment method |identifier|. Payment
+              handler registration merchanisms are outside the scope of this
+              specification. For one mechanism, see [[[payment-handler]]].
```
```suggestion
              payment handlers for the payment method |identifier|. 
              <aside class="note" title="Payment Handler registration">
              Merchanisms to register a Payment handler are outside the scope of this
              specification. For one mechanism, see [[[payment-handler]]].
              </aside>
```

index.html
```diff
@@ -977,6 +982,10 @@ <h2>
               say, an item at the end of the list always takes precedence over
               any item at the beginning of the list (see example below).
             </p>
+            <p>
+              Decisions on behalf of the user (e.g., steps 6 and 12) MUST NOT
+              restrict competition.
```
mountainhippo: I understand and support the intent of this change. However, I don't think a normative requirement around competition is the way to go because such a requirement would be fundamentally untestable.
My counter proposal would be to include a pointer to the [W3C Antitrust and Competition Guidance](https://www.w3.org/Consortium/Legal/2017/antitrust-guidance) in the "Status of this document" section at the beginning of the spec, an informative note directly above Example 7 in the spec text (so replacing your change at line 985) explicitly noting that the algorithm has the potential to be used in an anti-competitive way , and a reference pointer in D.1 Normative references to the W3C Antitrust and Competition Guidance policy (which I would justify as a normative reference due to the "MUST" statement here -> " Participants must ensure that their conduct does not violate antitrust and competition laws and regulations. "

marcoscaceres: Agree.

lbdvt: Thanks for the proposal, I agree.

ianbjacobs: Based on further discussion within the staff I have removed this piece of the pull request.

ianbjacobs:
Following a WG decision, I am going to merge this PR.
https://lists.w3.org/Archives/Public/public-payments-wg/2021Dec/0011.html

marcoscaceres (approved):

PR #982, opened 2022-01-19
passing data on complete()

### LINKED ISSUE #981: Allow custom data for the active payment method to be provided to `PaymentResponse.prototype.complete`
It'd be great for there to be a standardized way for developers/merchants to provide data back to the `PaymentRequest` after it's been accepted.  This would allow additional information to be provided back to the underlying platform payment method handler/UI.

I was thinking it'd be something like this:
```webidl
partial interface PaymentResponse {
  [NewObject] Promise<undefined> complete(optional PaymentComplete result = "unknown", optional PaymentMethodData paymentMethod);
}
```
or maybe even just
```webidl
partial interface PaymentResponse {
  [NewObject] Promise<undefined> complete(optional PaymentComplete result = "unknown", optional object paymentMethodData);
}
```

Alternatively, we maybe might want to wrap the `PaymentComplete` in a new IDL `dictionary` (just like `PaymentValidationErrors`), which would also allow for more extension in the future if desired.
```webidl
enum PaymentResultStatus {
  "fail",
  "success",
  "unknown"
}

dictionary PaymentResult {
    PaymentResultStatus status = "unknown";
    PaymentMethodData paymentMethod;
}

partial interface PaymentResponse {
  [NewObject] Promise<undefined> complete(optional PaymentResultStatus = "unknown"); // Would we still want/need this?
  [NewObject] Promise<undefined> complete(optional PaymentResult);
}
```

Though (in both cases) I'm not sure if we want just a `PaymentMethodData` or a `sequence<PaymentMethodData>` so that a developer/merchant can declaratively provide data back to all supported payment methods at once instead of having to adjust the data based on which payment method is currently being used.

marcoscaceres:
closes #981 

Implementation commitment:

 * [x] WebKit (link to issue)
 * [ ] Chromium (link to issue)
 * [ ] Gecko (link to issue)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/982.html" title="Last updated on Jul 6, 2022, 2:53 AM UTC (3eed041)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/982/b6b0630...3eed041.html" title="Last updated on Jul 6, 2022, 2:53 AM UTC (3eed041)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -95,7 +95,7 @@
         The working group maintains <a href=
         "https://github.com/w3c/payment-request/issues?utf8=%E2%9C%93&amp;q=is%3Aopen%20is%3Aissue%20-label%3A%22Priority%3A%20Postponed%22%20">
         a list of all bug reports that the group has not yet addressed</a>.
-        Pull requests with proposed-addition specification text for outstanding
+        Pull requests with proposed specification text for outstanding
         issues are strongly encouraged.
       </p>
       <p>
--- a/index.html
+++ b/index.html
@@ -95,8 +95,8 @@
         The working group maintains <a href=
         "https://github.com/w3c/payment-request/issues?utf8=%E2%9C%93&amp;q=is%3Aopen%20is%3Aissue%20-label%3A%22Priority%3A%20Postponed%22%20">
         a list of all bug reports that the group has not yet addressed</a>.
-        Pull requests with proposed specification text for outstanding
-        issues are strongly encouraged.
+        Pull requests with proposed specification text for outstanding issues
+        are strongly encouraged.
       </p>
       <p>
         The working group will demonstrate implementation experience by
--- a/index.html
+++ b/index.html
@@ -1761,7 +1761,7 @@
         </dd>
       </dl>
     </section>
-    <section data-dfn-for="PaymentCompleteDetails" class="addition proposed">
+    <section data-dfn-for="PaymentCompleteDetails">
       <h2>
         <dfn>PaymentCompleteDetails</dfn> dictionary
       </h2>
--- a/index.html
+++ b/index.html
@@ -95,8 +95,8 @@
         The working group maintains <a href=
         "https://github.com/w3c/payment-request/issues?utf8=%E2%9C%93&amp;q=is%3Aopen%20is%3Aissue%20-label%3A%22Priority%3A%20Postponed%22%20">
         a list of all bug reports that the group has not yet addressed</a>.
-        Pull requests with proposed specification text for outstanding issues
-        are strongly encouraged.
+        Pull requests with proposed-addition specification text for outstanding
+        issues are strongly encouraged.
       </p>
       <p>
         The working group will demonstrate implementation experience by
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2084,6 +2116,40 @@ <h2>
...
           <li>Set |response|.{{PaymentResponse/[[complete]]}} to true.
           </li>
+          <li class="addition proposed">Let |serializedData| be the result of
+          <a>JSON-serializing</a> |details|.{{PaymentCompleteDetails/data}}
+          into a string.
+          </li>
+          <li class="addition proposed">If serializing [=exception/throws=] an
+          exception, return <a>a promise rejected with</a> that exception.
+            <div class="issue">
+              What should happen if an exception is thrown here? Should we
```
dcrousso: I think we might want to wait to set `response.[[complete]]` to `true` until after we JSON serialize, re-parse, and convert to IDL so that if something happens we can re-call `complete()` (as otherwise the step 2 would always fail).

Also, I wonder if we should wait to set `response.[[complete]]` until right before we resolve the returned promise so that if for some reason the user agent finds issue with the data when it attempts to close the payment UI then the developer can still recover (just like the above)?

marcoscaceres: Agree on point 1 (being able to recover). My worry was that an unhandled promise rejection would leave the payment sheet hanging open, but the browser can either auto-close it after some time or provide some other means for the user to close it. 

About point 2, we could add some thing state "processing" or whatever, but it might be best to hold off until there is a concrete situation where that might happen. In theory, doing the IDL conversion steps should catch any issues... oh! That reminds me, we should do `.data` validation at this point also (same as we did in https://github.com/w3c/payment-request/pull/976).

marcoscaceres:
@dcrousso, ok... we now do IDL conversion and member validation early, giving the developer an opportunity to recover. And the [[\complete]] = true now happens after the checks.

index.html
```diff
@@ -2082,6 +2114,43 @@ <h2>
...
           <li>Let |promise:Promise| be <a>a new promise</a>.
           </li>
+          <li class="addition proposed">Let |serializedData| be the result of
+          <a>JSON-serializing</a> |details|.{{PaymentCompleteDetails/data}}
+          into a string.
+          </li>
+          <li class="addition proposed">If serializing [=exception/throws=] an
+          exception, return <a>a promise rejected with</a> that exception.
+          </li>
+          <li class="addition proposed">If required by the specification that
```
```suggestion
          <li>Let |serializedData| be the result of
          <a>JSON-serializing</a> |details|.{{PaymentCompleteDetails/data}}
          into a string.
          </li>
          <li>If serializing [=exception/throws=] an
          exception, return <a>a promise rejected with</a> that exception.
          </li>
          <li>If required by the specification that
```

index.html
```diff
@@ -2107,7 +2175,8 @@ <h2>
           <li>Otherwise:
             <ol>
               <li>Close down any remaining user interface. The <a>user
-              agent</a> MAY use the value |result| to influence the user
+              agent</a> MAY use the value |result| and <ins class=
+              "addition proposed">|serializedData|</ins> to influence the user
```
```suggestion
              agent</a> MAY use the value |result| and |serializedData| to influence the user
```

marcoscaceres:
@stephenmcgruer, @rsolomakhin, would appreciate your review on this addition?

rsolomakhin:
Is there a link to the WebKit issue available?

stephenmcgruer:
> Is there a link to the WebKit issue available?

I found https://bugs.webkit.org/show_bug.cgi?id=235415 and https://github.com/WebKit/WebKit/commit/2fbfd75778b823914bef59e83d4fb9dc4ddc95e4, but looks like the 'real' issue is an internal Apple one (which is of course ok :))

marcoscaceres:
Yeah, this just adds optional proprietary behavior for completing.

stephenmcgruer (approved):
LGTM (and @rsolomakhin )

I personally am a fan of having defined concepts rather than prose where possible (e.g., I don't like "an IDL value of the type specified by the specification"), but I acknowledge (a) that that may just be my opinion and (b) that at least for now no public specification exists that will link to this.

Putting on my Chrome hat, I believe we'll be happy to make the IDL change in our implementation though we will likely not do anything with the passed data.

Updating the WPT tests would be great! :)

marcoscaceres:
Thanks @stephenmcgruer! I agree with you about "an IDL value of the type specified by the specification". If eventually we have a spec to link to, will definitely update. 

> Updating the WPT tests would be great!

We should get some tests for free from WPT once we merge (at least for the method signature length).

I'm not sure it's worth adding a manual test, but can add one if you think it's worth it.

stephenmcgruer:
> We should get some tests for free from WPT once we merge (at least for the method signature length).

Of course, I completely forgot about idlharness tests >_<. (Which I really shouldn't, given my history in WPT!). I think that is sufficient for this use-case, thanks :)

PR #1005, opened 2023-04-13
[Spec] Only allow show() to be called in a foreground tab

stephenmcgruer:
This addresses this issue in SPC - https://w3c.github.io/secure-payment-confirmation/#issue-e92619e9

The behavior spec'd in SPC appears to already be followed in general Payment Request implementations already, based on testing in both Safari and Chrome, so it seems safe to just spec it here instead.

 * [x] Modified Web platform tests ([link](https://github.com/web-platform-tests/wpt/pull/40256))
 
Implementation commitment:

 * [x] WebKit (already passes test)
 * [x] Chromium (already passes test)
 * [x] Gecko (doesn't implement Payment Request)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/1005.html" title="Last updated on May 30, 2023, 1:19 PM UTC (1e7e2d4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/1005/08be866...1e7e2d4.html" title="Last updated on May 30, 2023, 1:19 PM UTC (1e7e2d4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -839,7 +839,7 @@
           not [=Document/fully active=], then return <a>a promise rejected
           with</a> an {{"AbortError"}} {{DOMException}}.
           </li>
-          <li class="addition proposed">If |document|'s [=Document/visibility state=] is not `"visible"`,
+          <li>If |document|'s [=Document/visibility state=] is not `"visible"`,
           then return <a>a promise rejected with</a> an {{"AbortError"}}
           {{DOMException}}.
           <li>
--- a/index.html
+++ b/index.html
@@ -839,7 +839,7 @@
           not [=Document/fully active=], then return <a>a promise rejected
           with</a> an {{"AbortError"}} {{DOMException}}.
           </li>
-          <li>If |document|'s [=Document/visibilityState=] is not `"visible"`,
+          <li>If |document|'s [=Document/visibility state=] is not `"visible"`,
           then return <a>a promise rejected with</a> an {{"AbortError"}}
           {{DOMException}}.
           <li>
--- a/index.html
+++ b/index.html
@@ -839,6 +839,9 @@
           not [=Document/fully active=], then return <a>a promise rejected
           with</a> an {{"AbortError"}} {{DOMException}}.
           </li>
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
The failing build appears to be from a change in permissions policy spec that broke Payment Request build:

```
  Error: ROR] Couldn’t find “default allowlist“ in this document or other cited documents: [dom], [fetch], [html], [infra], [payment-method-id], [permissions-policy], [url], and [webidl].
```

rsolomakhin (approved):

ianbjacobs (approved):

marcoscaceres (approved):

index.html
```diff
@@ -839,6 +839,9 @@ <h2>
           not [=Document/fully active=], then return <a>a promise rejected
           with</a> an {{"AbortError"}} {{DOMException}}.
           </li>
+          <li>If |document|'s [=Document/visibility state=] is not `"visible"`,
```
marcoscaceres, line 842: I think we need to include this for updating a Rec:

```suggestion
          <li class="addition proposed">If |document|'s [=Document/visibility state=] is not `"visible"`,
```

stephenmcgruer: Hrm, I merged this initially, but now am less sure. Is the Payment Request repo not now in a '1.1' state (see https://w3c.github.io/payment-request/) where published changes !== changes to a Rec? 

For example, we didn't mark https://github.com/w3c/payment-request/pull/982/files as 'addition proposed' as far as I know. (Though that may also have been an oversight of course!).

WDYT? (cc @ianbjacobs too)

stephenmcgruer: I checked with @ianbjacobs this morning and he agreed, so I reverted this commit. I'm going to land the spec change without it, but if you have concerns @marcoscaceres then lets definitely follow up on this and all the other commits that have landed in 1.1 and see if we should mark them all :)

stephenmcgruer:
Build failure appears unrelated to this change, so I'm going to merge it.

PR #1009, opened 2023-06-12
[Spec] Relax user activation requirement for show()

nickburris:
This gives the user agent the ability to relax the user activation requirement on the PaymentRequest.show() method.

This spec change largely follows the relevant change outlined in Secure Payment Confirmation: https://github.com/w3c/secure-payment-confirmation/pull/236

***
<a href="https://pr-preview.s3.amazonaws.com/nickburris/payment-request/pull/1009.html" title="Last updated on Jun 27, 2023, 6:53 PM UTC (3f335a0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/1009/f5b07d1...nickburris:3f335a0.html" title="Last updated on Jun 27, 2023, 6:53 PM UTC (3f335a0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -822,15 +822,22 @@
           <li data-tests=
           "payment-request-show-method.https.html, show-method-postmessage-manual.https.html">
           If the [=relevant global object=] of [=request=] does not have
-          [=transient activation=]:
+          [=transient activation=], the user agent MAY:
             <ol>
               <li>Return [=a promise rejected with=] with a {{"SecurityError"}}
               {{DOMException}}.
               </li>
             </ol>
           </li>
-          <li data-tests="show-consume-activation.https.html">[=Consume user
-          activation=] of the [=relevant global object=].
+          <p class="note">
+            This allows the user agent to not require user activation, for
+            example to support redirect authentication flows where a user
+            activation may not be present upon redirect. See
+            <a href="#user-activation-requirement"></a> for security
+            considerations.
+          </p>
+          <li data-tests="show-consume-activation.https.html">Otherwise,
+          [=consume user activation=] of the [=relevant global object=].
           </li>
           <li>Let |document| be |request|'s [=relevant global object=]'s
           [=associated `Document`=].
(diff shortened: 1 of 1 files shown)


nickburris:
@rsolomakhin, PTAL for a first look! First time in this spec 😄

index.html
```diff
@@ -822,15 +822,22 @@ <h2>
...
-          [=transient activation=]:
+          [=transient activation=], the user agent MAY:
             <ol>
               <li>Return [=a promise rejected with=] with a {{"SecurityError"}}
               {{DOMException}}.
               </li>
             </ol>
+            <p class="note">
+              This allows the user agent to not require user activation, for
+              example to support redirect authentication flows where a user
```
rsolomakhin: Let's remove the word "authentication."

index.html
```diff
@@ -3276,6 +3283,32 @@ <h2 id="canmakepayment-protections">
           opening multiple windows (tabs or pop-ups).
         </p>
       </section>
+      <section>
+        <h2 id="user-activation-requirement">
+          Lack of user activation requirement
```
rsolomakhin: Let's remove the "Lack of" part? Keep it only as "User activation requirement"?

rsolomakhin (dismissed):
Welcome to the spec! :-D

ianbjacobs (approved):

ianbjacobs:
Discussed today at the WPWG meeting: https://www.w3.org/2023/06/22-wpwg-minutes#t04

There were no objections to merging the pull request.

index.html
nickburris: Done

index.html
nickburris: Done

stephenmcgruer (approved):

marcoscaceres:
It would have been good to also document and specify the redirect feature... I have concerns this went in without WebKit review because I don't think we support this model (and the redirect without requiring user activation... is that is that thing, like, in HTML?).

marcoscaceres:
also, please kindly always make us the PR template. Otherwise, we miss tests, and bugs filed on other browser engines.

stephenmcgruer:
Hey Marcos; apologies for missing on the PR template, and I agree that we could have done better here to get WebKit views. I will push back slightly and note that at time of PR creation, to my knowledge Apple was not a member of the WPWG and so was not engaging on PaymentRequest/etc issues, **however** we could still have requested a standards position at the very least.

We deliberately wrote the spec change so that it was optional for the user agent to allow, but ack that it may not fully address your concerns. Happy to discuss in a future WPWG meeting!

> and the redirect without requiring user activation... is that is that thing, like, in HTML?

I may not understand your question correctly, but one can redirect in HTML without any user interaction or activation. Via either JavaScript or 302 redirects. One generally cannot open pop-ups without user interaction (though I don't think it's actually spec'd - it's about the individual browsers pop-up blocker logic), so it's fair to note that this change is risky in that direction. We did weight that up for Chrome very seriously and implemented some mitigations, but ultimately felt the use-case of redirect from aggregator --> frictionless merchant checkout flow was worth supporting.

Issue #1040, opened 2024-11-26
Distinguish between user aborting and handler errors


dtjones404:
Hi team, 

We wanted to see if there’s any interest in reviving the functionality described in issue #871.  For context, our PayPal SDK exposes onError and onCancel callbacks, which allow merchants to respond differentially to checkout failures caused by uncaught errors during the processing of a transaction, and those caused by the buyer intentionally abandoning the transaction (e.g. by closing the payment window.)  Currently, this requires us to distinguish between `AbortError`s based on `error.message`:

```
#handlePaymentError(
    error,
  ) {
    if (
      [
        "Request cancelled", // used by Chrome <= 76, Opera 63
        "User closed the Payment Request UI.", // Chrome 77+, Edge 79+, Opera 64+
      ].includes(
        error.message,
      )
    ) {
      this.paymentFlowOptions.onCancel?.();
    } else {
      this.paymentFlowOptions.onError?.(error);
    }
  } 
```

This works but we’d prefer to have a solution that isn’t dependent on vendor-specific error messages which might change in the future.

stephenmcgruer:
cc @rsolomakhin

rsolomakhin:
I would be in favor of using [OperationError](https://webidl.spec.whatwg.org/#operationerror) for checkout failures caused by uncaught errors and [AbortError](https://webidl.spec.whatwg.org/#aborterror) for the buyer intentionally abandoning the transaction.

@marcoscaceres - Would you be on-board for this change?

rsolomakhin:
@dtjones404 - A couple of questions for you:

1. Consider websites that check only for the error codes currently defined in the spec: `AbortError`, `SecurityError`, `RangeError`, `InvalidStateError`, `NotSupportedError`, and `NotAllowedError`. What will happen when a new error code is returned? Is there a way to make this transition smoother?
2. Consider different release cycles and cadences for different browsers: What if Safari ships the new error code a few months before Chrome or vice versa?
3. Consider supporting older browser versions* during the transition period - some newer browsers will have the new error code, while some older one's won't.
4. Consider how browsers gradually roll out features. For example, Chrome may enable new error codes for 1% of users to observe metrics. Then user agent version check would not work.

Some thoughts off the cuff:

1. Feature detection: Add a static method `PaymentRequest.isErrorCodeSupported(errorCode)`?
1. Feature control: Introduce `enableOperationErrorCode: true` boolean parameter to the `new PaymentRequest()` constructor?

[...]

PR #1050, opened 2025-11-13
Allow payment handlers to report back errors via Payment Request API

stephenmcgruer:
See https://github.com/w3c/payment-request/issues/1040

Co-authored by: @marcoscaceres

 * [ ] Modified Web platform tests (link) - **not possible to test**, as a test cannot make an app throw an internal error. We could potentially test this via web-based payment handlers, but those tests would only work in Chromium and not WebKit (as it doesn't ship web-based payment handlers). Otherwise we could define some sort of WebDriver API to install a fake payment app, similar to how WebAuthn works?
 
Implementation commitment:

 * [x] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=309077)
 * [x] [Chromium](https://crbug.com/473478138)
 * [ ] Gecko (link to issue) - N/A, don't ship Payment Request

https://github.com/w3c/web-based-payment-handler/issues/428

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/1050.html" title="Last updated on Mar 3, 2026, 5:33 PM UTC (236c39b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/1050/feba7f3...236c39b.html" title="Last updated on Mar 3, 2026, 5:33 PM UTC (236c39b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1318,10 +1318,11 @@
               </p>
             </aside>
             <p>
-              The |acceptPromise| will later be resolved or rejected by either
-              the <a>user accepts the payment request algorithm</a> or the
-              <a>user aborts the payment request algorithm</a>, which are
-              triggered through interaction with the user interface.
+              The |acceptPromise| will later be resolved or rejected by
+              the <a>user accepts the payment request algorithm</a>, the
+              <a>user aborts the payment request algorithm</a> (which are
+              triggered through interaction with the user interface), or
+              the <a>payment handler indicates an internal error algorithm</a>.
             </p>
             <p data-tests="rejects_if_not_active.https.html">
               If |document| stops being [=Document/fully active=] while the
@@ -2600,9 +2601,10 @@
           <li>Return |retryPromise|.
             <p class="note">
               The |retryPromise| will later be resolved by the <a>user accepts
-              the payment request algorithm</a>, or rejected by either the
-              <a>user aborts the payment request algorithm</a> or <a>abort the
-              update</a>.
+              the payment request algorithm</a>, or rejected by the <a>user
+              aborts the payment request algorithm</a>, <a>abort the update</a>,
+              or the <a>payment handler indicates an internal error
+              algorithm</a>.
             </p>
           </li>
         </ol>
@@ -3946,6 +3948,56 @@
           user interface.
           </li>
         </ol>
+      </section>
+      <section>
+        <h2>
+          Payment handler indicates an internal error algorithm
+        </h2>
+        <p>
+	  The <dfn>payment handler indicates an internal error algorithm</dfn>
+          runs when the payment handler that the user has selected reports an
+          internal error to the user agent. It MUST <a>queue a task</a> on the
+          <a>user interaction task source</a> to perform the following steps:
+        </p>
+        <ol class="algorithm">
+          <li>Let |request:PaymentRequest| be the {{PaymentRequest}} object
+          that the user is interacting with.
+          </li>
+          <li>If |request|.{{PaymentRequest/[[state]]}} is not
+          "[=PaymentRequest/interactive=]", then terminate this algorithm and
+          take no further action.
+          </li>
+          <li>Set |request|.{{PaymentRequest/[[state]]}} to
+          "[=PaymentRequest/closed=]".
+          </li>
+          <li>Set |request|'s <a>payment-relevant browsing context</a>'s
+          <a>payment request is showing</a> boolean to false.
+          </li>
+          <li>Let |error| be an {{"OperationError"}} {{DOMException}}.
+          </li>
+          <li>Let |response:PaymentResponse| be
+          |request|.{{PaymentRequest/[[response]]}}.
+          </li>
+          <li>If |response| is not null:
+            <ol>
+              <li>Set |response|.{{PaymentResponse/[[complete]]}} to true.
+              </li>
+              <li>Assert: |response|.{{PaymentResponse/[[retryPromise]]}} is
+              not null.
+              </li>
+              <li>Reject |response|.{{PaymentResponse/[[retryPromise]]}} with
+              |error|.
+              </li>
+            </ol>
+          </li>
+          <li>Otherwise, reject |request|.{{PaymentRequest/[[acceptPromise]]}}
+          with |error|.
+          </li>
+          <li>Abort the current user interaction and close down any remaining
+          user interface.
+          </li>
+        </ol>
+        </p>
       </section>
       <section>
         <h2>
--- a/index.html
+++ b/index.html
@@ -3956,8 +3956,8 @@
         <p>
 	  The <dfn>payment handler indicates an internal error algorithm</dfn>
           runs when the payment handler that the user has selected reports an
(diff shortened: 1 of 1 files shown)


ianbjacobs (approved):

index.html
```diff
@@ -1317,10 +1317,11 @@ <h2>
               </p>
             </aside>
             <p>
-              The |acceptPromise| will later be resolved or rejected by either
-              the <a>user accepts the payment request algorithm</a> or the
-              <a>user aborts the payment request algorithm</a>, which are
-              triggered through interaction with the user interface.
+              The |acceptPromise| will later be resolved or rejected by
+              the <a>user accepts the payment request algorithm</a>, the
```
```suggestion
              the <a>user accepts the payment request algorithm</a>, or the
```

index.html
```diff
@@ -2599,9 +2600,10 @@ <h2>
           <li>Return |retryPromise|.
             <p class="note">
               The |retryPromise| will later be resolved by the <a>user accepts
-              the payment request algorithm</a>, or rejected by either the
-              <a>user aborts the payment request algorithm</a> or <a>abort the
-              update</a>.
+              the payment request algorithm</a>, or rejected by the <a>user
+              aborts the payment request algorithm</a>, <a>abort the update</a>,
```
```suggestion
              aborts the payment request algorithm</a>, or <a>abort the update</a>,
```

index.html
```diff
@@ -3946,6 +3948,56 @@ <h2>
...
+          "[=PaymentRequest/interactive=]", then terminate this algorithm and
+          take no further action.
+          </li>
+          <li>Set |request|.{{PaymentRequest/[[state]]}} to
+          "[=PaymentRequest/closed=]".
+          </li>
+          <li>Set |request|'s <a>payment-relevant browsing context</a>'s
+          <a>payment request is showing</a> boolean to false.
+          </li>
+          <li>Let |error| be an {{"OperationError"}} {{DOMException}}.
```
marcoscaceres: Nit: Inline this where the rejection happens below (x2)

index.html
```diff
@@ -3946,6 +3948,56 @@ <h2>
...
         </ol>
       </section>
+      <section>
+        <h2>
+          Payment handler indicates an internal error algorithm
+        </h2>
+        <p>
+	  The <dfn>payment handler indicates an internal error algorithm</dfn>
+          runs when the payment handler that the user has selected reports an
+          internal error to the user agent. It MUST <a>queue a task</a> on the
```
marcoscaceres: Nit: 
```suggestion
          internal error to the user agent. <a>Queue a task</a> on the
```

Or move this into step 1. 

This is already running in a "MUST" context, so don't need the MUST here.

index.html
stephenmcgruer: Done.

index.html
stephenmcgruer: Done.

index.html
stephenmcgruer: Moved to step 1.

index.html
stephenmcgruer: No longer applicable; we now throw the `reason`.

stephenmcgruer (commented):
Ok I've incorporated your comments (thanks!) and put together three variants:

1. A version where we take the unspecified error from the platform handler, and use the "let |error| be an appropriate {{DOMException}}" wording -https://github.com/w3c/payment-request/pull/1050/changes/e9eec11036ebd2fcc0bbfd3a7c28cf958b1848ee
2. A version where the payment handler is presumed to pass us a DOMException directly (i.e., any conversion to DOMException happens 'inside' the relevant payment handler and any spec it has) - https://github.com/w3c/payment-request/pull/1050/changes/ae1280b092e7f2c0781ac539d80ba6f75c0e60b9
3. A version which uses a general JavaScript value [like AbortSignal does](https://dom.spec.whatwg.org/#abortsignal-abort-reason), and rejects with that reason - https://github.com/w3c/payment-request/pull/1050/changes/d9767303206ed7dca116564ea3e9aca74d54d334

I personally prefer (2), but I'm open to any of them.

marcoscaceres:
Alternative, building on @stephenmcgruer proposal:
https://github.com/w3c/payment-request/pull/1050/commits/236c39b6a6aa579c130b8b8b9e4c780536212cca

PayPal's SDK exposes `onError` and `onCancel` callbacks to merchants. Currently, the Payment Request API returns `AbortError` for both user cancellation AND payment handler errors, forcing PayPal to parse `error.message` strings that vary by browser version.

## Proposed Solution

Use the existing WebIDL `OperationError` DOMException type to distinguish payment handler errors from user cancellation (`AbortError`).

## Two `OperationError` Cases

### Case 1: OS-Level Error
When the operating system kills the payment UI (e.g., memory pressure when user switches apps):

```
User opens payment sheet → Switches to another app → 
OS kills payment UI for memory → OperationError
```

This is **not** user cancellation - the user didn't intentionally close the sheet.

### Case 2: Payment Handler Internal Error  
When the payment handler encounters an error during processing:

```
User selects PayPal → PayPal server returns 500 → 
PayPal handler signals error → OperationError
```

This is **not** user cancellation - the payment failed due to a technical issue.

## Backwards Compatibility

**Self-healing approach:**
1. PayPal continues using message-based detection for old browsers
2. New browsers will throw `OperationError` instead of `AbortError`
3. PayPal can add `error.name === "OperationError"` check
4. Code naturally works in both old and new browsers:

```javascript
function handlePaymentError(error) {
  // New path: check error type first
  if (error.name === "OperationError") {
    this.onError(error);
    return;
  }
  
  // Legacy path: message-based detection for old browsers
  if (error.name === "AbortError") {
    const cancelMessages = [
      "Request cancelled",
      "User closed the Payment Request UI."
    ];
    if (cancelMessages.some(m => error.message?.includes(m))) {
      this.onCancel();
    } else {
      this.onError(error);
    }
  }
}
```

stephenmcgruer:
(Removing Rouslan for review as he is no longer working on Payment Request :))

marcoscaceres (approved):

marcoscaceres:
I don't think this is testable... but we should file browser bugs and I need to check if WebKit is currently bailing out anywhere with `UnknownError` instead of `OperationError`.

marcoscaceres:
We should get an Ok from the folks in #1040 before merging, in case we've missed something... but otherwise, this seems ok to me.

stephenmcgruer:
Filed bugs for Chromium and WebKit, filed an issue for Web-Based Payment Handler API, and added the following for WPT impact:

> We could potentially test this via web-based payment handlers, but those tests would only work in Chromium and not WebKit (as it doesn't ship web-based payment handlers). Otherwise we could define some sort of WebDriver API to install a fake payment app, similar to how WebAuthn works?

stephenmcgruer:
> We should get an Ok from the folks in #1040 before merging, in case we've missed something... but otherwise, this seems ok to me.

Looks like we have that ok - @ianbjacobs any concerns/thoughts from your side, or are we good to land this?

ianbjacobs (approved):
I defer to the implementers here; and you seem happy so I'm happy.

PR #1066, opened 2026-03-27
fix: restore transient activation requirement in show()

### LINKED ISSUE #1064: Revert activation relaxation in show() -- find a general platform solution instead
## Problem

PR [#1009](https://github.com/w3c/payment-request/pull/1009) changed the transient activation requirement in `show()` from a hard requirement to a MAY, with security mitigations left to implementer discretion. The stated motivation was supporting redirect flows where activation is lost during navigation.

The problem is real. The solution is not right.

## Why this is wrong

Turning a hard requirement into a MAY means a conformant browser can skip the activation check entirely, with no normative constraints on when or how. The security considerations section added by #1009 describes mitigations browsers "may" apply -- none of which are required. That is not a spec constraint, it is a suggestion.

This also sets precedent. The Digital Credentials API is now citing this exact change to justify [the same pattern](https://github.com/w3c-fedid/digital-credentials/pull/486) in their spec. WebAuthn has the same gap with conditional mediation. Every API with a transient activation requirement is solving the redirect problem independently, inconsistently, and in ways that leave the door open for abuse.

## The right fix

[WICG Capability Delegation](https://wicg.github.io/capability-delegation/spec.html) was working toward a general solution: explicit, named, consumed delegation of activation through a trusted call chain via `postMessage`. The originating page explicitly passes a typed activation capability to the destination. The intent has a chain of custody. The platform can verify it was handed off intentionally and consumed exactly once.

This is the correct model. It stalled. The right response is to revive it as a joint effort across Payment Request, Digital Credentials, and WebAuthn, and to define the primitive at the HTML level.

## Proposal

1. Revert #1009 (or replace with a note acknowledging the open problem)
2. Open a joint effort with Digital Credentials and WebAuthn to bring a sanctioned-continuation primitive to HTML/WHATWG
3. Once that primitive exists, update Payment Request and other affected specs to use it

I recognise reverting may affect deployed implementations. Happy to discuss how to handle the transition.

marcoscaceres:
Reverts the relaxation introduced in #1009 and removes the non-normative "User activation requirement" section.

## What changed

- Restores `[=transient activation=]` as a hard requirement in `show()` (removes the MAY)
- Removes the "User activation requirement" security considerations section
- Replaces both with a note acknowledging the redirect flow problem and pointing to the general solution being tracked in #1064

## Why

The MAY introduced in #1009 means a conformant browser can skip the activation check entirely with no normative constraints. The security mitigations in the removed section were all non-normative suggestions. That is not a spec constraint.

This is also part of a pattern: Digital Credentials and WebAuthn have the same problem and are reaching for the same local workarounds. The right fix is a sanctioned-continuation primitive at the HTML level (see [WICG Capability Delegation](https://wicg.github.io/capability-delegation/spec.html)), tracked in #1064.

Keeping the activation requirement strict here maintains pressure to find the general solution rather than normalizing per-spec workarounds.

closes #1064 

Implementation commitment:

 * [ ] WebKit (link to issue)
 * [ ] Chromium (link to issue)
 * [ ] Gecko (link to issue)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/pull/1066.html" title="Last updated on May 11, 2026, 7:28 AM UTC (791f837)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-request/1066/0e99859...791f837.html" title="Last updated on May 11, 2026, 7:28 AM UTC (791f837)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1012,31 +1012,24 @@
           <li data-tests=
           "payment-request-show-method.https.html, show-method-postmessage-manual.https.html">
           If the [=relevant global object=] of [=request=] does not have
-          [=transient activation=], the user agent MAY:
+          [=transient activation=]:
             <ol>
               <li>Return [=a promise rejected with=] with a {{"SecurityError"}}
               {{DOMException}}.
               </li>
             </ol>
             <div class="note">
-              <p>
-                This allows the user agent to not require user activation, for
-                example to support redirect flows where a user activation may
-                not be present upon redirect. See <a href=
-                "#user-activation-requirement"></a> for security
-                considerations.
-              </p>
-              <p>
-                See also <a href=
-                "https://github.com/w3c/payment-request/issues/1022">issue
-                #1022</a> for discussion around providing more guidance in the
-                specification on when user agents should or should not require
-                a user activation for {{PaymentRequest/show()}}.
-              </p>
+              Redirect flows can cause legitimate loss of transient activation
+              before a call to {{PaymentRequest/show()}}. This is a known
+              platform-wide problem affecting Payment Request, Digital
+              Credentials, WebAuthn, and other APIs that require user
+              activation. A general solution is being tracked in <a href=
+              "https://github.com/w3c/payment-request/issues/1064">issue
+              #1064</a>.
             </div>
           </li>
-          <li data-tests="show-consume-activation.https.html">Otherwise,
-          [=consume user activation=] of the [=relevant global object=].
+          <li data-tests="show-consume-activation.https.html">[=Consume user
+          activation=] of the [=relevant global object=].
           </li>
           <li>Let |document| be |request|'s [=relevant global object=]'s
           [=associated `Document`=].
@@ -4614,32 +4607,6 @@
           with repeated calls, whether it is the cost of managing multiple
           [=host/registrable domains=] or the user experience friction of
           opening multiple windows (tabs or pop-ups).
-        </p>
-      </section>
-      <section>
-        <h2 id="user-activation-requirement">
-          User activation requirement
-        </h2>
-        <p>
-          If the user agent does not require user activation as part of the
-          {{PaymentRequest/show()}} method, some additional security
-          mitigations should be considered. Not requiring user activation
-          increases the risk of spam and click-jacking attacks, by allowing a
-          Payment Request UI to be initiated without the user interacting with
-          the page immediately beforehand.
-        </p>
-        <p>
-          In order to mitigate spam, the user agent may decide to enforce a
-          user activation requirement after some threshold, for example after
-          the user has already been shown a Payment Request UI without a user
-          activation on the current page. In order to mitigate click-jacking
-          attacks, the user agent may implement a time threshold in which
-          clicks are ignored immediately after a dialog is shown.
-        </p>
-        <p>
-          Another relevant mitigation exists in step 6 of
-          {{PaymentRequest/show()}}, where the document must be visible in
-          order to initiate the user interaction.
         </p>
       </section>
     </section>
--- a/index.html
+++ b/index.html
@@ -1025,7 +1025,9 @@
               Credentials, WebAuthn, and other APIs that require user
               activation. A general solution is being tracked in <a href=
               "https://github.com/w3c/payment-request/issues/1064">issue
-              #1064</a>.
+              #1064</a>. Some user agents have legacy behavior that allows
+              certain calls to {{PaymentRequest/show()}} without requiring
+              user activation.
             </div>
           </li>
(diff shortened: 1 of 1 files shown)


ianbjacobs:
Please don't merge until we've had an opportunity to discuss this.

marcoscaceres:
haha, totally not. Don't worry... this is going to be a long discussion :) 

We have the PR template for good reason too.

stephenmcgruer:
Hey Marcos; thanks for filing this and #1064 . Totally agree that what's in the spec right now is a hack; we had redirect flows that we needed to have working, and our initial attempt to work with the user activation flows folks ran out of steam, so we did the hacky thing. Time to pay down the debt (and thank you for taking on debt you didn't cause!).

I think I would mostly be ok with this revert, though it would be nice to not land it until we see progress in the general solution.

Given Chromium would become (and would remain) non-compliant, however, what do you think about the spec noting that some user agents are known to have non-compliant behaviour here? Too much of a violation of spec purity? :)

marcoscaceres:
We could definitely add a big red note to the spec here, as I think this is a really important issue and highlighting what’s been tried in practice is important. Again, the problem identified is very real and critical we find a solution for it. 

Should we draft something up?

stephenmcgruer:
Sorry for the delay! Yes, let's draft something up, wdyt of:

```
 <div class="note">
              Redirect flows can cause legitimate loss of transient activation
              before a call to {{PaymentRequest/show()}}. This is a known
              platform-wide problem affecting Payment Request, Digital
              Credentials, WebAuthn, and other APIs that require user
              activation. A general solution is being tracked in <a href=
              "https://github.com/w3c/payment-request/issues/1064">issue
              #1064</a>. Currently, some Chromium-based user agents
              are known to have non-compliant behavior here to allow certain
              calls to Payment Request to not require user activation.
</div>
```

? Happy to iterate. Also curious what @ianbjacobs thinks

ianbjacobs:
Could we say "legacy" instead of "non-compliant"?

marcoscaceres:
Thanks for the suggestions @stephenmcgruer and @ianbjacobs. I've updated the note to incorporate both — using "legacy" per Ian's suggestion, and keeping it implementation-neutral (not calling out Chromium specifically):

```html
<div class="note">
  Redirect flows can cause legitimate loss of transient activation
  before a call to {{PaymentRequest/show()}}. This is a known
  platform-wide problem affecting Payment Request, Digital
  Credentials, WebAuthn, and other APIs that require user
  activation. A general solution is being tracked in <a href=
  "https://github.com/w3c/payment-request/issues/1064">issue
  #1064</a>. Some user agents have legacy behavior that allows
  certain calls to {{PaymentRequest/show()}} without requiring
  user activation.
</div>
```

WDYT?

RByers:
FWIW I don't believe we should revert this. Without a viable replacement (no, capability delegation in it's current form is not it), Chrome will not unship our behavior here. So it would seem to be to be spec-fiction to have the spec indicate it's not allowed. Let's try to agree on a realistic plan to fully solve this use case (#1064) before suggesting we break the fix we have.

marcoscaceres:
As it stands, this documents single-engine behavior as a literal standard. WebKit has no plans to implement this, so its inclusion in the spec was accidental: I believe the PR template that prevented such single-engine additions was somehow disregarded without multiple implementer commitment (this should have never landed in the spec in the first place). 

I think we should remove the normative text and work on correcting it properly through the Capabilities proposal. The appropriate place to document browser quirks is within the Web Compat specification, if we want to do that.

stephenmcgruer:
I'd like to keep conversation to the issue as Rick has left a more indepth comment there (https://github.com/w3c/payment-request/issues/1064#issuecomment-4421890389), but just to note:

> As it stands, this documents single-engine behavior as a literal standard. 

To the best of my knowledge, this is incorrect. The current text is in the Candidate Recommendation Draft, which [is explicitly not a standard](https://www.w3.org/standards/types/#x4-2-1-candidate-recommendation-draft):

> These documents MUST NOT be cited as W3C standards and may or may not become W3C standards.

The standard document for Payment Request continues to be https://www.w3.org/TR/2022/REC-payment-request-20220908/ (correct me if I'm wrong @ianbjacobs !).

Again, let's continue discussion on the issue, just felt it was important to address this specific statement in context.

marcoscaceres:
If you click on that link it literally pops up a giant red box saying: 

> This version is outdated!
For the latest version, please look at [https://www.w3.org/TR/payment-request/.](https://www.w3.org/TR/payment-request/)

The standard is always what’s at https://www.w3.org/TR/payment-request/

